PR #93, opened 2021-02-28
Create user design color

raw-vitor:
- design
  - change in colors to standardize the application
- Input
  - color
  - responsiveness
- tooltips
  - position that is created
- background
  -  media queries
  - new background image
 
**New input design colors:**
input


src/views/CreateUser.vue
```diff
@@ -256,6 +249,8 @@ export default {
 
 #avatar{
 	margin-bottom: 20px;
+	background-color: rgb(30, 120, 193);
+	// color: rgb(255,255,255) !important;
```
vitormattos, line 253: lost css commented

PR #134, opened 2021-03-22
l10n: Add an apostrophe

Valdnet:


vitormattos:
@Valdnet the test failed. You need to correct the test. It is a test to verify that the failure does occur.

Valdnet:
@vitormattos Can the apostrophe `\'` be used in this form? If not then please give me a suggestion for improvement.

vitormattos:
Yes, is possible. You also need to change it in the test.

Valdnet:
> You also need to change it in the test.

Hmm. How to do it?

vitormattos:
In the broken test it says where you should correct it.

Click on details

image

You will see the file and the failed test method

image

It is this line that needs to be changed

https://github.com/LibreSign/libresign/blob/main/tests/Unit/Service/AccountServiceTest.php#L94

Another possibility is to search for a part of the text to find all occurrences

Valdnet:
@vitormattos Thanks for the clarification 😉.
PR: #137

rakekniven:
"Don't is your file" sounds a bit strange for me.
Idea: Is not your file

What do you think.

Valdnet:
@rakekniven I was wondering about that too :wink:. I think it should be consistent with this: https://github.com/LibreSign/libresign/blob/99d9ca0c534dec514bad57bf40f35cae8172757a/lib/Helper/JSConfigHelper.php#L83

Valdnet:
ping @vitormattos
Can the two messages be the same?

vitormattos:
`This is not your file` is the same as `Is not your file`. 
Very similar. Long form is better?

rakekniven:
Please align it to your other messages. You mostley write in short form like `Empty file` and not `This is an empty file`.

Valdnet:
@vitormattos But can the text string `Don't is your file` be changed to `This is not your file`.
Will it be as expected?

Valdnet:
@rakekniven But it's not about an empty file 😉.

vitormattos:
Correct, it is about a file that was not designed to be signed by the authenticated person.

rakekniven:
> But it's not about an empty file

This was just an example about long and short forms.

This has been merged in the meantime. I am fine with your changes.

Valdnet:
> This was just an example about long and short forms.

I understood now 😉.
Cheers!

PR #137, opened 2021-03-22
Fix test for apostrophe

Valdnet:


vitormattos:
I will accept both pull requests, one corrects the other. Ideally, I should have done it on the previous branch that would be on the same pull request.
Thank you very much for your help.

Issue #191, opened 2021-04-22
Block create account button


vitormattos:
Problem: When clicking on create account today the button remains active allowing the person to click on the button more than once, sending the request more than once to the backend.

PR #198, opened 2021-04-23
Disable create account button when submitting form

raw-vitor:
* JS/CSS
  *  Fluid button animation and scalability
* UX:
  * the button's text can now not be selected


vitormattos (approved):
Close #191

PR #210, opened 2021-05-03
Create option to sign or request signatures of a document through the app files.

vinicios-gomes:


vitormattos:
Branch identical to signature-validate

PR #221, opened 2021-05-07
Signature validate

raw-vitor:
visual rework 
date-fns(Unix epoch)


src/views/Validation.vue
```diff
@@ -72,12 +87,9 @@ export default {
 	},
 	methods: {
 		async validateByUUID() {
-			// eslint-disable-next-line no-console
-			console.log(this.uuid)
 			try {
 				const response = await axios.get(generateUrl(`/apps/libresign/api/0.1/file/validate/uuid/${this.uuid}`))
-				console.info(response)
-				showSuccess('ok')
+				showSuccess(t('libresign', 'ok'))
```
vinicios-gomes, line 92: Por favor, melhore a mensagem para usuario.

vinicios-gomes (changes requested):
Remove package-lock, default package manager is yarn.

Change the toast message that is sent to the user when validating that the document has information.

src/views/Validation.vue
```diff
@@ -90,6 +102,13 @@ export default {
 		changeInfo() {
 			this.hasInfo = !this.hasInfo
 		},
+		formatData(data) {
+			try {
+				return format(fromUnixTime(data), 'MM/dd/yyyy')
```
vitormattos, line 107: Need to be internationalized

PR #321, opened 2021-07-06
l10n: Correct text strings

Valdnet:


lib/Service/SignFileService.php
```diff
@@ -301,7 +301,7 @@ public function validateExistingFile(array $data) {
 			$this->validateHelper->validateLibreSignNodeId($data['file']['fileId']);
 			$this->validateHelper->iRequestedSignThisFile($data['userManager'], $data['file']['fileId']);
 		} else {
-			throw new \Exception($this->l10n->t('Inform or UUID or a File object'));
+			throw new \Exception($this->l10n->t('Inform or UUID or a file object'));
```
vitormattos, line 304: This `file` is upper case first, is reference to `File` object in request of API. Is a pattern write first letter in uppercase of objects.

```suggestion
			throw new \Exception($this->l10n->t('Inform or UUID or a File object'));
```

vitormattos (changes requested):

lib/Service/SignFileService.php
Valdnet: Done 😉.

Valdnet:
@vitormattos Shouldn't `Inform` be as `InForm`?

PR #405, opened 2021-09-29
Reader PDF

vinicios-gomes:
Create a component to view the pdf and to view the document and provide the coordinates to insert signatures or initials into the document.


vitormattos:
moved to new branch

PR #406, opened 2021-10-05
l10n: Add dots and remove space

Valdnet:


vitormattos:
Thanks @Valdnet !

Backend tests are not passing, we are working on resolving them.

I updated the Swagger documentation to implement changes for visible signatures and put the changes on the code before do the changes on API.

PR #452, opened 2021-11-21
Frontend: sms token

vitormattos:


lib/Controller/LibreSignFileController.php
```diff
@@ -154,6 +156,7 @@ private function validate(string $type, $identifier): JSONResponse {
 					'displayName' => $signer->getDisplayName(),
 					'fullName' => $signer->getFullName(),
 					'me' => false,
+					'signer_file_uuid' => null,
```
vitormattos, line 159: why? every time will be null

PR #491, opened 2021-12-06
l10n: Add an ellipsis

Valdnet:


Valdnet:
@vitormattos Does this text string "Loading" only appear when loading a file or signature?

vitormattos (approved):

PR #577, opened 2022-02-01
Fix a non-defaulted function-argement listed after defaulted function arguments.

rotdrop:
Please ask if questions remain, I think its self-explanatory.


rotdrop:
Which one is "the deprecated class"?

PR #578, opened 2022-02-01
Add missing IUserSession in constructor. Is used in ->getPage().

rotdrop:


vitormattos:
Good! I renamed the controller and don't remember to fix the URL routes.

PR #579, opened 2022-02-01
getPdf() should also return NOT_FOUND if the file aquired from the accounts-service is empty.

rotdrop:


vitormattos:
Will be best put this check on `accountService->getPdfByUuid` and throw OCP\AppFramework\Db\DoesNotExistException if !$file. Will make controller more clean and the `getPdfByUuid` more complete

Issue #1170, opened 2022-11-09
Don't work with ARM architecture


vitormattos:
**Describe the bug**
Don't work with ARM architecture

**To Reproduce**
Try to run download binaries.

**Expected behavior**
All of checks in Administration > Libresign working fine

**Environment information:**
 - OS: Linux - ARM

**Help to solve the problem**
File to change: InstallService.php and
- [x] Use the function `php_uname('m')`
- [x] Add the URL and hash of binaries in architecture ARM
- [ ] Test if the setup will work in arm architecture

vitormattos:
Fixed. Install release compatible with stable25

PR #1187, opened 2022-11-14
Make compatible with arm

vitormattos:


vitormattos:
close https://github.com/LibreSign/libresign/issues/1170

PR #1202, opened 2022-11-22
[stable23] Fix Exception Message

vitormattos:
Backport of #1184


vitormattos:
Stop support to stable23

Issue #889, opened 2022-08-12
Error signature - TCPDF


Alexis-M51:
**Describe the bug**
When I try to sign a document, libresign gives me an error saying that I must contact the administrator

**To Reproduce**

    From any file click on the details (three dots at the right)
    In the LibreSign tab click on "Request"
    Fill the email and click in "Add"
    Click on "Submit Request"
    A popup ask if we want to configure visible signatures, should be a YES/NO dialog but is a OK/CANCEL dialog
    Click "Yes"
    I add a signature
    I go to the email where the signature request was made
    I click on "sign"
    I type the password
    Is it when validating the password that I get the error that appears

**Expected behavior**
The goal is to have a validation of the signature

**Screenshots**
![avant signature](https://user-images.githubusercontent.com/90853905/184315013-5272e240-97f4-4e62-afd8-0ba1fdfd97f7.PNG)
![apres signature](https://user-images.githubusercontent.com/90853905/184315020-05f3b828-bfee-4184-aee0-280d2bf03af8.PNG)
![Inkedpost](https://user-images.githubusercontent.com/90853905/184316764-b6575a33-a657-42a2-8a4a-80fd31375242.jpg)

**Log**

Erreur | libresign | array_map(): Argument #2 ($array) must be of type array, bool given

{"reqId":"Q1qtzCJ3MCjWJ8Kn0wD4","level":3,"time":"2022-08-12T08:16:01+00:00","remoteAddr":"ipv6","user":"email","app":"libresign","method":"POST","url":"/index.php/apps/libresign/api/0.1/sign/uuid/ab3fd2b7-b7f5-4eff-91e9-655603b279f7","message":"array_map(): Argument #2 ($array) must be of type array, bool given","userAgent":"Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:103.0) Gecko/20100101 Firefox/103.0","version":"24.0.3.2","data":{"app":"libresign"}}

**Environment information (please complete the following information):**
 - OS: Ubuntu Server 20.04
 - Browser firefox, edge, chrome
 - LibreSign Version 5.1.4
 - Nextcloud Server Version 24.03

vitormattos:
Hi Alexis, thanks for always you send an excellent report with all steps to reproduce the scenarios.

[...]

PR #1317, opened 2022-12-31
Log error backtrace

vitormattos:


vitormattos:
Ref #889
Add code suggested here: https://github.com/LibreSign/libresign/issues/889#issuecomment-1213182990

PR #1622, opened 2023-04-12
Lock friendsofphp/php-cs-fixer version

vitormattos:
The newest version of this package broked the CI because implemented a new pattern.


vitormattos:
close #1618

PR #1643, opened 2023-04-13
If is a prop, can't be a data.

vitormattos:
Removed src from data and maintained only at props

Screenshot_20230412_212045

### DIFF
--- a/src/Components/PreviewSignature/PreviewSignature.vue
+++ b/src/Components/PreviewSignature/PreviewSignature.vue
@@ -1,7 +1,7 @@
 <template>
     <div class="modal-draw">
         <img :src="src">
-    </div>				
+    </div>
 </template>
 
 <script>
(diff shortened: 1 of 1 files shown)


src/Components/PreviewSignature/PreviewSignature.vue
```diff
@@ -26,7 +26,6 @@ export default {
 		canvas: null,
 		isDrawing: false,
 		color: '#000000',
-		src: null,
```
enieber, line 29: this define src and no allow use in template?

vitormattos: src already was defined at `props`, the reported error is that isn't ok to use the same property name as `data`. I only removed from data and worked to me.

enieber (approved):

PR #1712, opened 2023-05-18
add websocket to progressbar download libs

enieber:
- [x] add websocket in frontend
- [ ] add websocket in backend
- [ ] updated data of progressbar


vitormattos:
I started to implement the socket at backend but I haven't success to do the connection. The connection go to server (I verified with xdebug) but the browser says that can't establish the connection.

Other topic related with this is about security. Will be necessary validate the user that is trying to use the socket.

I believe that to implement this will be a bit complex task at this moment.

enieber:
Maybe we can be use the event bus of nextcloud

https://nextcloud.github.io/nextcloud-event-bus/

vitormattos:
I think that this is only to share values between components.

vitormattos:
Impossible to merge and this is a wip PR.

PR #1935, opened 2023-11-06
Change the default name of signature method

vitormattos:


vitormattos:
close https://github.com/LibreSign/libresign/issues/1936

PR #1937, opened 2023-11-07
Rename setting

vitormattos:
Rename setting to be more clear about the objective


vitormattos:
close https://github.com/LibreSign/libresign/issues/1938

PR #1980, opened 2023-12-01
Bump ocp

vitormattos:
necessary when have changes at Nextxcloud OCP package to don't have problems with signatures of methods/classes from public server context

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -756,16 +756,16 @@
         },
         {
             "name": "symfony/console",
-            "version": "v5.4.31",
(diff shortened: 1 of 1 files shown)


giovannism20 (approved):
why i not also have changes on composer.json file ? i know that is a minor update but probably is important to have it

vitormattos:
Every time we need to do this at the lock file because at composer.json the version is `master`. We need to put the hash of latest commit from master branch of ocp repository

PR #1985, opened 2023-12-03
Downgrade the PHP version to be compatible with Nextcloud server

vitormattos:
The minimum version of PHP at Nextcloud server is 8.0 and if we run the command composer update in an environment with PHP 8.2 will update a lot of packages to versions that isn't compatible with PHP 8.0


vitormattos:
Will replace: https://github.com/LibreSign/libresign/pull/1984

PR #1988, opened 2023-12-07
fix(i18n): Adapted Base64 occurences

rakekniven:


giovannism20:
DCO are breaking yet: https://github.com/LibreSign/libresign/pull/1988/checks

giovannism20:
If you open the failed job u can see the problem and instructions to solve it:
To add your Signed-off-by line to every commit in this branch:

Ensure you have a local copy of your branch by [checking out the pull request locally via command line](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally).
In your local branch, run: git rebase HEAD~2 --signoff
Force push your changes to overwrite the branch: git push --force-with-lease origin patch-5

vitormattos:
Follow an example about how need to be the last row to solve the DCO problem:

Screenshot_20231207_122438

I think that this problem also occur with your commits in other apps that have the same rule.

But I believe that good contributions are best than rules sometimes (not removing the importance of rules).

Again, thanks a lot by your contribution.

vitormattos:
I will verify if have anything else that we can do as configuration of this bot to make more light the DCO rule.

vitormattos:
Strange, you didn't added the signed-off row?

PR #2004, opened 2023-12-16
Cover afterException hook with tests

vitormattos:
This change is to implement a middleware to handle errors and return the erros to client.

The client could be a request from API or a request made from user.

When the request have the header `Accept: 'application/json'` is made to API and will return as json

When the request haven't the header `Accept: 'application/json'` is made from user and will return as HTML with all properties of JSON as initial state

Suggestion of response from backend:

```yaml
openapi: "3.0.2"
info:
  title: Response from routes
  version: "1.0"
paths:
  /test:
    get:
      responses:
        200:
          description: OK
          content:
            application/json:
              schema:
                type: object
                allOf:
                  - type: object
                    properties:
                      action:
                        enum:
                          # ACTION_REDIRECT
                          - 100
                          # ACTION_CREATE_USER
                          - 150
                          # ACTION_DO_NOTHING
                          - 200
                          # ACTION_SIGN
                          - 250
                          # ACTION_SHOW_ERROR
                          - 300
                          # ACTION_SIGNED
                          - 350
                          # ACTION_CREATE_SIGNATURE_PASSWORD
                          - 400
                oneOf:
                  - type: object
                    description: Success
                    properties:
                      message:
                        type: string
                        description: Success message
                        example: Success
                  - type: object
                    description: Error
                    properties:
                      errors:
                        type: array
                        items:
                          type: string
                        example:
                          - User 0 needs an email address
            text/html:
              example: |
                initial-state:
                  config: // Config is deprecated
                  action: 200
                  error:
                    - error1
                    - error2
```

### DIFF
--- a/lib/Middleware/InjectionMiddleware.php
+++ b/lib/Middleware/InjectionMiddleware.php
@@ -144,10 +144,8 @@ class InjectionMiddleware extends Middleware {
 					];
 				}
 				return new JSONResponse(
-					$body,
-					$exception->getCode() === 0
-						? AppFrameworkHttp::STATUS_UNPROCESSABLE_ENTITY
-						: $exception->getCode()
+					data: $body,
+					statusCode: $this->getStatusCodeFromException($exception)
 				);
 			case $exception instanceof PageException:
 				if ($this->isJson($exception->getMessage())) {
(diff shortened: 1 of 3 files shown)


giovannism20:
I understand that in fact whatever the request is, it should be interpreted as a client and there should not be this difference "The client could be a request from API or a request made from user". And therefore, I think we should standardize JSON in the response and the device interested in the request response would worry about parsing the data. But we can discuss more about it.

giovannism20 (approved):

giovannism20 (changes requested):
Wrong approval, sorry.

vitormattos:
Sure @giovannism20 
I maintained this PR as draft because I can't go ahead without made a best clarify round about this pattern and finish this implementation (is incomplete).

This is a very important point of app and will be very nice to make a brainstorm call about with other engineers and create a good pattern and  define how backend will send the data to front after an access to sign url.

giovannism20 (approved):

PR #2019, opened 2023-12-21
Administration settings to handle expiration rules

vitormattos:


src/views/Settings/ExpirationRules.vue
giovannism20, line 1: Can u add a print of this modification ?

vitormattos: [screen-recorder-wed-dec-27-2023-12-23-47.webm](https://github.com/LibreSign/libresign/assets/1079143/074adaf9-a2d5-4107-bfba-f5af33022294)

giovannism20: Is really necessary to have a request on each number update ? I know is not too much updates is something that we  can think about.

giovannism20 (approved):

src/views/Settings/ExpirationRules.vue
vitormattos: Isn't necessary but my limited knowledge about JS made me to do by this way. I also disliked this approach to send a request to each key up and down. Maybe a delay would be good but with my limited knowledge I think that the code would be very bad.

vitormattos: Maybe we can create a follow up issue to implement an improvement and reduce this, but is a JS from admin side and used only by the sysadmin. I think that isn't a big problem.

giovannism20: Yes, we can implement it after. And maybe not a big problem.

vitormattos: We need an issue about this. Could you help creating the issue?

giovannism20: Issue #2030

Issue #1921, opened 2023-11-01
Refactor identification methods


vitormattos:
We need to refactor the API to implement this flow:

![link_confirmação(2)](https://github.com/LibreSign/libresign/assets/1079143/209637d3-fddb-43b2-9210-2300b64ba707)

## Class diagram

```mermaid
classDiagram
    %% class Token {
    %%     +sendCode(identify, message)
    %%     +validateCode(identify, code)
    %% }
    class Certificate {
        +sign()
    }
    %% <<interface>> Token
    <<interface>> Certificate
    class IdentifyMethod {
        +notifyNewRequest()
        +notifyUpdatedRequest()
        +validateForRequest()
        +validateForCreateAccount()
        +processForSign()
        +getEntity()
        +setEntity()
        +cleanEntity()
        +save()
        +delete()
    }
    <<Interface>> IdentifyMethod
    class Sms~SignMethod~ 
    class Telegram~SignMethod~
    class Signal~SignMethod~
    class Email~SignMethod~
    class Password~SignMethod~
    class Account~IdentifyMethod~
    %% class Mediated~IdentifyMethod~
    %% Sms --|> Token : implements
    %% Telegram --|> Token : implements
    %% Signal --|> Token : implements
    %% Email --|> Token : implements
    Password --|> Certificate : implements
    class RequestSignature {
        request(File, Signers)
        -sendInviteUrl()
    }
    class Signer {
        list~IdentifyMethod~ identifyMethod
    }
    RequestSignature --> Signer

    %% Signer --|> Account
    Signer --|> IdentifyMethod
    %% Signer --|> Mediated
    IdentifyMethod --> Sms
    IdentifyMethod --> Telegram
    IdentifyMethod --> Signal
    IdentifyMethod --> Email
    IdentifyMethod --> Password
    IdentifyMethod --> Account
    %% IdentifyMethod --> Mediated
```

## TO-DO
- [ ] Refactor backend
- [ ] Refactor frontend to be compatible with possible changes at endpoints
- [ ] Make all tests to pass

PR #2020, opened 2023-12-21
Validate expired link server side

vitormattos:
To change the expire rules is necessary the follow PR otherwise will be necessary change the settings by occ has was made at integration tests.

 - https://github.com/LibreSign/libresign/pull/2019

## To-do
- [x] Validate if the link was expired
- [x] Validate if the link need to be renewed
- [x] Renew the link

### DIFF
--- a/lib/Controller/AEnvironmentPageAwareController.php
+++ b/lib/Controller/AEnvironmentPageAwareController.php
@@ -67,7 +67,44 @@ abstract class AEnvironmentPageAwareController extends Controller {
 				'errors' => [$this->l10n->t('Invalid UUID')],
 			]), AppFrameworkHttp::STATUS_NOT_FOUND);
 		}
+	}
+
+	/**
+	 * @throws LibresignException
+	 */
+	public function validateSignRequestUuid(string $uuid): void {
+		if (!$this->signRequestEntity instanceof SignRequestEntity
+			|| !$this->fileEntity instanceof FileEntity) {
+			$this->loadSignRequestUuid($uuid);
+		}
 		$this->signFileService->validateSigner($uuid, $this->userSession->getUser());
+		$this->nextcloudFile = $this->signFileService->getNextcloudFile(
+			$this->fileEntity->getNodeId(),
+		);
+	}
+
+	/**
+	 * @throws LibresignException
+	 */
+	public function validateRenewSigner(string $uuid): void {
+		if (!$this->signRequestEntity instanceof SignRequestEntity
+			|| !$this->fileEntity instanceof FileEntity) {
+			$this->loadSignRequestUuid($uuid);
+		}
+		$this->signFileService->validateRenewSigner($uuid, $this->userSession->getUser());
+		$this->nextcloudFile = $this->signFileService->getNextcloudFile(
+			$this->fileEntity->getNodeId(),
+		);
+	}
+
+	/**
+	 * @throws LibresignException
+	 */
+	public function loadNextcloudFileFromSignRequestUuid(string $uuid): void {
+		if (!$this->signRequestEntity instanceof SignRequestEntity
+			|| !$this->fileEntity instanceof FileEntity) {
+			$this->loadSignRequestUuid($uuid);
+		}
 		$this->nextcloudFile = $this->signFileService->getNextcloudFile(
 			$this->fileEntity->getNodeId(),
 		);
--- a/lib/Controller/AEnvironmentPageAwareController.php
+++ b/lib/Controller/AEnvironmentPageAwareController.php
@@ -55,7 +55,11 @@ abstract class AEnvironmentPageAwareController extends Controller {
 	/**
 	 * @throws LibresignException
 	 */
-	public function loadSignRequestUuid(string $uuid): void {
+	private function loadEntitiesFromUuid(string $uuid): void {
+		if ($this->signRequestEntity instanceof SignRequestEntity
+			&& $this->fileEntity instanceof FileEntity) {
+			return;
+		}
 		try {
 			$this->signRequestEntity = $this->signFileService->getSignRequest($uuid);
 			$this->fileEntity = $this->signFileService->getFile(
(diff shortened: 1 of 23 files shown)


vitormattos:
Current challenge:
- Renew the link

Appointments:
When we request to sign a file, will have a row to each identify method of a signer at the table `libresign_identify_method`. 

## To-do
- [ ] The field `last_attempt` need to be filled and the field `attempts` need to be incremented +1 each time when an expired link receive an access.

I think that we will need to review this:
- https://github.com/LibreSign/libresign/issues/1921

We also will need to identify how we will change the UUID of sign request.
A possible suggestion is implementing a way to expire the existing UUID and generate a new UUID but this could generate bad side effects because will be necessary generate a new relation between the UUID and the identify method.

The problem is a bit more complex than I thought.

vitormattos:
```bash
composer require lfkeitel/phptotp
```

```php
// Suggestion to be used as seed. At this example is the UUID of signer
$secret = 'ac58b389-8dab-4f13-af80-84f175e3f141';
// Size of token
$length = 6;
// Algorithm to use
$hash = 'sha1';
// The last time that the signer was notified
$time = 1703256322;
// The window size that the token will be valid
$timeInterval = 300;
$totp = new \lfkeitel\phptotp\Totp(
    algo: $hash,
    start: $time,
    ti: $timeInterval
);
$token = $totp->GenerateToken(
    key: $secret,
    time: $time,
    length: $length
);
```
This will generate a token with X chars based in date that the signer was requested to sign the file and using the UUID as secret. Maybe we can use the Instance ID and other data to define the secret.

Pros: no impact at database
Cons: we will need to think about if we need to store in a log all attempts 

Suggestion to solve the cons: store the data of attempts in a JSON field inside the identify_method table.

tests/psalm-baseline.xml
```diff
@@ -154,7 +149,8 @@
     </UndefinedClass>
   </file>
   <file src="lib/Service/RequestSignatureService.php">
-    <UndefinedClass occurrences="2">
+    <UndefinedClass occurrences="3">
+      <code>UUIDUtil</code>
       <code>UUIDUtil</code>
       <code>UUIDUtil</code>
     </UndefinedClass>
```
giovannism20, line 156: Why do we use this section?

lib/Controller/AEnvironmentPageAwareController.php
```diff
@@ -67,12 +67,49 @@ public function loadSignRequestUuid(string $uuid): void {
...
+	}
+
+	/**
+	 * @throws LibresignException
+	 */
+	public function loadNextcloudFileFromSignRequestUuid(string $uuid): void {
+		if (!$this->signRequestEntity instanceof SignRequestEntity
+			|| !$this->fileEntity instanceof FileEntity) {
+			$this->loadSignRequestUuid($uuid);
+		}
```
giovannism20, line 107: Repeated section of code in functions, we can extract this to another point or we can keep it for readability

lib/Controller/AEnvironmentPageAwareController.php
```diff
@@ -67,12 +67,49 @@ public function loadSignRequestUuid(string $uuid): void {
...
 	}
 
+	/**
+	 * @throws LibresignException
+	 */
+	public function validateRenewSigner(string $uuid): void {
+		if (!$this->signRequestEntity instanceof SignRequestEntity
+			|| !$this->fileEntity instanceof FileEntity) {
+			$this->loadSignRequestUuid($uuid);
+		}
```
giovannism20, line 93: Repeated section of code in functions, we can extract this to another point or we can keep it for readability

lib/Controller/AEnvironmentPageAwareController.php
```diff
@@ -67,12 +67,49 @@ public function loadSignRequestUuid(string $uuid): void {
...
+	}
+
+	/**
+	 * @throws LibresignException
+	 */
+	public function validateSignRequestUuid(string $uuid): void {
+		if (!$this->signRequestEntity instanceof SignRequestEntity
+			|| !$this->fileEntity instanceof FileEntity) {
+			$this->loadSignRequestUuid($uuid);
+		}
```
giovannism20, line 79: Repeated section of code in functions, we can extract this to another point or we can keep it for readability

lib/Controller/AEnvironmentPageAwareController.php
giovannism20, line 1: My biggest concern is regarding the frequent use of "instanceOf" mainly because hierarchy is taken into account. Ex:

```
<?php
class ParentClass
{
}

class MyClass extends ParentClass
{
}

$a = new MyClass;

var_dump($a instanceof MyClass);
var_dump($a instanceof ParentClass);
?>

```
The return of the code above:

```
bool(true)
bool(true)
```

From: https://www.php.net/manual/en/language.operators.type.php

giovannism20: We might consider using:

https://refactoring.guru/design-patterns/strategy
https://refactoring.guru/design-patterns/visitor

tests/psalm-baseline.xml
vitormattos: UUIDUtil is a method from a class that don't exists on our side and because this the psalm return an error. Is a class from 3rdparty of server side and if we add this dependency to LibreSign composer.json could generate a conflict when the version of 3rdparty will be updated.

The psalm-baseline file is a file that map all the cases to be ignored.

Every when have an error reported by psalm that we want to ignore, we need to update the psalm baseline. Exists a script at composer.json with the command to update this.

Other moment that we need to do this is when we fix a problem that was ignored at psalm baseline.

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -257,6 +257,13 @@ public function deleteSignerFromFileOfPreviousListing(int $signerSequence, int $
 		$this->sendOCSRequest('delete', '/apps/libresign/api/v1/sign/file_id/' . $fileId . '/'. $signRequestId);
 	}
 
+	/**
+	 * @When /^wait for ([0-9]+) (second|seconds)$/
+	 */
+	public function waitForXSecond($seconds): void {
+		sleep($seconds);
+	}
```
giovannism20, line 265: 👍🤣

tests/psalm-baseline.xml
giovannism20: Thank u for the explanation. Resolving it.

lib/Controller/AEnvironmentPageAwareController.php
vitormattos: The objective of this class is to check if is a controller that extends it. The type of child class isn't important, only is important if is a child of it.

giovannism20 (approved):

vitormattos:
ref: #2026

PR #2027, opened 2023-12-27
Disable page preview

vitormattos:
The default behavior is disabled.

To enable will be necessary run an `occ` command, but I think that we don't will need this code.

### DIFF
--- a/lib/Command/Configure/Check.php
+++ b/lib/Command/Configure/Check.php
@@ -25,7 +25,9 @@ declare(strict_types=1);
 namespace OCA\Libresign\Command\Configure;
 
 use OC\Core\Command\Base;
+use OCA\Libresign\AppInfo\Application;
 use OCA\Libresign\Service\ConfigureCheckService;
+use OCP\IConfig;
 use Symfony\Component\Console\Helper\Table;
 use Symfony\Component\Console\Helper\TableCell;
 use Symfony\Component\Console\Helper\TableCellStyle;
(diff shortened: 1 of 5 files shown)


lib/Command/Configure/Check.php
```diff
@@ -61,13 +62,27 @@ protected function configure(): void {
...
+		if ($this->pagePreviewAsImage) {
+			$preview = $input->getOption('preview');
+			$sign = $input->getOption('sign');
+			$certificate = $input->getOption('certificate');
+			$all = (!$preview && !$sign && !$certificate);
+		} else {
+			$sign = $input->getOption('sign');
+			$certificate = $input->getOption('certificate');
+			$all = (!$sign && !$certificate);
+		}
```
giovannism20, line 85: The code for each condition is so similar - just an aleatory comment.

lib/Command/Configure/Check.php
```diff
@@ -61,13 +62,27 @@ protected function configure(): void {
 				InputOption::VALUE_NONE,
 				'Check requirements to use root certificate'
 			);
+		if ($this->pagePreviewAsImage) {
+			$this
+				->addOption('preview',
+					'p',
+					InputOption::VALUE_NONE,
+					'Check requirements to generate image preview'
```
giovannism20, line 70: Can we use named parameters here ?

giovannism20 (changes requested):

enieber (approved):

lib/Command/Configure/Check.php
vitormattos: hum.... Can we do this in another PR as technical debt?

Motivation: We have others commands with options and inside this file have other options. If I change only this point will be strange because will be the unique with named parameters. If I change the other points will be strange because is outside of context of this PR.

lib/Command/Configure/Check.php
vitormattos: Yes, very ugly. I made a small improvement, would you check again?

lib/Command/Configure/Check.php
giovannism20: Yes. Can u create an issue for working on it ?
My problem is to see value `p` and need to go throw `addOption` definition.

lib/Command/Configure/Check.php
giovannism20, line 1: Thank u a lot for it!

giovannism20 (approved):

lib/Command/Configure/Check.php
vitormattos: Issue created

PR #2028, opened 2023-12-27
Generate a session as guest and send the action to view sign page

vitormattos:
Background scenario:

- after pass by all checks to access the sign page

## To-do
- [x] Implement integration tests
- renewal interval setting
  - if true
    ~- [ ] Renew the date of last attempt and increment the attempts counter~
  - else
    ~- [ ] Make possible sign the document~


tests/integration/features/sign/request.feature
giovannism20, line 1: Great modification!

vitormattos:
I didn't incremented the attempt. I used session and worked fine.

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -48,6 +48,16 @@ public function setOpenedEmailStorage(OpenedEmailStorage $storage): void {
...
 
+	public function sendRequest(string $verb, string $url, $body = null, array $headers = [], array $options = []): void {
+		if (!is_null($this->currentUser)) {
+			$options = array_merge(
+				['cookies' => $this->getUserCookieJar($this->currentUser)],
+				$options
+			);
+		}
+		parent::sendRequest($verb, $url, $body, $headers, $options);
+	}
```
giovannism20, line 59: Cannot find the usage of this method, can u appoint it, please ?

tests/integration/features/sign/request.feature
```diff
@@ -101,7 +101,7 @@ Feature: request-signature
     And I open the latest email to "signer2@domain.test" with subject "LibreSign: There is a file for you to sign"
     And I fetch the signer UUID from opened email
     And as user ""
-    When wait for 1 second
+    When wait for 2 second
```
giovannism20, line 104: Just to have a note: maybe in future will be better if we use static time - it will be better also for other programmers to understand the code

tests/integration/features/sign/request.feature
```diff
@@ -113,23 +113,25 @@ Feature: request-signature
...
     And my inbox is empty
     And run the command "libresign:developer:reset --all"
-    And run the command "config:app:set libresign renewal_interval --value 1"
-    When sending "post" to ocs "/apps/libresign/api/v1/request-signature"
+    And sending "post" to ocs "/apps/libresign/api/v1/request-signature"
       | file | {"url":"<BASE_URL>/apps/libresign/develop/pdf"} |
       | users | [{"identify":{"email":"signer2@domain.test"}}] |
       | name | document |
-    Then the response should have a status code 200
+    And the response should have a status code 200
```
giovannism20, line 120: In this case which is the difference of "Then" and "And" usage ? Maybe "Then" make more sense...

lib/Service/SessionService.php
```diff
@@ -0,0 +1,50 @@
...
+
+	public function getSignStartTime(): int {
+		return $this->session->get('libresign-sign-start-time') ?? 0;
+	}
+
+	public function resetDurationOfSignPage(): void {
+		$renewalInterval = (int) $this->config->getAppValue(Application::APP_ID, 'renewal_interval', '0');
+		if ($renewalInterval <= 0) {
+			return;
+		}
```
giovannism20, line 46:
```suggestion
		if ($renewalInterval <= 0) return;
```

lib/Service/SessionService.php
```diff
@@ -0,0 +1,50 @@
...
+		protected IConfig $config,
+	) {
+	}
+
+	public function getSignStartTime(): int {
+		return $this->session->get('libresign-sign-start-time') ?? 0;
+	}
+
+	public function resetDurationOfSignPage(): void {
+		$renewalInterval = (int) $this->config->getAppValue(Application::APP_ID, 'renewal_interval', '0');
```
giovannism20, line 43: This zero sound like a magic number. Move it for a constant.

lib/Service/SessionService.php
giovannism20, line 46:
```suggestion
  const NO_RENEWAL_INTERVAL = 0;

  $renewalInterval = (int) $this->config->getAppValue(Application::APP_ID, 'renewal_interval', NO_RENEWAL_INTERVAL);

  if ($renewalInterval <= NO_RENEWAL_INTERVAL) return;
```

lib/Service/IdentifyMethod/Account.php
giovannism20, line 1: How we are using `logger` and `sessionService` on this file ?

lib/Service/IdentifyMethod/Email.php
giovannism20, line 1: How we are using `logger` and `sessionService` on this file ?

giovannism20 (changes requested):

lib/Service/SessionService.php
```diff
@@ -0,0 +1,50 @@
...
+
+class SessionService {
+	public function __construct(
+		protected ISession $session,
+		protected IConfig $config,
+	) {
+	}
+
+	public function getSignStartTime(): int {
+		return $this->session->get('libresign-sign-start-time') ?? 0;
```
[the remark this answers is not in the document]
vitormattos: Is the default value when we have not the value in session.

The default is zero seconds.

tests/integration/features/bootstrap/FeatureContext.php
vitormattos: When we need to use a guest account, is the account that don't have an username, we need to send the cookie as the browser do.

The method stay at parent class and by default only set the option `cookie` when exists an username.

The default value of `currentUser` is null.

We have a step to set the user to "":
```gherkin
And as user ""
```

At this point we will need to send the session cookie to backend. If we don't do this, the logic that get the start time by session, don't will work.

tests/integration/features/sign/request.feature
vitormattos: This maybe will be complex because this isn't an unit test.

When is integration test, if we want to use static time, maybe will be necessary do a big work to identify if the environment is executing as integration test and get the time defined by an external way or if is executing normally to follow with dynamic times.

lib/Service/IdentifyMethod/Account.php
vitormattos: Only because we need to send this to abstract class (parent)

lib/Service/IdentifyMethod/Email.php
vitormattos: Only because we need to send this to abstract class (parent)

lib/Service/SessionService.php
vitormattos: To use xdebug this isn't too cool because don't will have a explicit return in a step.

tests/integration/features/sign/request.feature
vitormattos: I used `And` because is part of a background scenario but if I put as  `background scenario`, this will run for each `scenario` in this `feature`.

I only used `Then` when is the objective of action that we made after a `When`

Given: preparing the screnario
When: The start of action that we need to check by Then
Then: The objective of scenario

vitormattos: I made a fix in a point that was `And` and was best to be `When`

tests/integration/features/bootstrap/FeatureContext.php
vitormattos: Maybe we can create a technical debt issue to identify a best way to define this logic.

To me was a bit strange to use the currentUser by this way. It is not clear to anyone who looks this method because why this was made.

vitormattos:
Considering that now only have questions at this code review, I will mark as resolved and if we have any pending point, we can open a follow up issue to fix the pending points.

We will need this PR here: https://github.com/LibreSign/libresign/pull/2037

PR #2035, opened 2023-12-29
Remove unused method

vitormattos:
I made a grep to find by this method. Is an unused method


giovannism20:
Can u explain more about this grep ?

vitormattos:
Not exactly a grep but I made a search at my IDE and didn't found any reference to these methods.

```
getInfoOfFileToSignUsingSignRequestUuid
getInfoOfFileToSignUsingFileUuid
```

giovannism20:
So maybe would be more interesting if we make some manual tests just to guarantee it

vitormattos:
The previous context:
Before to implement the identify methods logic, all the logic to validate the signer was made by these methods.
Part of logic was moved to identify methods classes, in particular the logic that throw exceptions and validate if the file was signed.
The other part, that return data, was moved to other points of code because, as is possible to see at removed unit test, was very hard to maintain the methods. Was methods with a lot of responsibility and I followed removing the dependency of them. Now, when I started to look the code to make a task, I identified that haven't any place using them.

giovannism20 (approved):

PR #2037, opened 2023-12-29
Organize initial states

vitormattos:
I separated the content of complex methods in specific initial state values.

Making this I identified code executions that wasn't necessary. As example, the endpoint to display page to reset password.

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -303,7 +303,7 @@ class PageController extends AEnvironmentPageAwareController {
 	#[RequireSignRequestUuid]
 	public function resetPassword(): TemplateResponse {
 		$this->initialState->provideInitialState('config',
-			$this->accountService->getConfig($this->userSession->getUser()),
+			$this->accountService->getConfig($this->userSession->getUser())
 		);
 
 		Util::addScript(Application::APP_ID, 'libresign-main');
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -144,6 +144,7 @@ class PageController extends AEnvironmentPageAwareController {
 			->showVisibleElements()
 			->showSigners()
 			->formatFile();
+		$this->initialState->provideInitialState('status', $file['status']);
 		$this->initialState->provideInitialState('visibleElements', $file['visibleElements']);
 		$this->initialState->provideInitialState('signers', $file['signers']);
 		$this->initialState->provideInitialState('description', $this->getSignRequestEntity()->getDescription() ?? '');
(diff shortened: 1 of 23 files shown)


lib/Service/SignFileService.php
```diff
@@ -537,31 +537,17 @@ public function validateRenewSigner($uuid, $user): void {
 		$this->validateHelper->validateRenewSigner($uuid, $user);
 	}
 
-	public function getFileData(FileEntity $fileData, ?IUser $user, ?SignRequestEntity $signRequest = null): array {
-		$return['action'] = JSActions::ACTION_SIGN;
-		$return['sign'] = [
-			'uuid' => $fileData->getUuid(),
-			'filename' => $fileData->getName()
-		];
+	public function getSignerData(?IUser $user, ?SignRequestEntity $signRequest = null): array {
```
giovannism20, line 540:
```suggestion
	public function getSignerData(?IUser $user, ?SignRequestEntity $signRequest = null): array {
  $return = ['user' => ['name' => null]];

  if ($signRequest) {
    $return['user']['name'] = $signRequest->getDisplayName();
  } elseif ($user) {
    $return['user']['name'] = $user->getDisplayName();
  }
  
  return $return;
}
```

lib/Controller/PageController.php
```diff
@@ -335,20 +313,16 @@ public function resetPassword(): TemplateResponse {
...
-			];
+			$this->initialState->provideInitialState('action', JSActions::ACTION_DO_NOTHING);
+			$this->initialState->provideInitialState('errors', [$this->l10n->t('Invalid UUID')]);
 		}
-		$this->initialState->provideInitialState('config', array_merge(
-			$config,
+		$this->initialState->provideInitialState('config',
 			$this->accountService->getConfig($this->userSession->getUser())
-		));
+		);
```
giovannism20, line 325: Have one more closed parentheses here

lib/Controller/PageController.php
```diff
@@ -306,19 +294,9 @@ public function validationFileWithShortUrl(): RedirectResponse {
 	#[PublicPage]
 	#[RequireSignRequestUuid]
 	public function resetPassword(): TemplateResponse {
-		$this->initialState->provideInitialState('config', array_merge(
+		$this->initialState->provideInitialState('config',
 			$this->accountService->getConfig($this->userSession->getUser()),
```
giovannism20, line 298: Can remove the comma on final of the line

giovannism20 (changes requested):

lib/Controller/PageController.php
vitormattos: At row 297? If yes, isn't possible. If are you talking about row 298, yes, is possible.

lib/Controller/PageController.php
vitormattos: was removed

lib/Controller/PageController.php
giovannism20: Yes, I was talking about 298 line.

giovannism20 (approved):

PR #2039, opened 2023-12-30
Follow the same rules as Nextcloud server

vitormattos:
https://github.com/nextcloud/server/blob/master/.editorconfig

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,10 +1,33 @@
+# https://editorconfig.org
+
 root = true
 
 [*]
-end_of_line = lf
-insert_final_newline = true
-
-[*.{js,json,yml}]
 charset = utf-8
-indent_style = space
+end_of_line = lf
+indent_size = 4
+indent_style = tab
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.feature]
 indent_size = 2
+indent_style = space
+
+[*.yml]
+indent_size = 2
+indent_style = space
+
+[*.md]
+trim_trailing_whitespace = false
+
+[*.svg]
+insert_final_newline = false
+
+[package*.json]
+indent_size = 2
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -1,10 +1,33 @@
+# https://editorconfig.org
+
 root = true
 
 [*]
+charset = utf-8
 end_of_line = lf
+indent_size = 4
+indent_style = tab
```
giovannism20, line 9: Why not spaces ?

vitormattos: I don't know, is the pattern used by Nextcloud.

To PHP files is a bit different from PSR.

Maybe because is oldest of PSR.

Nextcloud is a project since 2010. The PSR that defined spaces was made in 2012

vitormattos: I prefer to follow PSR but is very strange to have only LibreSign following a different pattern of Nextcloud. Because this we follow the pattern from server and most of apps.

vitormattos: https://github.com/nextcloud/coding-standard

giovannism20: From PSR we should use spaces over tabs: https://www.php-fig.org/psr/psr-2/#1-overview

Also, if we are thinking about to have LibreSign as a product without all NextCloud actual things make sense to have different code standard or we can just follow it.

giovannism20 (approved):

.editorconfig
vitormattos: Isn't possible to be a product that isn't a Nextcloud app.

To to this will be a very very big refactor, we will need to replace the backend and frontend framework of LibreSign and implement a lot of functionalities that Nextcloud solve to us.

For now we will follow the Nextcloud pattern to be easy to other Nextcloud developers.

PR #2041, opened 2023-12-31
Refactor frontend router

vitormattos:
| ⚠️ |
|---|
| **ATTENTION!!** |

This PR was created from the branch of https://github.com/LibreSign/libresign/pull/2037

Before review this PR, is necessary merge https://github.com/LibreSign/libresign/pull/2037 to reduce the quantity of changes.

## Motivations of this PR

* Organize the frontend routes to follow the same pattern of other apps
* Fixes that was identified during manual tests
* Move part of logic from frontend to backend


lib/Controller/PageController.php
```diff
@@ -122,21 +126,36 @@ public function indexFPath(): TemplateResponse {
 	 */
 	#[NoAdminRequired]
 	#[NoCSRFRequired]
+	#[RequireSetupOk]
```
giovannism20, line 129: Why not main template here also ?

lib/Controller/PageController.php
```diff
@@ -122,21 +126,36 @@ public function indexFPath(): TemplateResponse {
...
 	#[PublicPage]
 	#[RequireSignRequestUuid]
 	public function sign($uuid): TemplateResponse {
 		$this->initialState->provideInitialState('action', JSActions::ACTION_SIGN);
-		$this->initialState->provideInitialState('config', array_merge(
-			$this->accountService->getConfig($this->userSession->getUser()),
+		$this->initialState->provideInitialState('config',
```
giovannism20, line 134: This code section are really confusing and in others areas that we are using for provide initial state. Let's consider to add some design patterns here to split this code.

lib/Controller/PageController.php
vitormattos: What about main template?
Can you explain more about this?

lib/Controller/PageController.php
vitormattos: Maybe after. At this moment we need to think more about if is necessary or not or also if we will split again the config.

For now the intention is only to reduce the responsibility of config method.

lib/Controller/PageController.php
giovannism20: Like this: `#[RequireSetupOk(template: 'main')]`

giovannism20 (approved):

PR #2042, opened 2023-12-31
Refactor: File Controller Refactor

giovannism20:
I was taking a look on: lib/Controller/FileController.php and see that the coding between for `validate` method which is between lines 74:133 is a little bit nested and probably for each type of file will be added a new `else if` condition which can be bad to maintain.

Maybe we can just ignore it but a way for try the implementation of strategy design pattern: https://refactoring.guru/design-patterns/strategy. The idea of it is to split logic for each type a file into a class and for each method we can just create a new class responsible for the logic.


vitormattos:
abandoned

PR #2060, opened 2024-01-04
Make possible change renderAs

vitormattos:
To be possible customize the renderAs argument of TemplateResponse class, I implemented the same of was made to customize the template name. In the future, maybe will be good replace this by specific error class, an error class to each kind of template. But for now maybe this was the most simple way that I found to solve the problem.


vitormattos:
This PR is no longer needed for now. Maybe we can keep him waiting and then close.

vitormattos:
Sounds that isn't more necessary

PR #2061, opened 2024-01-04
Disable execution of composer update every when install a new package

vitormattos:
This is very bad because every time when we need only to install a package, the update script was executed updating packages that I don't want to update. I removed this.


giovannism20 (approved):
Agree with this change. Doesn't make sense to have all packages update when we installing a package.

Issue #2344, opened 2024-02-22
pdftk error using 8.0.0-rc4 with arch64


vitormattos:
![image](https://github.com/LibreSign/libresign/assets/29739547/80c89a49-67ed-4364-9520-b1e58bb98fa8)
![image](https://github.com/LibreSign/libresign/assets/29739547/7b7fee03-b384-4db0-9245-18e6d23859a8)

Did download 3 times, each time deleting the binary pdftk file, in case it was a corrupted download.

EDIT : Server is arch64, downloaded binary is only available for x86_64.

_Originally posted by @Aeris1One in https://github.com/LibreSign/libresign/issues/2327#issuecomment-1959953087_

Aeris1One:
Uh, that's either ARM64 or aarch64, not arch64, sorry for the typo :/

Server is a Hetzner ARM64 CAX11 Cloud Server

vitormattos:
@Aeris1One could you provide the output of the follow command?
```bash
php -r "echo php_uname('m').PHP_EOL;"
```

vitormattos:
I used this project:
https://gitlab.com/pdftk-java/pdftk
At this project, have a version of PDFTK to ARM, here: https://gitlab.com/pdftk-java/pdftk/uploads/2b32d1c2f855f92cea379f566a882ac3/native-image.zip
Could you test if the file pdftk inside this zip file work fine to you?

Aeris1One:
It doesn't. Seems I'm not the only one to have a problem with that executable, see the comment just below the attachment in https://gitlab.com/pdftk-java/pdftk/-/issues/129

Can't you use the .jar file ? It should be arch-agnostic

vitormattos:
Yes, could be this.
We already have the java binary, we only need to use the java binary to run the jar....

vitormattos:
I made a new push to PR replacing by jar version.

Aeris1One:
Thanks, I'll try it in a few hours.

> Yes, could be this. We already have the java binary, we only need to use the java binary to run the jar....

Just need to be careful to install the java binary before the pdftk one.

Also, unrelated but it seems in 8.0.0-rc4 that the /java/ subdirectory in app data directory isn't created. Had to manually create it before running `occ libresign:install --java` to make it working.

PR #2345, opened 2024-02-22
Add support to arm64

vitormattos:
ref: #2344


giovannism20:
Thank u! ❤️ :octocat:

PR #2717, opened 2024-04-12
fix: Read legacy certificate

vitormattos:
Reference:

- https://github.com/php/php-src/issues/12128
- https://www.php.net/manual/en/function.openssl-pkcs12-read.php#128992


cabaseira:
Thank you!!

PR #2795, opened 2024-04-18
Fix css of vue advanced cropper

vitormattos:
|before|after|
|---|---|
|Screenshot_20240418_081003|Screenshot_20240418_080503|


cabaseira:
Thank you!!

PR #2798, opened 2024-04-18
fix: prevent to send non integer value to API

vitormattos:


cabaseira:
Thank you!!

Issue #3026, opened 2024-05-16
Extended check for Java binaries. Support for Alpine/Nextcloud-AIO


neocult-de:
**Is your feature request related to a problem? Please describe.**
The currently used version of java supports the majority of glibc based linux distributions. The Nextcloud-AIO system though builds its containers with Alpine where possible and it is musl libc based. Then the java build with glibc will not work. Manual modifications are necessary and version check might fail mostly. 

**Describe the solution you'd like**
There should be a detection mechanism for musl libc and use an according openjdk version and allow the version and runtime check for it. 

**Describe alternatives you've considered**
I have tried to add compatibility to glibc with "apk add gcompat libstdc++" in the nextcloud-container. With initially good results. But it can break anytime.
I also added the repository version: "apk add openjdk11" and link it instead "ln -s /usr/bin/java java" #inside the java/bin folder of libresign - But then version mismatch is obviously a thing, but it is usable. (pdftk is on a green status and java running)

**Additional context**
As Nextcloud-AIO is an official release of Nextcloud which is quite popular, support for "Alpine"-based Java should increase adoption and reach of LibreSign as soon as it is easy to install there too.

vitormattos:
Thanks by your help to LibreSign!

I will look as a soon.

neocult-de:
As the issue would seem just to get a Alpine compatible openjdk release I dig a bit further to advance the cause of LibreSign for all of us.

It might be already in use for LibreSign, but the current free releases of the Eclipse Temurin project would fit LibreSign needs well. As the only issue at the moment is that the downloaded java version is the "linux x64" release for libc and there is an aquivalent release for "alpine linux x64" available (Version details below).

If we download the same release 21.0.2_13 for Alpine Linux and drop it in the java folder we get a successful version check and everything is running (See attached picture as well):

`
sudo docker exec --user www-data -it nextcloud-aio-nextcloud php occ libresign:configure:check
 --------- ----------------- ----------------------------------------------------------------------------------------- -----
  Status    Resource          Message                                                                                   Tip
 --------- ----------------- ----------------------------------------------------------------------------------------- -----
  success   java              Java version: openjdk version "21.0.2" 2024-01-16 LTS
  success   java              Java binary: /mnt/ncdata/appdata_ocn3bta2ji2r/libresign/java/jdk-21.0.2+13-jre/bin/java
  success   pdftk             PDFtk version: 3.3.3
[...]

PR #3049, opened 2024-05-23
fix: check if is alpine

vitormattos:
This is necessary to download specific Java package to Alpine.

Reference:
- #3026


neocult-de:
The additional checks you implemented to determine alpine work, but when you check on the result you use the Identifier ID instead of "version" which would be the correct one (named group in the regex)

I made a pull request here: https://github.com/LibreSign/libresign/pull/3053 to address that issue. With your two commits and the bugfix in the pull request it works now well to install compatible java with LibreSign in Nextcloud AIO / Alpine based Nextcloud

PR #3053, opened 2024-05-23
Wrong identifier on new distribution check (Successfully tested on Nextcloud-AIO)

neocult-de:
Fixing the check on regular expression matches. The named group is <version> not <ID> therefore the field have to be $matches['version'] to make a correct check

### Related Issue
https://github.com/LibreSign/libresign/pull/3049

### Pull Request Type
- Bugfix

### Pull request checklist
Now testing the modifications on InstallService.php with the two commits and with the change of this pull request the branch of fix/check-if-is-alpine will work well with Nextcloud-AIO / Alpine based Nextcloud - If the distribution is not containing "alpine" the fix does not interfere with any other installations.

### DIFF
--- a/lib/Service/Install/InstallService.php
+++ b/lib/Service/Install/InstallService.php
@@ -363,7 +363,7 @@ class InstallService {
 	private function getLinuxDistributionToDownloadJava(): string {
 		$distribution = shell_exec('cat /etc/*-release');
 		preg_match('/^ID=(?<version>.*)$/m', $distribution, $matches);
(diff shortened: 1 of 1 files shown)


neocult-de:
The additional checks you implemented to determine alpine work, but when you check on the result you use the Identifier ID instead of "version" which would be the correct one (named group in the regex)

I made a pull request here: https://github.com/LibreSign/libresign/pull/3053 to address that issue. With your two commits and the bugfix in the pull request it works now well to install compatible java with LibreSign in Nextcloud AIO / Alpine based Nextcloud

Issue #1889, opened 2023-10-19
Security: Contracts available through user_id only.


feefladder:
**Is your feature request related to a problem? Please describe.**
If you delete your nextcloud account and someone else creates an account with the same username, all contracts the previous user has signed show up in Libresign as if signed by the new user. This is a security risk.
![image](https://github.com/LibreSign/libresign/assets/33122845/8ba14c8b-e73a-4cd9-ae06-5ea8a54193ed)

**Describe the solution you'd like**
All data related to a user should be deleted when they delete their user account, (right to be forgotten).
Set the `'user_id'` to be a foreign key of the `'users'` table with `'onDelete' => 'CASCADE'`

**Describe alternatives you've considered**
I don't know, maybe some hooks on when a user is deleted or something.

**Additional context**
I'm testing Libresign to enhance data sharing as the only FOSS contracting/licensing solution I could find.

vitormattos:
Will be fixed in the next version. Thanks for your issue.

vitormattos:
Maybe a possible problem to use a logic like cascade is that will delete all files that an user signed.

Maybe will be necessary add random chars at LibreSign table and concatenate to uid at LibreSign tables. I don't know what Nextcloud do in other cases, by example when an user send a comment in a file or create a card on deck or anything else. Will be necessary check what will the best approach.

vitormattos:
> there would be reasons to keep such contracts, for example if a dispute should arrive after a user has left the Nextcloud

This is a very important point.

We will need to think more about this to identify a way to solve this issue.

Maybe, a possible solution would be to don't associate the file directly to Nextcloud account, I think that this will make possible to delete the association from the person that will sign the file and the Nextcloud account associated with this person when an account is deleted.

I'm working to implement the follow issue:
- https://github.com/LibreSign/libresign/issues/1921
As you can look, have in the down left side a database diagram and I think that this structure of tables will solve the problem.

This will be a big work and I will need help to do this as you can look here:
- https://github.com/LibreSign/libresign/issues/1959

vitormattos:
Could you check again at the newest release of LibreSign?

I made a lot of improvements.

Closing this issue as solved.

If this issue persists, don't hesitate to open a new issue making reference to this.

> [!NOTE]
> ### If you like this app, don't hesitate to help us
> Ways to help this project:
> - Creating a very nice review of this project at:
>   - **social networks** like LinkedIn, Instagram, etc and putting the @LibreSign and @LibreCodeCoop
[...]

PR #3211, opened 2024-06-25
fix: neutralize deleted users

vitormattos:
Prevent to show files of deleted accounts when is created a new account with the same user id of deleted account.


vitormattos:
close #1889

Issue #2916, opened 2024-04-26
Only one signature gets written to the PDF, even when multiple signature placements are present.


metheos:
When defining signature locations you are able to click the signer's name multiple times to place multiple signatures throughout the document to be filled. Upon signing, only the first signature location specified is actually filled.

Place more than one signature placement box on a document for the same signer.
Submit the document for signatures.
Open the signing request and sign the document.
View the signed document with only 1 signature present.

**Expected behavior**
Each specified signature location should have the signer's signature present.

**Screenshots**
![Screenshot 2024-04-26 at 15-03-33 MCC Cloud](https://github.com/LibreSign/libresign/assets/4603749/b2ee8d3c-e82f-4e4e-9f64-f0843aced7ab)
![Screenshot 2024-04-26 at 15-04-03 SureColor_T5170_brochure (copy) signed pdf](https://github.com/LibreSign/libresign/assets/4603749/bf7078e4-4565-4788-9460-2daa5b29f0d3)

**Environment information (please complete the following information):**
Ubuntu 22.04.4 LTS
Chrome 124.0.6367.78
LibreSign 8.0.0
Nextcloud 28.0.4
Nothing logged

**Additional context**
Signing request was sent to an email address without an account. Account requirement is not enabled.

vitormattos:
Fixed by:
- #2994

Validate at next release and if possible, go back here with your feedback about the fix.

Thanks a lot by your help to LibreSign.

> [!NOTE]
> ### If you like this app, don't hesitate to help us
> Ways to help this project:
> - Creating a very nice review of this project at:
>   - **social networks** like LinkedIn, Instagram, etc and putting the @LibreSign and @LibreCodeCoop
>   - **AlternativeTo** https://alternativeto.net/software/libresign/
>   - **Nextcloud apps store**: https://apps.nextcloud.com/apps/libresign
> - Sponsoring the development by **GitHub sponsor** https://github.com/sponsors/LibreSign
> - **helping the translations** on [Transifex](https://app.transifex.com/nextcloud/nextcloud/libresign/)
> - contacting us to have **Enterprise support**: https://libresign.coop

metheos:
I installed 8.0.1 and this issue is still happening. I defined a signature for the same person on each page of a 3 page document and only the first signature was filled after signing.

vitormattos:
> I installed 8.0.1 and this issue is still happening. I defined a signature for the same person on each page of a 3 page document and only the first signature was filled after signing.

[...]

PR #3226, opened 2024-06-25
fix: Make possible use multiple signatures of same signer

vitormattos:
|🏚**Before**|🏡**After**|
|---|---|
|Screenshot_20240625_153526|Screenshot_20240625_153437|


vitormattos:
close #2916

Issue #2880, opened 2024-04-24
Duplicated notification message


bigbeka:
**Describe the bug**
When creating a signature (typed or drawn) it shows duplicated message.

**To Reproduce**
Steps to reproduce the behaviour:
1. Send document to sign
2. Try signing the document and create a signature by drawing or by trying the name.

<img width="752" alt="Screenshot 2024-04-24 at 17 00 52" src="https://github.com/LibreSign/libresign/assets/53529055/91802157-1ba2-4731-96fc-0c13788f8298">

henmohr:
The message occurs even if you don't upload a document, just by creating a new signature.

vitormattos:
Could you check again at the newest release of LibreSign?

I made a lot of improvements.

Closing this issue as solved.

If this issue persists, don't hesitate to open a new issue making reference to this.

> [!NOTE]
> ### If you like this app, don't hesitate to help us
> Ways to help this project:
> - Creating a very nice review of this project at:
>   - **social networks** like LinkedIn, Instagram, etc and putting the @LibreSign and @LibreCodeCoop
>   - **AlternativeTo** https://alternativeto.net/software/libresign/
>   - **Nextcloud apps store**: https://apps.nextcloud.com/apps/libresign
> - Sponsoring the development by **GitHub sponsor** https://github.com/sponsors/LibreSign
> - **helping the translations** on [Transifex](https://app.transifex.com/nextcloud/nextcloud/libresign/)
> - contacting us to have **Enterprise support**: https://libresign.coop

PR #3665, opened 2024-09-13
fix: prevent duplicate text

vitormattos:
The plural function of Nextcloud only work fine to identify if is singular or plural if the text contains a placeholder with a modifier. Only using a count to identify if is singular or plural, don't woks fine.


vitormattos:
Fix #2880

PR #3714, opened 2024-09-16
fix: can close modal after sign file at app Files

vitormattos:


vitormattos:
Replaced by:
- https://github.com/LibreSign/libresign/pull/3716

Issue #3143, opened 2024-06-10
Change the file list to be more similar to app Files


vitormattos:
- implement pagination by scrolling
- Toggle the view mode between list and grid view
- Implement ordering
  - Date
  - Name
  - Status
- Implement menu with actions
  - Be possible to remove a file from LibreSign
    - https://github.com/LibreSign/libresign/issues/2399
    - https://github.com/LibreSign/libresign/issues/831
  - Sign the document when is possible
  - Validate
  - Manage signers, when is possible
    Go to add signers flow

vitormattos:
Hi @DazeEnd, I like a lot your contributions to LibreSign and I will enjoy too if you can give some feedback about this issue and about the PR that we made to solve this issue.

If you wish to make tests, we can create a pre-release with LibreSign builded with the PR #3746.

DazeEnd:
@vitormattos I'm excited that this is being implemented. I would be happy to test, if you can put together a pre-release version that I can install.

One question: Will this new feature "clean up" files that have already been deleted or had signers removed? Or will this just fix the issue going forward?

PR #3746, opened 2024-09-20
feat: rewrite file list

vitormattos:
Implement
- https://github.com/LibreSign/libresign/issues/3143


src/views/FilesList/FileListFilter/FileListFilter.vue
```diff
@@ -0,0 +1,58 @@
...
+	color: var(--color-error-text);
+}
+
+:deep(.button-vue) {
+	font-weight: normal !important;
+
+	* {
+		font-weight: normal !important;
+	}
+}
```
v-bianchi, line 57: Usually it is not recommended to use `!important` statements in CSS, but since the styles are scoped, there is no risk of messing up other components with these CSS rules

v-bianchi (approved):

vitormattos:
This:

Screenshot_20241003_085036

Need to stay here too:

Screenshot_20241003_085132

vitormattos:
## to-do
- [ ] Make possible delete PDF files and don't only delete the signature request

For now, the the delete feature only delete the signature request, only at LibreSing tables..

I thought in 2 possible ways to solve this but will be necessary to think what's the best:
1. When confirm to delete a file, put a checkbox to say that want to delete the file too and not only the signature request.
2. Put two delete options:
  1. Delete signature request
  2. Delete file. When delete file, will also delete the signature request.

Going more deep into other points:
Now, when upload a new file, the file only is added when you define the signer, but the file stay at Nextcloud side. I think that would be good to store at LibreSign side too as draft before to define the signers. With this change will make possible to see and delete a file that was uploaded by LibreSign side.

- [ ] Save file reference in the database as a draft as soon as it is uploaded, before defining who will sign.

src/views/FilesList/FileListFilter/FileListFilter.vue
vitormattos: I only made a "copy" from app Files. Maybe we can refactor this.

vitormattos:
To who see this PR:

A point that will make me and everyone involved in the development of LibreSign very happy is if  we receive more recurring funding from LibreSign's GitHub Sponsor. This will greatly help to keep the project developing.

If you have any other ideas on how to help the project continue to improve, we would be happy to hear from you: contact@libesign.coop

Fuseteam:
> I thought in 2 possible ways to solve this but will be necessary to think what's the best:
> 
>     1. When confirm to delete a file, put a checkbox to say that want to delete the file too and not only the signature request.
> 
>     2. Put two delete options:
>        
>        1. Delete signature request
>        2. Delete file. When delete file, will also delete the signature request.
> 

I agree that option 1 would be easier UX to most users, it is also very similar to how whatsapp handles "delete media" in chats

p.s.
unfortunately i'm not in a position to able to do a donation

vitormattos:
Screenshot_20241105_125131

PR #3831, opened 2024-10-26
feat: add OWASP dependency check

vitormattos:
## to-do
- [x] fix package-lock.json?pdfjs-dist: GHSA-wgrm-67xf-hhpq(10.0), CVE-2024-4367(8.800000190734863)

Will be necessary check if is possible:
- [x] bump dependencies of https://github.com/LibreSign/vue-pdf-editor
  - https://github.com/LibreSign/libresign/pull/3832

Sounds that this action do the same as dependabot but dependabot have a best result. Maybe we don't need to merge this PR.


vitormattos:
Another reason to don't consider this PR:
- Sounds that the repository of this action is abandoned:
    - https://github.com/dependency-check/Dependency-Check_Action

PR #4449, opened 2025-01-22
fix: save as 0 or 1

vitormattos:


vitormattos:
backport/4446/stable30

Issue #4455, opened 2025-01-23
Libresign and Nextcloud: error during configuration step of a sign on a pdf file


ktrls:
Hi, i'm testing your solutions to sign with self-generated certificate the docs on my nextcloud server.

So, I've try to sign the Nextcloud Manual, after that I copied with the name _Nextcloud Manual (copia)_, but appears a notification with this error: _Impossible get metadata from this file._

In the nextcloud logs I read this messages:

_Undefined array key "MediaBox" at /var/www/nextcloud/apps/libresign/lib/Service/PdfParserService.php#83 
Trying to access array offset on null at /var/www/nextcloud/apps/libresign/lib/Service/PdfParserService.php#83
Impossible get metadata from this file: Error to get page width and height. If possible, open an issue at github.com/libresign/libresign with the file that you used._

Before the test I have only made the installation of all the dipendencies (in the libresign administration page all the items are greeen) and record the root certification data.

Could you help me?

ktrls:
Here it is. Thanks!

[Nextcloud Manual (copia).pdf](https://github.com/user-attachments/files/18520973/Nextcloud.Manual.copia.pdf)

vitormattos:
This is an edge case that when the package to parse the PDF file don't identify the MediaBox to get the document dimension.

https://github.com/smalot/pdfparser/pull/700

vitormattos:
I used the command pdfinfo of poppler-utils as fallback to solve this issue.

vitormattos:
Could you check again after next release of LibreSign? Only wait to be released.

Closing this issue as solved.

If this issue persists, don't hesitate to open a new issue making reference to this.

PR #4459, opened 2025-01-24
chore: use fallback to get page dimension

vitormattos:
Related with:
- https://github.com/LibreSign/libresign/issues/4455


vitormattos:
The solution using smalot/pdfparser didn't worked fine and was replaced by pdfinfo from poppler utils as fallback.

PR #4798, opened 2025-04-09
chore: replace deprecated properties

vitormattos:
Follow up of:
https://github.com/nextcloud-libraries/nextcloud-vue/releases/tag/v8.24.0


samuelsonbrito (approved):
Looks good to me. Props updated according to the current standard.

Issue #2780, opened 2024-04-16
"Next" button should not appear for files in "available for signature" state


DazeEnd:
**Is your feature request related to a problem? Please describe.**
Currently, a "Next" button is displayed for files that are in the "available for signature" state. This "Next" button opens the user interface for placing signature fields.

This "Next" button could be confusing for users (it is certainly confusing to me), because the user is being prompted to continue a process by taking a "Next" action, when they are not in the middle of a process. In fact they probably just finished a process by sending the file to someone else for a signature.

**Describe the solution you'd like**
When a file is in the "available for signature" state, the "Next" button should not be displayed unless the user adds an additional signer by clicking the "Add signer" button. Ideally a message reading "Waiting for signature" would be displayed where the "Next" button is currently displayed.

**Describe alternatives you've considered**
Instead of a "Waiting for signature" message, you could replace the "Next" button with a "Cancel" button that cancels the signature request.

**Additional context**
Current UI as of v8.0.0-rc11:
![Screenshot from 2024-04-16 14-12-13](https://github.com/LibreSign/libresign/assets/2699247/4af53e0e-78fb-4318-9bcf-a461349f7653)
![Screenshot from 2024-04-16 14-13-21](https://github.com/LibreSign/libresign/assets/2699247/df7d241e-1647-45b9-9e64-76be5a939bb5)

Mock-up of one possible solution:
![Mock Up](https://github.com/LibreSign/libresign/assets/2699247/1592c343-0c30-4954-b4ff-7d1e81dd1532)

vitormattos:
This is a good point about UX and also about new feature. I will talk more to be possible we think together and found a good solution.

## Next button

The objective of Next button is to go to step to define the position of visible signature and after this, send the file to be signed changing the status from "_draft_" to "_available for signature_".

The button do two actions and maybe this could be a problem because we will need to think more about how to split if we identify that is necessary split the next button.

## Cancel button

Other point is the action to cancel the signature. We will need to think more about this.... for now, if you open the tree dots menu on each signer, you will see an option to delete the signer. If you delete all, don't will remove the entry from table `libresign_file`. I think that `cancel` sound more to delete the entry from `libresign_file` table, that is the action that will remove the file from LibreSign and turn again to be a simple PDF file at Nextcloud. Have more points to think about to do this:
- When we delete the file `signed.pdf` at app files, will trigger a LibreSign job to delete all data about this file at LibreSign tables removing completely the file from LibreSign.
[...]

PR #5126, opened 2025-06-11
chore: edit visible signatures

samuelsonbrito:
### Pull Request Description
image

### Related Issue

Issue Number: #2780

### Pull Request Type

- Feature

### DIFF
--- a/src/Components/RightSidebar/RequestSignatureTab.vue
+++ b/src/Components/RightSidebar/RequestSignatureTab.vue
@@ -125,6 +125,10 @@ export default {
 	},
 	computed: {
 		labelOfSaveButton() {
+
+			if (this.filesStore.canSign()) {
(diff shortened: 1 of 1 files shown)


vitormattos:
@DazeEnd what do you think?

DazeEnd:
Sorry for the delay in getting back to you guys on this. I was traveling all last week, and I'm just now getting caught up.

I have not tested this change in action, but I think it looks good based on the screenshots above. This solution removes that "Next" button which I (and my users) found confusing, and replaces it with buttons for more concrete actions. I think this UI is definitely an improvement.

vitormattos:
Exactly this, the real is that the button receive a name using a condition.

vitormattos:
Can we go ahead and merge this to become at the next release?

Issue #1967, opened 2023-11-23
Improvements at response from API when generate root certificate


vitormattos:
## Scenario
When we don't fill the field "`Name (CN)`" and click at the button "`Generate root certificate`" we will see the message:
![Screenshot_20231123_144626](https://github.com/LibreSign/libresign/assets/1079143/4c0e5c52-0d89-4d22-86fa-c27623b990a0)

We need to see the message:
> Name (CN) is required

## Tips
Search by '`Name (CN)`' in the code and search by '`parameter '{$key}' is required`'. You will see that the array with the translated name stay at frontend and the text in backend. You also will found parameters here: `apps-extra/libresign/src/helpers/certification.js`

## To-do

Will be necessary move the parameters names to backend because the validation occur at API side. To do this you will need to move the file `apps-extra/libresign/src/helpers/certification.js` to be a helper in API side and return this as JSON in an initial state to frontend.

PR #5198, opened 2025-07-04
chore: Improvements at response from API when generate root certificate

Any97Cris:
### Related Issue

Issue Number: #1967 

### Pull Request Type

Chore

### DIFF
--- a/tests/php/Unit/Service/Certificate/RulesServiceTest.php
+++ b/tests/php/Unit/Service/Certificate/RulesServiceTest.php
@@ -77,10 +77,9 @@ final class RulesServiceTest extends \OCA\Libresign\Tests\Unit\TestCase {
 		$this->assertArrayHasKey('helperText', $actual);
 		unset($actual['helperText']);
 
-		foreach ($actual as $fieldCode => $value) {
-			$this->assertArrayHasKey($fieldCode, $expected);
-			$this->assertSame($expected[$fieldCode], $value, "Mismatch for field: $fieldCode");
-		}
+		$this->assertCount(count($expected), $actual, "Expected count does not match actual count for field: $fieldName");
+
+		$this->assertSame($expected, $actual, "Mismatch for field: $fieldName");
 	}
 
 	public static function providerGetRuleToValidField(): array {


tests/php/Unit/Service/Certificate/ValidateServiceTest.php
```diff
@@ -8,36 +8,107 @@
...
-		$this->expectExceptionMessage("Parameter 'CN' should be betweeen 1 and 64.");
-		$service->validate('CN', str_repeat('a', 65));
+	public static function providerValidInputs(): array {
+		return [
+			['CN', 'John Doe'],       // requerido, dentro do limite
+			['C', 'BR'],              // exatamente 2 letras
+			['ST', 'Amazonas'],       // válido
+			['L', 'Manaus'],          // válido
+			['O', 'LibreCode'],       // válido
+			['OU', 'Development'],    // válido
```
vitormattos: Only English, but all commentaries are unnecessary because it's already to test valid inputs.

tests/php/Unit/Service/Certificate/ValidateServiceTest.php
```diff
@@ -8,36 +8,107 @@
...
+			// CN é obrigatório → vazio deve falhar
+			['CN', '', "Parameter 'CN' is required!"],
+			// CN muito longo
+			['CN', str_repeat('a', 65), "Parameter 'CN' should be betweeen 1 and 64."],
+			// C muito curto
+			['C', 'B', "Parameter 'C' should be betweeen 2 and 2."],
+			// C muito longo
+			['C', 'BRA', "Parameter 'C' should be betweeen 2 and 2."],
+			// ST acima do limite
+			['ST', str_repeat('x', 129), "Parameter 'ST' should be betweeen 1 and 128."],
```
vitormattos: The error message already explains what's happening. Isn't necessary to have comments.

tests/php/Unit/Service/Certificate/ValidateServiceTest.php
```diff
@@ -8,36 +8,107 @@
...
+		$service = $this->getService();
+
+		$names = [
+			['id' => 'C', 'value' => 'BRA'], // inválido, deve ter 2 chars
+		];
+
+		$this->expectException(InvalidArgumentException::class);
+		$this->expectExceptionMessage("Parameter 'C' should be betweeen 2 and 2.");
+
+		$service->validateNames($names);
```
vitormattos: We can have a dataprovider with less and more than 2 chars.

lib/Service/Certificate/RulesService.php
```diff
@@ -0,0 +1,70 @@
...
+	];
+
+	public function __construct(
+		protected IL10N $l10n,
+	) {
+
+	}
+
+	public function getRule(string $fieldName): array {
+		if (!isset($this->rules[$fieldName]['helperText'])) {
```
vitormattos, line 50: Would be good to have a logic to verify if have a rule with "fieldName"

```suggestion
		if (!array_key_exists($fieldName, $this->rules)) return [];
		if (!isset($this->rules[$fieldName]['helperText'])) {
```

lib/Service/Certificate/RulesService.php
```diff
@@ -0,0 +1,70 @@
...
+		'OU' => [
+			'min' => 1,
+			'max' => 64,
+		],
+	];
+
+	public function __construct(
+		protected IL10N $l10n,
+	) {
+
```
vitormattos, line 46:
```suggestion
```

lib/Service/Certificate/ValidateService.php
```diff
@@ -0,0 +1,48 @@
...
+
+	public function __construct(
+		protected RulesService $rulesService,
+		protected IL10N $l10n,
+	) {
+
+	}
+
+	public function validate(string $fieldName, string $value):void {
+		$rule = $this->rulesService->getRule($fieldName);
```
vitormattos, line 24: Would be good to verify if the return is empty, and if is empty: throw new InvalidArgumentException('Parameter id is invalid');

Making this change will be necessary implement a test with a dataprovider containing invalid id values as example:

''
null
'AA'
'BBBB',
'A'

tests/php/Unit/Service/Certificate/ValidateServiceTest.php
```diff
@@ -0,0 +1,114 @@
...
+		$this->expectExceptionMessage('Parameter id is required!');
+
+		$service->validateNames($names);
+	}
+
+	public function testValidateNamesWithInvalidValueShouldFail(): void {
+		$service = $this->getService();
+
+		$names = [
+			['id' => 'C', 'value' => 'BRA'], // inválido, deve ter 2 chars
```
vitormattos: Unnecessary comment.

tests/php/Unit/Service/Certificate/ValidateServiceTest.php
```diff
@@ -0,0 +1,114 @@
...
+		$service->validateNames($names);
+
+		$this->assertTrue(true);
+	}
+
+	public function testValidateNamesWithoutIdShouldFail(): void {
+		$service = $this->getService();
+
+		$names = [
+			['id' => '', 'value' => 'Invalid Name'],
```
vitormattos: Would be good to be a dataprovider like as the follow change. This test is to verify if the ID is invalid, the string "Invalid Name" could generate an ambiguous meaning.

```suggestion
			['id' => '', 'value' => 'Name'],
			['value' => 'Name'],
```

tests/php/Unit/Service/Certificate/ValidateServiceTest.php
```diff
@@ -0,0 +1,114 @@
...
+			$rulesService,
+			$this->l10n
+		);
+	}
+
+	#[DataProvider('providerValidInputs')]
+	public function testValidateWithValidInput(string $fieldName, string $value): void {
+		$service = $this->getService();
+		$service->validate($fieldName, $value);
+		$this->assertTrue(true); // se não lançar exceção, passou
```
vitormattos: Without comments

tests/php/Unit/Service/Certificate/ValidateServiceTest.php
Any97Cris: In this case, it is necessary to create one array. Like this:
$arrayExample [
['id' => 'C', 'value' => ''],
['id' => 'C', 'value' => 'BRA'],
['id' => 'C', 'value' => 'BRAA'],
];

This afimative is not clare to me.

lib/Service/Certificate/RulesService.php
Any97Cris: you mean like this way:
```
public function getRule(string $fieldName): array {
		if (!array_key_exists($fieldName, $this->rules)) return [];
		if (empty($this->rules[$fieldName]['helperText'])) {
			unset($this->rules[$fieldName]['helperText']);
		}
		
		return $this->rules[$fieldName];
	}
```

tests/php/Unit/Service/Certificate/ValidateServiceTest.php
vitormattos: $arrayExample [
['id' => 'C', 'value' => ''],
['id' => 'C', 'value' => 'B'],
['id' => 'C', 'value' => 'BRA'],
['id' => 'C', 'value' => 'BRAA'],
];

lib/Service/Certificate/RulesService.php
vitormattos: What do you think about this suggestion?

Any97Cris: It will broke, because I tested and return error.
So I not sure about this change.

Any97Cris: So my interpretation about this modification was wrong. This modification is like this:
```
public function getRule(string $fieldName): array {
		if (!array_key_exists($fieldName, $this->rules)) return [];
		if (!isset($this->rules[$fieldName]['helperText'])) {
			$this->rules[$fieldName]['helperText'] = $this->getHelperText($fieldName);
			if (empty($this->rules[$fieldName]['helperText'])) {
				unset($this->rules[$fieldName]['helperText']);
			}
		}
		return $this->rules[$fieldName];
	}
```

vitormattos: Sounds good.

lib/Service/Certificate/RulesService.php
```diff
@@ -59,12 +59,40 @@ public function getRule(string $fieldName): array {
...
+			'C' => $tcertificate/his->l10n->t('Two-letter ISO 3166 country code'),
 			'ST' => $this->l10n->t('Full name of states or provinces'),
 			'L' => $this->l10n->t('Name of a locality or place, such as a city, county, or other geographic region'),
 			'O' => $this->l10n->t('Name of an organization'),
 			'OU' => $this->l10n->t('Name of an organizational unit'),
 			default => null,
 		};
 	}
+
+	public function getAllRules(): array {
```
vitormattos:
```suggestion
	public function toArray(): array {
```

lib/Service/Certificate/RulesService.php
```diff
@@ -59,12 +59,40 @@ public function getRule(string $fieldName): array {
 	public function getHelperText(string $fieldName): ?string {
 		return match ($fieldName) {
 			'CN' => $this->l10n->t('Common Name (CN)'),
-			'C' => $this->l10n->t('Two-letter ISO 3166 country code'),
+			'C' => $tcertificate/his->l10n->t('Two-letter ISO 3166 country code'),
```
vitormattos:
```suggestion
			'C' => $this->l10n->t('Two-letter ISO 3166 country code'),
```

tests/php/Unit/Service/Certificate/RulesServiceTest.php
```diff
@@ -0,0 +1,148 @@
...
+	#[DataProvider('providerGetHelperTextWithInvalidFieldName')]
+	public function testGetHelperTextWithInvalidFieldName(string $fieldName): void {
+		$service = $this->getService();
+		$helperText = $service->getHelperText($fieldName);
+		$this->assertNull($helperText);
+	}
+
+	public static function providerGetHelperTextWithInvalidFieldName(): array {
+		return [
+			['INVALID'],
```
vitormattos, line 58:
```suggestion
			[null],
			['INVALID'],
```

tests/php/Unit/Service/Certificate/RulesServiceTest.php
```diff
@@ -0,0 +1,148 @@
...
+	#[DataProvider('providerGetRuleToInvalidField')]
+	public function testGetRuleToInvalidField(string $fieldName): void {
+		$service = $this->getService();
+		$actual = $service->getRule($fieldName);
+		$this->assertEmpty($actual);
+	}
+
+	public static function providerGetRuleToInvalidField(): array {
+		return [
+			['INVALID'],
```
vitormattos, line 136:
```suggestion
			[null],
			['INVALID'],
```

lib/Controller/AdminController.php
```diff
@@ -136,15 +138,14 @@ private function generateCertificate(
 	): IEngineHandler {
 		$names = [];
 		if (isset($rootCert['names'])) {
+			$this->validateService->validateNames($rootCert['names']);
 			foreach ($rootCert['names'] as $item) {
-				if (empty($item['id'])) {
-					throw new LibresignException('Parameter id is required!', 400);
-				}
-				$names[$item['id']]['value'] = $this->trimAndThrowIfEmpty($item['id'], $item['value']);
+				$names[$item['id']]['value'] = trim((string)$item['value']);
```
AntDavidLima, line 143: It may be good to rewrite this line in multiple lines, to make it easier to read. Something like:
```php
$itemId = $item['id'];
$itemValue = strval($item['value']);

$names[$itemId]['value'] = trim($itemValue);
```

It's something silly, but may help people new to the application to understand it's code.

lib/Service/Certificate/ValidateService.php
```diff
@@ -0,0 +1,57 @@
...
+		if (!$length && isset($rule['required']) && $rule['required']) {
+			throw new InvalidArgumentException(
+				$this->l10n->t("Parameter '%s' is required!", [$fieldName])
+			);
+		}
+
+		if ($length > $rule['max'] || $length < $rule['min']) {
+			throw new InvalidArgumentException(
+				$this->l10n->t(
+					"Parameter '%s' should be between %s and %s.",
```
AntDavidLima: `Parameter '%s' should be between %s and %s characters.` will be clearer, I think.

AntDavidLima (changes requested):
Everything looks right to me. Great job!
I've just requested some changes related to some minor, not that important things.

lib/Service/Certificate/ValidateService.php
vitormattos: `characters` only make sense when is about a string. The interval is numeric.

lib/Controller/AdminController.php
vitormattos: Isn't this too much vars?

vitormattos:
I only made a rebase without any change at commits of this PR

lib/Controller/AdminController.php
vitormattos: Maybe would be best create a fallback issue to implement a `NormalizeService` class to add the trim rule into all values and also add an uppercase rule into Country and maintain the scope of this PR more amalgamated with the issue that motivated this PR.

vitormattos (approved):

Issue #4160, opened 2024-12-16
Add status to LibreSign document at app Files


Extarys:
**Is your feature request related to a problem? Please describe.**
I created a test PDF `Test signature.pdf` for the admin to sign, once signed, a new file `Test signature.signed.pdf` was created.
But now I have two files: `Test signature.pdf` and `Test signature.signed.pdf` and both shows LibreSign panel on the right saying the admin signed it.

![image](https://github.com/user-attachments/assets/87f25ae4-b4a2-42ac-81d5-1197e0dca8a6)

If I removed the "signed" from the name, there is no way to show which document was signed precisely. Yes I can check the file size, but I should not have to.

**Describe the solution you'd like**
- Show something near the file icon to show that this file contains a digital signature from Libresign (this might be nextcloud upstream to show that the file contain _a_ digital signature ?)
- Update the sidebar
   - For the original document: Show who requested the signature of that document and if/when it was signed
   - For the signed document: Show the digital signature/metadata of that file + what's above (For the original document) _Clearly_ state that this document is the one containing a signature. 

It would also be nice to have more instructions like : once signed, the original document can be safely deleted if not needed.
Also, a text for the modal when we can create signature boxes to explain how it works (or a ? icon). I didn't know and found out while looking at issues here.
  

**Describe alternatives you've considered**

**Additional context**

vitormattos:
Hi @Extarys,

Thanks a lot by your considerations. Follow my points about your considerations.

## Add status to LibreSign documents

Maybe this issue could be renamed to this title and go ahead with this discussion. Could you rename this issue to this (Add status to LibreSign documents)?

> If I removed the "signed" from the name, there is no way to show which document was signed precisely. Yes I can check the file size, but I should not have to.

The signed document is the document that have the suffix: signed.pdf. And yes, if you rename this, don't will be possible identify by a easy way if this document was signed or not.

> Show something near the file icon to show that this file contains a digital signature from Libresign (this might be nextcloud upstream to show that the file contain a digital signature ?)

We can put an icon together of file icon but what kind of icon could represent that this document was digitally signed? Other question to think about is that LibreSign have other status and will we also use this? Follow below a screenshot of status list of LibreSign took from filter of LibreSign (go to LibreSign, Files and will be possible see the filters):

![Screenshot_20241217_104948](https://github.com/user-attachments/assets/b29a938a-dec3-4e8d-a37a-823fe3a55a61)

[...]

PR #5225, opened 2025-07-21
feat: show signature status on files app

vitormattos:
### Pull Request Description
This pull request adds an icon to the right of the files, in the files app, showing the file signature status on the original and the signed files. The original will have the status 'Original file signed elsewhere' and the signed file will have the status 'Partially signed' and 'Signed', depending on the signing progress.

| Before: | After: |
|--------|--------|
| before | after |

To test the feature, after checking out to the pull request:
- upload a PDF file to Nextcloud using the files app;
- request the signature from two different users;
- sign the file with one of the requested users;
- the original file will show a green sign icon, with the text 'original file' when hovering over it;
- the signed file will show a yellow ½ icon, with the text 'partially signed' when hovering over it;
- now sign the file with the second requested user;
- the signed file will show a blue sign icon, with the text 'signed' when hovering over it.

### Related Issue
Issue Number: #4160 

### Pull Request Type
- Feature

### Pull request checklist


lib/Dav/SignatureStatusPlugin.php
```diff
@@ -0,0 +1,39 @@
...
+	public function propFind(PropFind $propFind, INode $node): void {
+		if ($node instanceof File) {
+			$fileService = OC::$server->get(FileService::class);
+			$nodeId = $node->getId();
+
+			if ($fileService->isLibresignFile($nodeId)) {
+				$fileService->setFileByType('FileId', $nodeId);
+
+				$propFind->handle('{http://nextcloud.org/ns}signature-status', $fileService->getStatus());
+				$propFind->handle('{http://nextcloud.org/ns}signed-node-id', $fileService->getSignedNodeId());
```
vitormattos: Whats do you think to prefix this with `libresign-` ?
```suggestion
				$propFind->handle('{http://nextcloud.org/ns}libresign-status', $fileService->getStatus());
				$propFind->handle('{http://nextcloud.org/ns}libresign-signed-node-id', $fileService->getSignedNodeId());
```

lib/Db/FileMapper.php
```diff
@@ -167,6 +167,30 @@ public function getByFileId(?int $nodeId = null): File {
 		return $file;
 	}
 
+	/**
+	 * Check if file exists
+	 */
```
vitormattos: The method name already explain this.
```suggestion
```

lib/Db/FileMapper.php
```diff
@@ -167,6 +167,30 @@ public function getByFileId(?int $nodeId = null): File {
...
+	public function fileIdExists(int $nodeId): bool {
+		$exists = array_filter($this->file, fn ($f) => $f->getNodeId() === $nodeId || $f->getSignedNodeId() === $nodeId);
+		if (!empty($exists)) {
+			return true;
+		}
```
vitormattos, line 174: If it's a lazy loading, you didn't stored the result of query at file property on this class.

lib/Dav/SignatureStatusPlugin.php
```diff
@@ -0,0 +1,39 @@
...
+class SignatureStatusPlugin extends ServerPlugin {
+	protected $server;
+
+	public function initialize(Server $server): void {
+		$this->server = $server;
+		$server->on('propFind', [$this, 'propFind']);
+	}
+
+	public function propFind(PropFind $propFind, INode $node): void {
+		if ($node instanceof File) {
```
vitormattos: What do  you think about flip this condition to return early and after the condition, go ahead with the code?
```suggestion
		if (!$node instanceof File) {
			return;
		}
```
Advantage: Will reduce a bit the row size of this entire method.

lib/Dav/SignatureStatusPlugin.php
vitormattos: Isn't about webdav plugins but here have a convention about naming:
https://docs.nextcloud.com/server/latest/developer_manual/basics/storage/configuration.html#naming-conventions

PR #5230, opened 2025-07-21
fix: store the date that the file was signed

vitormattos:
### Pull Request Description

The sign engine is used to sign the file and between the interval that the file was signed and the moment to save this at the database, is possible to be different. To prevent issues about different dates (seconds or milisseconds), I made a parse at signed file and after then, I fetch the last signature date to store this at signed field at the database.

The unit test that was changed don't validate this scenario because this test have a lot of mocks. I didn't implemented more tests because to sign a document will be necessary use the PKCS12 engine and to this will be necessary download a lot of dependencies making the unit test more slow.

Maybe would be good think more about how to implement more tests to cover this change.

### Pull request checklist

### DIFF
--- a/lib/Handler/SignEngine/SignEngineHandler.php
+++ b/lib/Handler/SignEngine/SignEngineHandler.php
@@ -18,6 +18,8 @@ use OCA\Libresign\Handler\CertificateEngine\IEngineHandler;
 use OCA\Libresign\Service\FolderService;
 use OCP\Files\File;
 use OCP\Files\GenericFileException;
+use OCP\Files\InvalidPathException;
+use OCP\Files\NotFoundException;
 use OCP\Files\NotPermittedException;
 use OCP\IL10N;
 
@@ -165,6 +167,7 @@ abstract class SignEngineHandler implements ISignEngineHandler {
 			$file->delete();
 		} catch (NotPermittedException) {
 			throw new LibresignException($this->l10n->t('You do not have permission for this action.'));
+		} catch (NotFoundException|InvalidPathException) {
 		}
 	}
 


lib/Service/SignFileService.php
```diff
@@ -367,6 +367,13 @@ public function sign(): File {
 
 		return $signedFile;
 	}
+
+	private function getLastSignedDdate(File $signedFile): \DateTime {
```
AntDavidLima: The method name part 'Ddate' was a typo?
If yes, it would be good to fix it now, before it is used in other places.

AntDavidLima (approved):

lib/Handler/SignEngine/Pkcs7Handler.php
```diff
@@ -28,15 +26,25 @@ public function sign(): File {
...
+	public function getCertificateChain($resource): array {
+		$metadata = stream_get_meta_data($resource);
+		$lastModifiedTime = filemtime($metadata['uri']);
+		return [
+			[
+				'signingTime' => (new DateTime())->setTimestamp($lastModifiedTime),
+			],
+		];
+	}
+
```
AntDavidLima, line 49: Since the certificate chain is not required yet, why name this method getCertificateChain instead of getFileLastModificationTime?
I believe the use of such method is planned on the future, but I think prematurely creating it, without a actual implementation, it's not a good practice. In case something changes and it is not needed anymore in the future, the method name will remain as a confusing deadcode.

lib/Handler/SignEngine/SignEngineHandler.php
```diff
@@ -90,4 +104,146 @@ public function setSignatureParams(array $params): self {
...
+		$folder = $this->folderService->getFolder();
+		if ($folder->nodeExists($this->pfxFilename)) {
+			$file = $folder->get($this->pfxFilename);
+			if (!$file instanceof File) {
+				throw new LibresignException("path {$this->pfxFilename} already exists and is not a file!", 400);
+			}
+			try {
+				$file->putContent($content);
+			} catch (GenericFileException) {
+				throw new LibresignException("path {$file->getPath()} does not exists!", 400);
```
AntDavidLima, line 152: If $folder->nodeExists is true, a path does not exists exception can really be caught here? Maybe this error message for a GenericFileExecption could be misleading.

lib/Handler/SignEngine/SignEngineHandler.php
```diff
@@ -90,4 +104,146 @@ public function setSignatureParams(array $params): self {
...
+	}
+
+	public function deletePfx(string $uid): void {
+		$this->folderService->setUserId($uid);
+		$folder = $this->folderService->getFolder();
+		try {
+			$file = $folder->get($this->pfxFilename);
+			$file->delete();
+		} catch (\Throwable) {
+		}
```
AntDavidLima: Isn't dangerous to no treat this exception correctly?

lib/Handler/SignEngine/SignEngineHandler.php
```diff
@@ -90,4 +104,146 @@ public function setSignatureParams(array $params): self {
...
+	}
+
+	/**
+	 * Get content of pfx file
+	 */
+	public function getPfxOfCurrentSigner(?string $uid = null): string {
+		if (!empty($this->certificate) || empty($uid)) {
+			return $this->certificate;
+		}
```
AntDavidLima, line 180: If $uid is empty, return $this->certificate even if $this->certificate is empty is the expected output?

lib/Handler/SignEngine/Pkcs7Handler.php
vitormattos: > why name this method getCertificateChain instead of getFileLastModificationTime

We have an interface that require this implementation.

I added the flag `@todo` at this method because we need to identify a best way to sign a non PDF file.

For now, LibreSign is generating only a p7s file, but to each signer is necessary to have a p7s file. Looking the RFC about this, I identified that could be possible implement a way to add multiple signatures at the same p7s file.

I didn't implemented a definitive way to this because LibreSign haven't support to sign non PDF file. This implementation is only a POC and haven't implementation at interface to support this and I also don't know if the API endpoints will work with non PDF file.

Only to have the same output signature that is required by interface, I returned the modification time of p7s file.

lib/Handler/SignEngine/SignEngineHandler.php
vitormattos: NodeExists will do a request to a file at storage, and after this, the putContent will do the same.

To reduce the IO, is a good practice to use `try putContent cat` at Nextcloud apps .

lib/Handler/SignEngine/SignEngineHandler.php
vitormattos: I made this because don't matter if we will really delete the pdx file, it's only to reduce the storage.

This method could throw the follow:

InvalidPathException
NotFoundException
NotPermittedException

If the path is invalid, have nothing to do, we can consider that the file don't exists
If the file wasn't found, don't matter, we want this

Maybe the unique possible problem is:

NotPermittedException

I changed to catch this now.

lib/Handler/SignEngine/SignEngineHandler.php
vitormattos: Conditions under which the value may be empty:

- When signing a file using the "click to sign" method: in this case, the PFX is empty.

- When the signer is not authenticated: the UID will be empty, this method will return an empty string because haven't a pfx file and a new certificate will need to be generated.

We have not implemented support for allowing unauthenticated signers to upload a PFX file.
If the signer is unauthenticated, a new certificate must be created, and this method will return an empty string.

PR #5316, opened 2025-08-13
fix: configure check with poppler

vitormattos:
### Pull Request Description

Repoened from:
- #5280
- 
### Related Issue

Issue Number:
- https://github.com/LibreSign/libresign/issues/5259

### Pull request checklist

- [x] Did you explain or provide a way of how can we test your code ?
- [x] If your pull request is related to frontend modifications provide a print of before and after screen
- [x] Did you provide a general summary of your changes ?
- [x] Try to limit your pull request to one type, submit multiple pull requests if needed
- [x] I implemented tests that cover my contribution

PR #5317, opened 2025-08-13
Fix ConfigurationCheck failure on poppler-utils

ASerbinski:
By default, PHP-FPM clears all environment variables in workers. The result of this is that the "which" command, previously used to determine that pdfsig and pdfinfo commands are available on the system would return NULL instead of the path to the commands.

We solve this by instead simply executing the command "pdfinfo -v" or "pdfsig -v" and checking the return value. A return value of anything besides "0" indicates that the command failed.

### Related Issue

Issue Number:

### Pull request checklist

- [ ] Did you explain or provide a way of how can we test your code ?
- [ ] If your pull request is related to frontend modifications provide a print of before and after screen
- [ ] Did you provide a general summary of your changes ?
- [ ] Try to limit your pull request to one type, submit multiple pull requests if needed

### DIFF
--- a/lib/Service/Install/ConfigureCheckService.php
+++ b/lib/Service/Install/ConfigureCheckService.php
@@ -78,7 +78,10 @@ class ConfigureCheckService {
 		if (!empty($this->result['poppler'])) {
 			return $this->result['poppler'];
 		}
-		if (shell_exec('which pdfsig') === null) {
+		// The output of this command go to STDERR and exec get the STDOUT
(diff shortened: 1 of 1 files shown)


vitormattos:
Could you go ahead at https://github.com/LibreSign/libresign/pull/5316 as replacement of your PR to simplify the process a bit?

vitormattos:
closed by: https://github.com/LibreSign/libresign/pull/5316

PR #5327, opened 2025-08-16
chore: remove unecessary check if node exists

vitormattos:
### Pull Request Description

The method get already do this and throw an exception when the file don't exists.

### Related Issue

- https://github.com/LibreSign/libresign/issues/5321-

### Pull Request Type

- Refactoring (no functional changes, no api changes)

### Pull request checklist

- [x] I implemented tests that cover my contribution


lib/Service/AccountService.php
```diff
@@ -252,7 +251,7 @@ public function hasSignatureFile(?IUser $user = null): bool {
 		try {
 			$this->pkcs12Handler->getPfxOfCurrentSigner($user->getUID());
 			return true;
-		} catch (\Throwable) {
+		} catch (LibresignException) {
```
AntDavidLima, line 254: Can you refresh my memory on why should we catch an exception and do nothing with that?

In this case, wouldn't be clerar to return false on the catch block?

lib/Controller/SignatureElementsController.php
```diff
@@ -160,7 +160,7 @@ public function getSignatureElements(): DataResponse {
 	#[ApiRoute(verb: 'GET', url: '/api/{apiVersion}/signature/elements/preview/{nodeId}', requirements: ['apiVersion' => '(v1)'])]
 	public function getSignatureElementPreview(int $nodeId) {
 		try {
-			$node = $this->accountService->getFileByNodeIdAndSessionId(
+			$node = $this->accountService->getFileByNodeId(
 				$nodeId,
 				$this->sessionService->getSessionId()
```
AntDavidLima: The method signature now only receives one argument.

AntDavidLima (changes requested):

lib/Service/AccountService.php
vitormattos: When we haven't a pfx file, the get method will throw an exception but this method need to return false when we haven't a pfx file.

vitormattos:
The broke tests aren't about LibreSign, already was fixed at guests app.

```json
{
	"app": "PHP",
	"message": "Declaration of OCA\\Guests\\AppConfigOverwrite::getValue(string $app, string $key, ?string $default = null): string must be compatible with OC\\AppConfig::getValue($app, $key, $default = '') at \/var\/www\/html\/apps-extra\/guests\/lib\/AppConfigOverwrite.php#22"
}
```

vitormattos:
> The broke tests aren't about LibreSign, already was fixed at guests app.
> 
> ```json
> {
> 	"app": "PHP",
> 	"message": "Declaration of OCA\\Guests\\AppConfigOverwrite::getValue(string $app, string $key, ?string $default = null): string must be compatible with OC\\AppConfig::getValue($app, $key, $default = '') at \/var\/www\/html\/apps-extra\/guests\/lib\/AppConfigOverwrite.php#22"
> }
> ```

Fixed, was created a new release with the fix at app Guests and I made a rebase at this branch to be possible restart the CI.

After your last review I only made this commit:

https://github.com/LibreSign/libresign/pull/5327/commits/7a4f366ad1683366201e0f855c026bcd10c5f284

lib/Service/AccountService.php
AntDavidLima: Understood, but in this case, wouldn't be cleaner to return false on the catch block? It will make no difference, but I thought it would be more elegant.

AntDavidLima (approved):

PR #5331, opened 2025-08-19
chore: bump dependencies

vitormattos:


vitormattos:
Waiting the next release of app guests to fix the follow fatal error:

```json
{
	"app": "PHP",
	"message": "Declaration of OCA\\Guests\\AppConfigOverwrite::getValue(string $app, string $key, ?string $default = null): string must be compatible with OC\\AppConfig::getValue($app, $key, $default = '') at \/var\/www\/html\/apps-extra\/guests\/lib\/AppConfigOverwrite.php#22"
}
```

PR #5415, opened 2025-09-04
fix: prevent error when try to create a folder two times

vitormattos:
### Pull Request Description

The LibreSign folder is created by every event that need to fetch files from this folder. When we try to create a folder or a file, before to finish the creation proccess, the folder is locked by Nextcloud. To prevent this, I made first a get, if is not found, I try to create.

### Pull Request Type

- Bugfix

### Pull request checklist

<details>
<summary>How to see this running using GitHub Codespaces</summary>

### 1. Open the Codespace
- Authenticate to GitHub
- Go to the branch: [chore/reduce-configure-check-time](https://github.com/LibreSign/libresign/tree/chore/reduce-configure-check-time)
- Click the `Code` button and select the `Codespaces` tab.
- Click **"Create codespace on feat/customize-signature-stamp"**

### 2. Wait for the environment to start
- A progress bar will appear on the left.  
- After that, the terminal will show the build process.
- Wait until you see the message:  
  ```bash
  ✍️ LibreSign is up!
  ```
  This may take a few minutes.

### 3. Access LibreSign in the browser
- Open the **Ports** tab (next to the **Terminal**).
- Look for the service running on port **80**.
- Hover over the URL and click the **globe icon** 🌐 to open it in your browser.

### 4. (Optional) Make the service public
- If you want to share the app with people **not logged in to GitHub**, you must change the port visibility:
  - Click the three dots `⋮` on the row for port 80.
  - Select `Change visibility` → `Public`.

### 5. Login credentials
- **Username**: `admin`  
- **Password**: `admin`

Done! 🎉
You're now ready to test this.
</details>


AntDavidLima (commented):
Is there a way to test this with PHPUnit? If so, it would be good.

vitormattos:
Could be a bit complex because is out of scope of LibreSign but we can try.

AntDavidLima:
Understood. If possible, it would be good to have at least a test that proves that the bug beeing fixed here doen't happen anymore.

vitormattos:
I only reproduced the bug because the browser send two asynchronous requests to server and when we got two requests at the same time creating the same folder was possible to see the problem.

It's sounds more that will be necessary an integration test than unit because I think that also will be necessary to authenticate an user.

I think that at this case the best way is believe that it's OK, I don't know how to reproduce this easy into an unit test.

AntDavidLima (approved):
Understood, in this case, I too don't think it is worth implementing a integration test.

vitormattos:
Only by documentation purpose, follow a screenshot that reproduce the issue:

Screenshot_20250909_121355

This occur at the first time that we open a document form app Files, choose to see the document at LibreSign and click to sign the document.

After this fix, this flow worked fine.

PR #5418, opened 2025-09-08
Fixed #5347 issue.

### LINKED ISSUE #5347: failed signing pdf: Invalid signing Time in certificate chain
### Describe the bug

failed signing pdf: Invalid signingTime in certificate chain

### To reproduce

after upgrading to version I have this problem when I'm signing anything, my system and php timezone didn't change (Eastern Europe) , but libresign see that there is UTC time and this window pups up:

Image

### Expected behavior

signed document

### Screenshots

_No response_

### Environment information

Debian
Apache2
php8.2

### Additional context

_No response_

oktay454:
Required `occ config:app:set libresign server_timezone --value="YOUR_SERVER_TIMEZONE"` command

### Related Issue
Issue Number: #5347

### Pull Request Type
- Bugfix

### DIFF
--- a/lib/Handler/SignEngine/Pkcs12Handler.php
+++ b/lib/Handler/SignEngine/Pkcs12Handler.php
@@ -172,7 +172,9 @@ class Pkcs12Handler extends SignEngineHandler {
 			if ($isSecondLevel) {
 				switch ((string)$match['key']) {
 					case 'Signing Time':
-						$this->signaturesFromPoppler[$lastSignature]['signingTime'] = DateTime::createFromFormat('M d Y H:i:s', $match['value']);
+						$timezoneString = $this->appConfig->getValueString('libresign', 'server_timezone', 'UTC');
+						$serverTimeZone = new \DateTimeZone($timezoneString);
+						$this->signaturesFromPoppler[$lastSignature]['signingTime'] = DateTime::createFromFormat('M d Y H:i:s', $match['value'], $serverTimeZone);
 						break;
 					case 'Signer full Distinguished Name':
 						$this->signaturesFromPoppler[$lastSignature]['chain'][0]['subject'] = $this->parseDistinguishedNameWithMultipleValues($match['value']);


vitormattos:
@oktay454 did you tested the follow PR?

- https://github.com/LibreSign/libresign/pull/5417

oktay454:
@vitormattos Yes, I tested it, but the time zone of my server is different from UTC. The result of the pdfsig command shows the local time zone. Therefore the problem is not fixed.

vitormattos:
Are you on the latest LibreSign? In the latest release we standardized all **server-side** `DateTime` handling to **UTC** (see PR #5396). It looks like this particular code path wasn’t covered there.

In your PR you reference a system setting named `server_timezone`. That setting doesn’t exist in Nextcloud (see this [code search](https://github.com/search?q=org%3Anextcloud%20server_timezone&type=code)). The closest is [`default_timezone`](https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/config_sample_php_parameters.html#default-timezone), which is for **user/client display**, not server-side logic.

For multi-timezone environments, server-side timestamps should be stored in **UTC** (including when stored as Unix epoch) and treated as timezone-agnostic. Clients are responsible for converting and displaying in their local timezone. That’s why we default to UTC—otherwise we’d need to persist the originating timezone on every field and branch logic accordingly.

Regarding the validation page: the backend sends timestamps in **UTC**; the frontend renders them using the **client’s** timezone. So the displayed time varies by client, but the underlying value is the same.

Going ahead, **requested changes:**

* Remove reliance on the non-existent `server_timezone`. A custom entry in your `config.php` would only work at your environment and isn’t portable.
* Keep server-side in UTC and convert only at the presentation layer. See the approach in the follow PR:
  #5417

If you’re on a version prior to the changes from #5396, that could explain the behavior you’re seeing that only worked locking the timezone as the same of your server and your client.

Happy to receive your comment. Would love to have you among the LibreSign contributors!

oktay454:
I am using LibreSign 11.3.2 on NextCloud 31.
For a different reason I cannot change the time zone on my server.
Already a variable like server_timezone is not standard. I defined it for LibreSign as a solution. I was able to run it without applying the changes in my PR, that is, with the original code, only with the following change.
When I changed the pdfsig and pdfinfo commands to `env TZ=UTC pdfXXX`, the problem was fixed.
https://github.com/LibreSign/libresign/blob/stable31/lib/Handler/SignEngine/Pkcs12Handler.php#L154
https://github.com/LibreSign/libresign/blob/stable31/lib/Service/PdfParserService.php#L121

oktay454:
Then I hope this amendment will be complementary to the solution you have implemented. #5424

vitormattos:
Hi, I think that now we can close this and don't merge because you already solved at the other PR.

PR #5472, opened 2025-09-24
chore: adjustment in the filters modified and status

Any97Cris:
Add button "Set maked filter" to fix the option selected the filter modified
image

Add button "Set maked filter" to fix the option selected the filter status
image

### DIFF
--- a/src/views/FilesList/FileListFilter/FileListFilter.vue
+++ b/src/views/FilesList/FileListFilter/FileListFilter.vue
@@ -18,10 +18,6 @@
 				@click="$emit('reset-filter')">
 				{{ t('files', 'Clear filter') }}
 			</NcActionButton>
-			<NcActionButton close-after-click
-				@click="$emit('set-marked-filter')">
-				{{ t('files', 'Save filter') }}
-			</NcActionButton>
 		</template>
 	</NcActions>
 </template>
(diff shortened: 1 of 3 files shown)


lib/Service/AccountService.php
```diff
@@ -233,12 +233,20 @@ public function getConfig(?IUser $user = null): array {
...
 		$info['isApprover'] = $this->validateHelper->userCanApproveValidationDocuments($user, false);
-		$info['grid_view'] = $this->getUserConfigGridView($user);
+		$info['grid_view'] = $this->getUserConfigByKey($user, 'grid_view') === '1';;
 
 		return $info;
 	}
 
+	/**
+	 * @return array[]
+	 */
```
vitormattos: The method already have a return type.
```suggestion
```

lib/Db/SignRequestMapper.php
```diff
@@ -551,13 +551,15 @@ private function getFilesAssociatedFilesWithMeQueryBuilder(string $userId, array
 				);
 			}
 			if (!empty($filter['start'])) {
+				$start = (new \DateTime('@'.$filter['start']))->format('Y-m-d H:i:s');
```
vitormattos: I think that will be necessary to add the timezone at both cases.

```suggestion
				$start = (new \DateTime('@' . $filter['start'], new \DateTimeZone('UTC')))->format('Y-m-d H:i:s');
```

PR #5475, opened 2025-09-30
chore: bump PHP dependencies

vitormattos:


vitormattos:
The integration tests broke because the app guests need to be updated to be compatible with Nextcloud 33. We can ignore this because will have a lot of time to this be fixed at app guests.

Issue #6027, opened 2025-12-08
Add Unified Search Provider for LibreSign


vitormattos:
### Is your feature request related to a problem? Please describe.

Currently, LibreSign does not appear in the results of Nextcloud's "Unified Search". It would be useful to allow LibreSign files to be found through the platform's unified search.

### Describe the solution you'd like

Implement a Search Provider for LibreSign, following the official Nextcloud documentation:  
https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/search.html

The provider should be registered in the main application class (`Application`) and return files managed by LibreSign, making them discoverable via Unified Search.

A good suggestion for the LibreSign search provider would be:

- **id:** libresign
- **name:** LibreSign

These follow the convention used by other Nextcloud apps and are clear, concise, and unique.

### Describe alternatives you've considered

- Keep the current behavior, without unified search integration.
- Use manual searches within LibreSign.

### Additional context

- If you are unsure how to structure the template, start by looking at the [example template](https://github.com/nextcloud/spreed/blob/main/.github/pull_request_template.md) and adapt it for LibreSign.
- For UI changes, you can use any image (even a placeholder) for before/after. Markdown tables are fine.
- If you have questions, feel free to ask in this issue.
- Give a ⭐️ star to this repository 

---

This task is suitable for first-time contributors!

PR #6243, opened 2025-12-20
feat: unified search

Any97Cris:
Issue: https://github.com/LibreSign/libresign/issues/6027

✨ Features

Global Search Integration: Documents can now be found using Nextcloud's unified search
Smart Filtering: Search by document name with automatic filtering
Rich Results: Display document thumbnails, file paths, and signature status
Status Labels: Show document status (Draft, Able to sign, Partially signed, Signed, Deleted)
Dynamic Icons: Automatically detect and display appropriate MIME type icons
Pagination Support: Handle large result sets efficiently

image

### DIFF
--- a/lib/Db/SignRequestMapper.php
+++ b/lib/Db/SignRequestMapper.php
@@ -566,11 +566,11 @@ class SignRequestMapper extends QBMapper {
 	}
 
 	private function getFilesAssociatedFilesWithMeQueryBuilder(
-			string $userId,
-			array $filter = [],
-			bool $count = false,
-			array $sort = []
-		): IQueryBuilder {
+		string $userId,
+		array $filter = [],
+		bool $count = false,
+		array $sort = [],
+	): IQueryBuilder {
 		$qb = $this->db->getQueryBuilder();
 		$qb->from('libresign_file', 'f')
 			->leftJoin('f', 'libresign_sign_request', 'sr', 'sr.file_id = f.id')


lib/Search/FileSearchProvider.php
```diff
@@ -0,0 +1,225 @@
...
+		}
+	}
+
+	/**
+	 * Get the translated label for a file status
+	 *
+	 * @param int|null $status File status code
+	 * @return string Translated status label
+	 */
+	private function getStatusLabel(?int $status): string {
```
vitormattos: We already have this enum to this:
`OCA\Libresign\Enum\FileStatus`

lib/Search/FileSearchProvider.php
```diff
@@ -0,0 +1,225 @@
...
+
+		return $searchResultEntry;
+	}
+
+	/**
+	 * Format the subline showing file location
+	 *
+	 * @param string $path File path
+	 * @return string Formatted subline text
+	 */
```
vitormattos: Isn't necessary to add docblock to private methods when the method name is clear, the same to other docblocks.

lib/Search/FileSearchProvider.php
```diff
@@ -0,0 +1,225 @@
...
+				$path = $userFolder->getRelativePath($node->getPath());
+				$subline = $this->formatSubline($path);
+			}
+		} catch (\Exception $e) {
+		}
+
+		$status = $this->getStatusLabel($file->getStatus());
+		if ($status) {
+			$subline .= $subline ? ' • ' . $status : $status;
+		}
```
vitormattos: I liked this but could have problems when the language is RTL and not LTR. I haven't suggestions to solve this, maybe could be removed for now.

lib/Search/FileSearchProvider.php
```diff
@@ -0,0 +1,225 @@
...
+		$path = '';
+
+		try {
+			$nodes = $userFolder->getById($file->getNodeId());
+			if (!empty($nodes)) {
+				$node = array_shift($nodes);
+
+				$icon = $node->getMimetype() === FileInfo::MIMETYPE_FOLDER
+					? 'icon-folder'
+					: $this->mimeTypeDetector->mimeTypeIcon($node->getMimetype());
```
vitormattos: LibreSign haven't folder.

lib/Search/FileSearchProvider.php
```diff
@@ -0,0 +1,225 @@
...
+				$file->setCreatedAt($row['created_at'] ?? '');
+				$files[] = $file;
+			} catch (\Exception $e) {
+				continue;
+			}
+		}
+
+		$result->closeCursor();
+
+		return $files;
```
vitormattos: We already have a method to make a query like this:

```
SignRequestMapper->getFilesAssociatedFilesWithMeQueryBuilder
```

But, is a private method for now. 

The public method that expose this is:

```
getFilesAssociatedFilesWithMeFormatted
```

But is too complex to use the `getFilesAssociatedFilesWithMeFormatted` method.

I recommend to create a wrapper to `getFilesAssociatedFilesWithMeStmt`, run the statement and return the rows to `FileSearchProvider`

Could be a method at `SignRequestMapper` called as `getFilesToSearchProvider`

lib/Search/FileSearchProvider.php
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	private function formatSubline(string $path): string {
+		if (strrpos($path, '/') > 0) {
+			$path = ltrim(dirname($path), '/');
+			return $this->l10n->t('in %s', [$path]);
+		} else {
+			return '';
+		}
+	}
```
vitormattos, line 138: Is this method necessary?

lib/Db/SignRequestMapper.php
```diff
@@ -416,6 +416,49 @@ public function getFilesAssociatedFilesWithMeFormatted(
...
+			try {
+				$file = new File();
+				$file->setId((int)$row['id']);
+				$file->setUserId($row['user_id']);
+				$file->setNodeId((int)($row['node_id'] ?? 0));
+				$file->setSignedNodeId($row['signed_node_id'] ? (int)$row['signed_node_id'] : null);
+				$file->setName($row['name'] ?? '');
+				$file->setStatus((int)($row['status'] ?? 0));
+				$file->setUuid($row['uuid'] ?? '');
+				$file->setCreatedAt($row['created_at'] ?? '');
```
vitormattos: I think that this also works fine.
```suggestion
				$file = new File();
				$file->fromRow($row);
```

lib/Db/SignRequestMapper.php
```diff
@@ -416,6 +416,49 @@ public function getFilesAssociatedFilesWithMeFormatted(
...
+			'page' => ($offset / $limit) + 1,
+			'length' => $limit,
+		];
+
+		$qb = $this->getFilesAssociatedFilesWithMeQueryBuilder($user->getUID(), $filter);
+
+		if (!empty($term)) {
+			$qb->andWhere(
+				$qb->expr()->like('f.name', $qb->createNamedParameter('%' . $this->db->escapeLikeParameter($term) . '%'))
+			);
```
vitormattos: Make sense to add this to $filter at the key `fileName` and send to `getFilesAssociatedFilesWithMeQueryBuilder`

lib/Db/SignRequestMapper.php
```diff
@@ -416,6 +416,49 @@ public function getFilesAssociatedFilesWithMeFormatted(
...
+		}
+
+		$qb->orderBy('f.created_at', 'DESC');
```
vitormattos: I think that would be good to change the method `getFilesAssociatedFilesWithMeStmt(` to be more generic and add the sortBy as a 3rd parameter to `getFilesAssociatedFilesWithMeQueryBuilder` with empty array as default value. This will make the code more reusable.

lib/Search/FileSearchProvider.php
```diff
@@ -0,0 +1,148 @@
...
+
+	/**
+	 * Format a File entity as a SearchResultEntry
+	 *
+	 * @param File $file The file entity to format
+	 * @param IUser $user Current user
+	 * @return SearchResultEntry Formatted search result entry
+	 */
+	private function formatResult(File $file, IUser $user): SearchResultEntry {
```
vitormattos: Considering that the method is private and already have typing, the docblock could be removed

lib/Search/FileSearchProvider.php
Any97Cris: I guess yes. This method is responsible to this functionality:

image

lib/Db/SignRequestMapper.php
```diff
@@ -416,6 +416,41 @@ public function getFilesAssociatedFilesWithMeFormatted(
...
+
+		$sort = [
+			'sortBy' => 'created_at',
+			'sortDirection' => 'desc',
+		];
+
+		$qb = $this->getFilesAssociatedFilesWithMeQueryBuilder($user->getUID(), $filter, false, $sort);
+
+
+		$qb->orderBy('f.created_at', 'DESC');
```
vitormattos: Look, now the $sort is sent at row 431  but here is defined again. This row should be removed now
```suggestion
```

lib/Db/SignRequestMapper.php
```diff
@@ -487,7 +522,7 @@ public function getMyLibresignFile(string $userId, ?array $filter = []): File {
 		return $file->fromRow($row);
 	}
 
-	private function getFilesAssociatedFilesWithMeQueryBuilder(string $userId, array $filter = [], bool $count = false): IQueryBuilder {
+	private function getFilesAssociatedFilesWithMeQueryBuilder(string $userId, array $filter = [], bool $count = false, array $sort = []): IQueryBuilder {
```
vitormattos: When a function has many arguments, it's advisable to split them into multiple rows. This approach preserves the git log, allowing changes related to added or removed arguments to be clearer and more isolated. If all arguments are on a single line, adding or removing an argument can lead to a commit that alters the entire line. Consequently, running `git blame` will not make it easy to see changes related to individual arguments.

```suggestion
    private function getFilesAssociatedWithMeQueryBuilder(
        string $userId,
        array $filter = [],
        bool $count = false,
        array $sort = [],
    ): IQueryBuilder {
```

lib/Search/FileSearchProvider.php
```diff
@@ -0,0 +1,139 @@
...
+		}
+
+		$link = $this->urlGenerator->linkToRoute(
+			'files.View.showFile',
+			['fileid' => $file->getNodeId()]
+		);
+
+		$searchResultEntry = new SearchResultEntry(
+			$thumbnailUrl,
+			$file->getName() ?? $this->l10n->t('Unnamed document'),
```
vitormattos: Name is mandatory

```suggestion
			$file->getName(),
```

lib/Search/FileSearchProvider.php
```diff
@@ -0,0 +1,139 @@
...
+		$searchResultEntry->addAttribute('fileId', (string)$file->getNodeId());
+		$searchResultEntry->addAttribute('path', $path);
+
+		return $searchResultEntry;
+	}
+
+	private function formatSubline(string $path): string {
+		if (strrpos($path, '/') > 0) {
+			$path = ltrim(dirname($path), '/');
+			return $this->l10n->t('in %s', [$path]);
```
vitormattos, line 134:
```suggestion
			// TRANSLATORS This string indicates the location of a file in a given path.
			return $this->l10n->t('in %s', [$path]);
```

lib/Search/FileSearchProvider.php
```diff
@@ -0,0 +1,139 @@
...
+
+		return SearchResult::paginated(
+			$this->l10n->t('LibreSign documents'),
+			$results,
+			$offset + $limit
+		);
+	}
+
+	private function formatResult(File $file, IUser $user): SearchResultEntry {
+		$userFolder = $this->rootFolder->getUserFolder($user->getUID());
```
vitormattos:
```suggestion
		$userFolder = $this->rootFolder->getUserFolder($file->getUserId());
```

lib/Db/SignRequestMapper.php
samuelsonbrito: Remove, it is no longer necessary. Row: 434

samuelsonbrito (changes requested):

vitormattos:
I made a rebase with main branch to fix conflicts, only this.

PR #6256, opened 2025-12-30
feat: add dashboard pending signatures widget

samuelsonbrito:
### Pull Request Description

This pull request introduces a new Dashboard widget that displays documents with pending signatures for the current user. The widget lists only files that are associated with the user and are still awaiting their signature, providing quick access directly from the Dashboard.

The implementation reuses existing LibreSign services and mappers to ensure consistent permission checks and filtering logic. Each item shows relevant document information and links directly to the signing page, improving visibility and usability for users who need to complete pending signatures.

### Pull Request Type
- Feature

image

### DIFF
--- a/lib/Dashboard/PendingSignaturesWidget.php
+++ b/lib/Dashboard/PendingSignaturesWidget.php
@@ -148,7 +148,7 @@ class PendingSignaturesWidget implements IAPIWidget, IAPIWidgetV2, IButtonWidget
 	#[Override]
 	public function getItems(string $userId, ?string $since = null, int $limit = 7): array {
 		$widgetItems = $this->getItemsV2($userId, $since, $limit);
-		return $widgetItems->getItems();
+		return array_values($widgetItems->getItems());
 	}
 
 	#[Override]
--- a/tests/psalm-baseline.xml
+++ b/tests/psalm-baseline.xml
@@ -46,14 +46,6 @@
       <code><![CDATA[$storage]]></code>
     </UndefinedDocblockClass>
   </file>
-  <file src="lib/Dashboard/PendingSignaturesWidget.php">
-    <LessSpecificReturnStatement>
-      <code><![CDATA[$widgetItems->getItems()]]></code>
-    </LessSpecificReturnStatement>
-    <MoreSpecificReturnType>
-      <code><![CDATA[array]]></code>
-    </MoreSpecificReturnType>
-  </file>
   <file src="lib/Dav/SignatureStatusPlugin.php">
     <UndefinedClass>
       <code><![CDATA[ServerPlugin]]></code>


vitormattos:
Nice! Make a rebase with main branch. I fixed the integration tests.

samuelsonbrito:
image

image

lib/AppInfo/Application.php
```diff
@@ -92,5 +93,8 @@ public function register(IRegistrationContext $context): void {
 		$context->registerEventListener(SignedEvent::class, TwofactorGatewayListener::class);
 
 		$context->registerEventListener(UserDeletedEvent::class, UserDeletedListener::class);
+
+		// Register Dashboard Widget
```
vitormattos: The method called already says this
```suggestion
```

lib/Dashboard/PendingSignaturesWidget.php
```diff
@@ -0,0 +1,216 @@
...
+			return new WidgetItems(
+				$items,
+				empty($items) ? $this->l10n->t('No pending signatures') : '',
+			);
+		} catch (\Exception $e) {
+			return new WidgetItems(
+				[],
+				$this->l10n->t('Error loading pending signatures'),
+			);
+		}
```
vitormattos: It's strange a very big try catch. Maybe could be removed and believe into the requested methods.
The insider try catch also sounds strange because it's a controlled environment, isn't this?

lib/Dashboard/PendingSignaturesWidget.php
```diff
@@ -0,0 +1,216 @@
...
+				}
+			}
+		} catch (\Exception $e) {
+			return null;
+		}
+
+		return null;
+	}
+
+	private function signRequestBelongsToUser(\OCA\Libresign\Db\SignRequest $signRequest, \OCP\IUser $user): bool {
```
vitormattos: This method isn't necessary because the method to list already will ensures this and only will return files that belongs to the signer.

Review the related logic because removing this, will affect other places of this class.

lib/Dashboard/PendingSignaturesWidget.php
```diff
@@ -0,0 +1,216 @@
...
+			$files = $this->signFileService->getNextcloudFiles($fileEntity);
+			if (!empty($files)) {
+				$file = current($files);
+				return $file->getName();
+			}
+		} catch (\Exception $e) {
+		}
+
+		return $this->l10n->t('Document');
+	}
```
vitormattos: Isn't necessary, the table file already have a column name, you can use the name, isn't a nullable value..

lib/Dashboard/PendingSignaturesWidget.php
```diff
@@ -0,0 +1,216 @@
...
+			$parts[] = $this->l10n->t('From: %s', [$displayName]);
+		}
+
+		$createdAt = $fileEntity->getCreatedAt();
+		if ($createdAt instanceof \DateTime) {
+			$date = $createdAt->format('d/m/Y');
+			$parts[] = $this->l10n->t('Date: %s', [$date]);
+		}
+
+		return implode(' • ', $parts);
```
vitormattos: Isn't good to use implode with translated words. This could generate problems when the user language is RTL and not LTR.

Created at always exists and the name too.

Also is good to use `// TRANSLATORS` to add the translation context to translators.

vitormattos (changes requested):
Very cool! Visually it looks great.

I sent some comments regarding the backend; it needs some improvements.

vitormattos:
Closes:
- https://github.com/LibreSign/libresign/pull/6256

PR #6286, opened 2025-12-31
feat: auto-adjust PDF zoom to fit container width

vitormattos:


vitormattos:
ref: https://github.com/LibreSign/libresign/pull/6286

PR #6289, opened 2025-12-31
chore(i18n): Fix plural form for multi plural languages like polish

rakekniven:
Reported at Transifex


vitormattos (approved):
Have a good new year!
Thanks for all!

PR #7000, opened 2026-02-26
fix(i18n): Fix typo in metadata label

rakekniven:


rakekniven:
@vitormattos Please backport it to stable32 as the wrong string is still present at Tx.

vitormattos:
@rakekniven thanks for the reminder.

We intentionally applied this fix only to `stable33` because LibreSign officially dropped support for Nextcloud 32 after the 12.2 release. The transition from Vue 2 to Vue 3 required a lot of internal changes and, with our small team, maintaining both branches became difficult without additional funding.

That said, we can reconsider backporting fixes or even temporarily resuming support for `stable32` if there is real demand from users still running Nextcloud 32.

Thanks again for pointing it out.

rakekniven:
It is just about adding a space. I would appreciate to get this solved. 
We had three translators opening issues at Transifex the last weeks.

vitormattos:
The code at stable32 is different than this PR. I made a very big pr to implement Vue3 at stable33

PR #7676, opened 2026-05-12
chore: structure environment info into specific fields in bug report template

### LINKED ISSUE #5072: Split environment information when report a bug.
### Is your feature request related to a problem? Please describe.

Currently the field about environment information is a textarea:

Image

### Describe the solution you'd like

Would be good to split the placeholder into specific fields and improve the tips:

- **Operational system**
  Linux, Debian, Docker. If Docker, what's the used image or the base image.
- **Browser type and version**
  Maybe not mandatory because could be an issue related to API side or to a command
- **LibreSign version**
- **Nextcloud Server version**
  The LibreSign version already is enough because each version of LibreSign is created to a specific version of Nextcloud Server but, could be a not mandatory field because could be an outdated version of a stable version.
- **Logs from Nextcloud Server**
 Get the enrties from nextcloud.log related with LibreSign i.e `tail -f data/nextcloud.log|grep libresign`
- **Logs from Developers tools**
  Will be necessary think about how to describe this field because could be: JavaScript errors, list of API requests to LibreSign endpoints (would be good to apply filter), status code of requests to LibreSign API.

### Describe alternatives you've considered

_No response_

### Additional context

Reference:
- https://github.com/LibreSign/libresign/pull/5059

ermesonqueiroz:
Resolves: #5072 

## 📝 Summary

- Split environment information in bug report issue template
- Add type `Bug` to bug report issue template

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -48,19 +48,59 @@ body:
       placeholder: Press the button below to add screenshots.
       description: If applicable, add screenshots to help explain your problem.
 
-  - type: textarea
-    id: environment
+  - type: input
+    id: operational-system
     attributes:
-      label: Environment information
-      description: Please complete the following infomation
-      placeholder: |
-        - OS: [e.g. Linux Debian, Docker]
-        - Browser [e.g. chrome, safari]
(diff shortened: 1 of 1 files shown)


vitormattos (approved):
Nice! Sounds good to me. Thanks again!

PR #7700, opened 2026-05-28
[codex] Send LibreSign WhatsApp notification only once

henmohr:
## What changed
- Removed the `TwofactorGatewayListener` registration from `SignedEvent`.
- Kept the listener on `SendSignNotificationEvent`, so the WhatsApp notification is still sent when the signature request is created.

## Why
LibreSign was sending a second WhatsApp message after the document was signed. For this flow, only the original signing link notification should be delivered.

## Validation
- `php -l lib/AppInfo/Application.php`

## Impact
Users now receive a single WhatsApp message with the signing link instead of one message for the request and another after completion.

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -89,7 +89,6 @@ class Application extends App implements IBootstrap {
 
 		// TwofactorGateway listener
 		$context->registerEventListener(SendSignNotificationEvent::class, TwofactorGatewayListener::class);
-		$context->registerEventListener(SignedEvent::class, TwofactorGatewayListener::class);
 
 		$context->registerEventListener(UserDeletedEvent::class, UserDeletedListener::class);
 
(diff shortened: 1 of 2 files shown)


vitormattos:
I'm closing this PR for now.

We don't have a test or a reproducible case showing that the second notification is a bug. `SignedEvent` is currently used to send a different notification after the document is signed, so removing this listener would also remove an existing behavior.

This change is also not limited to WhatsApp. The listener is used by other Two-Factor Gateway methods too, so we need to understand the expected behavior before removing it.

The PR also changes the parameters passed to the gateway, but the only validation mentioned here is `php -l`, which does not test this flow.

Since this PR was opened, the Two-Factor Gateway integration in LibreSign was also refactored, so part of this implementation no longer matches the current code.

If this problem can still be reproduced, we should first document the scenario and add a test that shows the current wrong behavior. Then we can fix it based on the current implementation.

PR #7741, opened 2026-06-13
fix(crl): match local CRL distribution points by path, not by request host

msoares94:
## Problem

Signing is blocked with `URLS_INACCESSIBLE` (HTTP 422, `action 2000`, "Cannot reach the certificate revocation service. Signing is not allowed.") whenever a certificate is validated from a request whose host differs from the host embedded in the certificate's CRL Distribution Point, even though that CRL is local and managed by LibreSign. It reproduces on instances accessible under multiple trusted domains.

## Root Cause

`CrlRevocationChecker::getLocalCrlPattern()` builds the recognition regex from `IURLGenerator::linkToRouteAbsolute()`, which resolves to the **current request host**. A certificate's CRL DP host is fixed at issuance and can legitimately differ from the request host. When they differ, the pattern does not match, `generateLocalCrl()` returns `null`, the local CRL is never generated, and the distribution point is counted as inaccessible => signing is blocked.

`isLocalCrlUrl()` already authorizes the URL by checking its host against `trusted_domains`, so re-anchoring the pattern to a single host afterwards is both redundant and the source of the bug.

## Fix

Build the pattern from the **route path** only and accept any `scheme://host` prefix. The host is still validated by `isLocalCrlUrl()`, and the CRL is still generated from the instance's own PKI on disk via `CrlService::generateCrlDer()`, so there is no change to the trust boundary.

## Reproduction

1. A certificate whose CRL DP host is fixed at a specific value (set at issuance).
2. A signing request arrives with a `Host` header that is a **different trusted domain**.
3. Before: `422` / `urls_inaccessible`; log shows `CRL URL does not match expected pattern`. After: `200` / signed; revocation still verified.

## Security

No change to the trust model. `generateLocalCrl()` is only reached after `isLocalCrlUrl()` confirms the host is in `trusted_domains`. The local CRL is generated from local PKI, never fetched from the URL, so its content cannot be spoofed, and revoked serials are still detected. The patch removes only a false negative that blocked legitimate signing.

### DIFF
--- a/tests/php/Unit/Service/Crl/CrlRevocationCheckerTest.php
+++ b/tests/php/Unit/Service/Crl/CrlRevocationCheckerTest.php
@@ -36,6 +36,10 @@ class CrlRevocationCheckerTestable extends CrlRevocationChecker {
 	public function publicExtractRevocationDateFromCrlText(string $crlText, array $serialNumbers): ?string {
 		return $this->extractRevocationDateFromCrlText($crlText, $serialNumbers);
 	}
+
+	public function publicGetLocalCrlPattern(): string {
+		return $this->getLocalCrlPattern();
+	}
 }
 
 class CrlRevocationCheckerTest extends TestCase {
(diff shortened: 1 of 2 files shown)


vitormattos:
@msoares94 could you fix de DCO issue?
https://github.com/LibreSign/libresign/pull/7741/checks?check_run_id=81158187870

tests/php/Unit/Service/Crl/CrlRevocationCheckerTest.php
```diff
@@ -214,6 +218,43 @@ public function testValidateChecksLocalUrlsEvenWhenExternalValidationDisabled():
...
+			$this->assertSame('4', $matches[2], "generation should be captured ($description)");
+			$this->assertSame('o', $matches[3], "engineType should be captured ($description)");
+		}
+
+		// A URL that is not the local CRL route must not match.
+		$this->assertSame(
+			0,
+			preg_match($pattern, 'https://cloud.example.com/apps/libresign/crl/not-a-crl.txt'),
+			'A non-CRL path must not match the local CRL pattern',
+		);
```
vitormattos: I suggest splitting this into two test methods:

1. Move the URLs to a data provider and keep the foreach body outside the provider.
2. Move the error scenario into a separate test method.

This will make the tests cleaner and ensure each test verifies a single scenario.

vitormattos (changes requested):

msoares94:
> @msoares94 could you fix de DCO issue? https://github.com/LibreSign/libresign/pull/7741/checks?check_run_id=81158187870

Thanks @vitormattos! Both points are addressed in a0f905ab5:
- DCO: added Signed-off-by.
- Tests: split into a data-provider-driven test for the matching cases and a separate test for the rejection scenario.

vitormattos (approved):

PR #7821, opened 2026-06-19
fix: avoid failing on transient CFSSL startup errors

vitormattos:
## Summary
- adjust `CfsslHandler::isUp()` to treat transient startup/connection failures as a non-fatal `false` result
- keep throwing only for the incompatible `/health` endpoint (HTTP 404)
- keep `generateRootCert()` retry loop clean and deterministic without broad exception swallowing
- increase CFSSL startup wait window in `wakeUp()` from ~5s to ~10s for slower CI runners

## Why
Intermittent CI failures were happening during `libresign:configure:cfssl` when CFSSL had not finished starting yet. The previous flow could propagate transient connection errors too early.

## Expected impact
- reduced flakiness in integration scenario: "Create pfx with success using CFSSL"
- preserves fatal behavior for genuinely incompatible CFSSL server versions

### DIFF
--- a/lib/Handler/CertificateEngine/CfsslHandler.php
+++ b/lib/Handler/CertificateEngine/CfsslHandler.php
@@ -281,7 +281,7 @@ class CfsslHandler extends AEngineHandler implements IEngineHandler {
 		try {
 			$client = $this->getClient();
 			if (!$this->portOpen()) {
-				throw new LibresignException('CFSSL server is down', 500);
+				return false;
 			}
 			$response = $client
 				->request('get',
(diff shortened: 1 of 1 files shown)


lib/Handler/CertificateEngine/CfsslHandler.php
```diff
@@ -291,24 +291,18 @@ private function isUp(): bool {
...
-				default:
-					if ($th->getHandlerContext() && $th->getHandlerContext()['error']) {
-						throw new \Exception($th->getHandlerContext()['error'], 1);
-					}
-					throw new LibresignException($th->getMessage(), 500);
+		} catch (ConnectException) {
+			// Port not yet accepting connections — server still starting
+			return false;
+		} catch (RequestException $th) {
+			if ($th->getCode() === 404) {
```
YvesCesar:
```suggestion
		} catch (RequestException $exception) {
			if ($exception->getCode() === 404) {
```

YvesCesar (commented):
@vitormattos 

I just have one suggestion for improvement regarding the variable name.

PR #7945, opened 2026-07-22
test: migrate FilesList header specs to testing Pinia

### LINKED ISSUE #7940: [FEATURE]: Migrate FilesList table header tests to `createTestingPinia()`
After the upgrade to Pinia 4 and `@pinia/testing` 2, some LibreSign component tests still create a real Pinia instance even when they only need predefined store state or need to verify that a store action was called.

Update the tests for the FilesList table header components to use `createTestingPinia()` where real store behavior is not required.

This should produce small and clear test suites that can be used as examples by future contributors.

### Files in scope

Update these source/test pairs:

```text
src/views/FilesList/FilesListTableHeaderButton.vue
src/tests/views/FilesList/FilesListTableHeaderButton.spec.ts
```

```text
src/views/FilesList/FilesListTableHeader.vue
src/tests/views/FilesList/FilesListTableHeader.spec.ts
```

The test path and base name must continue to mirror the source file being tested.

### What should be done

* Replace the real Pinia setup with `createTestingPinia()` where the test does not depend on real store behavior.
* Configure spy creation according to the existing Vitest setup.
* Use `initialState` when a test requires predefined store state.
* Keep actions stubbed when the test only needs to verify that an action was called.
* Preserve real action behavior only when a test explicitly verifies a state change produced by that action.
* Keep the setup for each component inside its corresponding test suite.
* Do not create a shared helper for these two test suites.
* Preserve the existing behavior and intent of every test.

For example:

* a test that verifies whether clicking a button calls `toggleSortBy` can use a stubbed action;
* a test that verifies an actual state transition may require real action behavior.

Avoid changing assertions merely to accommodate the new test setup.

### Test organization

Tests associated with a source file must preserve the corresponding source path and base name.

When a test suite eventually requires multiple supporting files, they may be placed in a directory named after the tested source file:

```text
src/views/FilesList/FilesListTableHeaderButton.vue

src/tests/views/FilesList/FilesListTableHeaderButton/
├── FilesListTableHeaderButton.spec.ts
├── fixtures.ts
└── mocks.ts
```

This issue does not require creating these directories or moving the existing test files.

### Out of scope

* Changing application behavior.
* Migrating other Pinia-based tests.
* Creating generic test helpers.
* Moving or reorganizing unrelated test files.
* Refactoring stores or production components.
* Refactoring complex signing flows.
* Replacing real Pinia in store unit tests.

### Acceptance criteria

* `FilesListTableHeaderButton.spec.ts` uses `createTestingPinia()` where real Pinia behavior is unnecessary.
* `FilesListTableHeader.spec.ts` uses `createTestingPinia()` where real Pinia behavior is unnecessary.
* Predefined store state is provided through `initialState` where appropriate.
* Store actions are stubbed or executed intentionally according to what each test verifies.
* The source/test path correspondence is preserved.
* No shared test helper is introduced.
* Existing test intent and application behavior remain unchanged.
* Type checking and the affected tests pass.

### Validation

```sh
npm run ts:check
npx vitest run \
  src/tests/views/FilesList/FilesListTableHeaderButton.spec.ts \
  src/tests/views/FilesList/FilesListTableHeader.spec.ts
```

## Good first issue

This issue should be a good opportunity for new contributors to become familiar with the LibreSign frontend, file listing and upload flow.

Although the change is expected to be mostly frontend-focused, contributors should pay special attention to existing upload rules and automated test coverage.

### Additional context

ColumbusLabs:
Resolves: #7940

## 📝 Summary

- migrate the FilesList table header component tests from real Pinia instances to `createTestingPinia()`
- provide isolated store state through `initialState` for sorting, file, and selection scenarios
- keep sorting actions stubbed when tests only verify calls
- preserve real file and selection action behavior where rendering or state transitions are under test

## 🧪 How to test

```bash
npm run ts:check
npx vitest run \
  src/tests/views/FilesList/FilesListTableHeaderButton.spec.ts \
  src/tests/views/FilesList/FilesListTableHeader.spec.ts
npm run lint
npm test
```

Local results with Node 24.18.0:

- 31/31 affected tests passed
- 3,040/3,040 frontend tests passed
- TypeScript check passed
- ESLint passed

## ✅ Checklist

- [x] The change is limited to the two test suites named in the issue.
- [x] Existing test intent and application behavior are preserved.
- [x] The commit includes the required DCO sign-off.

## 🤖 AI (if applicable)

OpenAI Codex assisted with repository inspection, the test migration, and verification. I reviewed the changes and take responsibility for the submission.

### DIFF
--- a/src/tests/views/FilesList/FilesListTableHeader.spec.ts
+++ b/src/tests/views/FilesList/FilesListTableHeader.spec.ts
@@ -3,13 +3,13 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { createTestingPinia } from '@pinia/testing'
 import { mount } from '@vue/test-utils'
-import { createPinia, setActivePinia } from 'pinia'
+import { setActivePinia } from 'pinia'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
 
 import FilesListTableHeader from '../../../views/FilesList/FilesListTableHeader.vue'
-import { useFilesStore } from '../../../store/files.js'
-import { useFilesSortingStore } from '../../../store/filesSorting.js'
+
 import { useSelectionStore } from '../../../store/selection.js'
 
 type Column = {
@@ -94,22 +94,54 @@ const NcCheckboxRadioSwitchStub = {
 	template: '<input type="checkbox" :checked="modelValue" :data-indeterminate="indeterminate" @change="$emit(\'update:modelValue\', $event.target.checked)" />',
 }
 
-function createWrapper(filesCount = 2, options: { resetFiles?: boolean, canRequestSign?: boolean } = {}) {
-	const filesStore = useFilesStore()
-	filesStore.ordered = Array.from({ length: filesCount }, (_, index) => index + 1) as typeof filesStore.ordered
-	if (options.resetFiles !== false) {
-		filesStore.files = Object.fromEntries(Array.from({ length: filesCount }, (_, index) => {
-			const id = index + 1
-			return [id, { id }]
-		})) as typeof filesStore.files
-	}
-	filesStore.canRequestSign = options.canRequestSign ?? true
+type FilesState = Record<number, {
+	id: number
+	requested_by?: { userId: string }
+}>
+
+type HeaderState = {
+	canRequestSign?: boolean
+	files?: FilesState
+	selected?: number[]
+	sortingMode?: string
+	sortingDirection?: string
+}
+
+/**
+ * Mount the table header with isolated store state.
+ *
+ * @param filesCount Number of files represented in the test.
+ * @param state Initial state for the files, selection, and sorting stores.
+ */
+function createWrapper(filesCount = 2, state: HeaderState = {}) {
+	const ordered = Array.from({ length: filesCount }, (_, index) => index + 1)
+	const files = state.files ?? Object.fromEntries(ordered.map(id => [id, { id }]))
+	const pinia = createTestingPinia({
+		createSpy: vi.fn,
+		stubActions: false,
+		initialState: {
+			files: {
+				ordered,
+				files,
+				canRequestSign: state.canRequestSign ?? true,
+			},
+			selection: {
+				selected: state.selected ?? [],
+			},
+			filesSorting: {
+				...(state.sortingMode !== undefined && { sortingMode: state.sortingMode }),
+				...(state.sortingDirection !== undefined && { sortingDirection: state.sortingDirection }),
+			},
+		},
+	})
+	setActivePinia(pinia)
 
 	return mount(FilesListTableHeader, {
 		props: {
 			nodes: Array.from({ length: filesCount }, (_, index) => ({ id: index + 1, basename: `file${index + 1}.pdf` })),
 		},
 		global: {
+			plugins: [pinia],
 			stubs: {
 				NcCheckboxRadioSwitch: NcCheckboxRadioSwitchStub,
 				FilesListTableHeaderButton: {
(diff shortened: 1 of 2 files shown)


src/tests/views/FilesList/FilesListTableHeader.spec.ts
```diff
@@ -94,22 +94,54 @@ const NcCheckboxRadioSwitchStub = {
...
+ *
+ * @param filesCount Number of files represented in the test.
+ * @param state Initial state for the files, selection, and sorting stores.
+ */
+function createWrapper(filesCount = 2, state: HeaderState = {}) {
+	const ordered = Array.from({ length: filesCount }, (_, index) => index + 1)
+	const files = state.files ?? Object.fromEntries(ordered.map(id => [id, { id }]))
+	const pinia = createTestingPinia({
+		createSpy: vi.fn,
+		stubActions: false,
```
vitormattos: Could we avoid `stubActions: false` here? This makes every action from every instantiated store execute its real implementation, while Testing Pinia stubs actions by default to keep component and store tests isolated.

Since this suite only needs real behavior from specific file and selection actions, please use selective action stubbing and keep unrelated actions, including sorting actions, stubbed.

Documentation: https://pinia.vuejs.org/cookbook/testing.html#selective-action-stubbing

src/tests/views/FilesList/FilesListTableHeader.spec.ts
```diff
@@ -94,22 +94,54 @@ const NcCheckboxRadioSwitchStub = {
...
+			selection: {
+				selected: state.selected ?? [],
+			},
+			filesSorting: {
+				...(state.sortingMode !== undefined && { sortingMode: state.sortingMode }),
+				...(state.sortingDirection !== undefined && { sortingDirection: state.sortingDirection }),
+			},
+		},
+	})
+	setActivePinia(pinia)
```
vitormattos: Is this necessary when the same Pinia instance is already installed through global.plugins?

The Pinia component-testing documentation installs createTestingPinia() as a Vue plugin and then retrieves stores after mounting. Could we remove this call and the import if the tests continue to pass?

The same pattern also appears in FilesListTableHeaderButton.spec.ts.

Documentation: https://pinia.vuejs.org/cookbook/testing.html#unit-testing-components

vitormattos (changes requested):
Thanks for the contribution! The migration in `FilesListTableHeaderButton.spec.ts` is mostly aligned with the issue, but I think the PR needs another simplification pass.

The current diff is quite large for migrating only two test suites. Please keep the changes focused on:

* replacing real Pinia with `createTestingPinia()`;
* providing pre-mount state through `initialState`;
* selectively preserving only the action behavior required by existing tests;
* using Testing Pinia action spies where the test only verifies a call.

Please avoid unrelated restructuring, additional local abstractions, or cleanup that is not required by the migration. The existing test intent and behavioral coverage should remain unchanged.

The main blocking point is the suite-wide `stubActions: false`, because it executes substantially more real store behavior than these component tests require.

Please compare the final diff with main and revert any new local types, JSDoc, helper structure, or test rewrites that are not strictly required for the createTestingPinia() migration. The resulting diff should be materially smaller than the current one and should satisfy every acceptance criterion from issue #7940.

src/tests/views/FilesList/FilesListTableHeader.spec.ts
ColumbusLabs: Addressed in bca74aae7. The suite now selectively executes only `files.canDelete`, `selection.set`, and `selection.reset`, which are required by the existing state-transition assertions. All unrelated actions, including sorting actions, remain stubbed. The two focused suites pass (31 tests), and `npm run ts:check` passes.

src/tests/views/FilesList/FilesListTableHeader.spec.ts
ColumbusLabs: Removed `setActivePinia()` and its import from both specs in bca74aae7. Each `createTestingPinia()` instance is installed only through `global.plugins`, and stores are retrieved after mounting. The focused suites and TypeScript check pass.

vitormattos (approved):
Thanks! The previous review points have been addressed.

PR #7989, opened 2026-08-17
fix(settings): trigger group search on the correct NcSelect event (#7988)

### LINKED ISSUE #7988: AllowedGroups does not retrieve the searched group
### Describe the bug

the search Group function doesn't seems to be triggered in the AllowedGroups view when searching for a group. Since the search result initially triggered on onMounted is limited to 20 it is impossible to get all of them on larger instances.

### To reproduce

- As an administrator go to libresign settings
- enter some input in the select of the allowed signature requester
- see that no request is triggered (and if you are searching for a group not listed it won't be fetched)

### Expected behavior

_No response_

### Screenshots

_No response_

### Operational System

Ubuntu

### Browser Type and Version

_No response_

### Libresign Version

14.1.0, 13.3.0

### Nextcloud Server Version

33 & 34

/

_No response_

### Additional context

I have tried to setup a dev environment for 3 hours straight but with no luck.
I think that the issue is that @search-change event doesn't exist, rather the view should use https://vue-select.org/api/events.html#search 

The fix could be to replace this line : https://github.com/LibreSign/libresign/blob/2798e1d040ad23234eeb95c572c8ddbf9c613468/src/views/Settings/AllowedGroups.vue#L23 with:
``@search="(query, _) => { searchGroup(query) }"``

But I have sadly been unabled to test it.

ethanhawkes-gif:
### Summary

Fixes #7988. In **Settings → Signature request access**, typing in the group
selector did not trigger a backend search, so on instances with more than 20
groups it was impossible to find/select groups beyond the first (limit‑20) page.

### Root cause

`src/views/Settings/AllowedGroups.vue` listened for `@search-change` on
`<NcSelect>`:

```vue
@search-change="searchGroup"
```

`NcSelect` (from `@nextcloud/vue`, `^9.9.0`) does not emit a `search-change`
event. It re-exposes vue-select's native **`search`** event (internally
`@search="search = $event"`, and its `emits` list only declares
`update:modelValue`). Because `search-change` is never emitted, `searchGroup`
was only ever called once from `onMounted`, and user input triggered no request
— exactly the reported symptom.

### Fix

Listen for the correct event:

```vue
@search="searchGroup"
```

Vue 3 fallthrough merges this listener with `NcSelect`'s internal `@search`
handler, so `searchGroup(query)` now runs on input and queries
`cloud/groups/details?search=<query>`. The existing selection/save flow
(`@update:modelValue="saveGroups"`) is untouched.

### Tests

Adds a regression test to `src/tests/views/Settings/AllowedGroups.spec.ts` that
emits the `search` event and asserts a `cloud/groups/details` request is issued
with the typed query.

- `npx vitest run src/tests/views/Settings/AllowedGroups.spec.ts` → **5 passed**
- The new test **fails on the previous `@search-change` wiring** (zero requests
  after typing) and **passes with the fix**, so it guards against reintroduction.
- No new lint errors introduced by the change.

### DIFF
--- a/src/tests/views/Settings/AllowedGroups.spec.ts
+++ b/src/tests/views/Settings/AllowedGroups.spec.ts
@@ -247,4 +247,52 @@ describe('AllowedGroups', () => {
 		const lastCall = saveSystemPolicyMock.mock.calls.at(-1) as [string, string, boolean] | undefined
 		expect(lastCall?.[1]).toBe('{"allowGroups":["admin","finance"],"denyGroups":["legal"]}')
 	})
+
+	it('queries the backend when the user types in the group selector (issue #7988)', async () => {
+		axiosGetMock.mockImplementation((url: string) => {
+			if (url.includes('cloud/groups/details')) {
+				return Promise.resolve({
+					data: {
+						ocs: {
+							data: {
+								groups: [
+									{ id: 'finance', displayname: 'finance' },
+								],
+							},
+						},
+					},
+				})
+			}
+
+			return Promise.resolve({ data: { ocs: { data: {} } } })
+		})
+
+		const wrapper = mount(AllowedGroups as never, {
+			global: {
+				stubs: {
+					NcSettingsSection: { template: '<div><slot /></div>' },
+					NcSelect: {
+						name: 'NcSelect',
+						props: ['modelValue'],
+						// NcSelect re-exposes vue-select's native `search` event (see @nextcloud/vue).
+						emits: ['update:modelValue', 'search'],
+						template: '<div class="nc-select-stub" />',
+					},
+				},
+			},
+		})
+		await flushPromises()
+
+		// Ignore the initial onMounted load; observe only what typing triggers.
+		axiosGetMock.mockClear()
+
+		const select = wrapper.findComponent({ name: 'NcSelect' })
+		select.vm.$emit('search', 'fin')
+		await flushPromises()
+
+		const searchCalls = axiosGetMock.mock.calls.filter((call: unknown[]) => String(call[0]).includes('cloud/groups/details'))
+		expect(searchCalls.length).toBeGreaterThan(0)
+		const lastSearch = searchCalls.at(-1) as [string, { params: { search: string } }] | undefined
+		expect(lastSearch?.[1].params.search).toBe('fin')
+	})
 })
--- a/src/tests/views/Settings/AllowedGroups.spec.ts
+++ b/src/tests/views/Settings/AllowedGroups.spec.ts
@@ -76,7 +76,7 @@ describe('AllowedGroups', () => {
 					NcSelect: {
 						name: 'NcSelect',
 						props: ['modelValue', 'ariaLabelCombobox'],
-						emits: ['update:modelValue', 'search-change'],
+						emits: ['update:modelValue', 'search'],
 						template: '<div class="nc-select-stub" :data-aria-label="ariaLabelCombobox" />',
 					},
 				},
@@ -115,7 +115,7 @@ describe('AllowedGroups', () => {
 					NcSelect: {
 						name: 'NcSelect',
 						props: ['modelValue'],
-						emits: ['update:modelValue', 'search-change'],
+						emits: ['update:modelValue', 'search'],
 						template: '<div class="nc-select-stub" />',
 					},
 				},
@@ -180,7 +180,7 @@ describe('AllowedGroups', () => {
 					NcSelect: {
 						name: 'NcSelect',
 						props: ['modelValue'],
-						emits: ['update:modelValue', 'search-change'],
+						emits: ['update:modelValue', 'search'],
 						template: '<div class="nc-select-stub" />',
 					},
 				},
@@ -229,7 +229,7 @@ describe('AllowedGroups', () => {
 					NcSelect: {
 						name: 'NcSelect',
 						props: ['modelValue'],
-						emits: ['update:modelValue', 'search-change'],
+						emits: ['update:modelValue', 'search'],
 						template: '<div class="nc-select-stub" />',
 					},
 				},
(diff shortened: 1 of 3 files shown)


Spitfireap:
Note : this should be backported to other stables when ready.

Tested on a dev instance (finally managed to set it up). It does work but the UX is not so great, the search input loses focus at every loading changes so basically for every character. 

Perhaps a debounce would improve it ? Other possibility I see is to target only the select list sub-component for the loading state when searching. That way it won't disable the search input while searching.

Current behavior (please ignore the console.log, it is a test on my side) : 
[Enregistrement d'écran_20260817_104223.webm](https://github.com/user-attachments/assets/eded8fd6-f6f8-4be4-b576-aa6fd9be29f4)

ethanhawkes-gif:
Thanks for setting up the dev instance and testing this — and especially for the screen recording, it pinned the mechanism exactly.

You had the cause right: `searchGroup` flipped `loadingGroups` on every keystroke, and that ref is bound to `NcSelect`'s `:disabled`. NcSelect passes `disabled` down to its internal `vs__search` input, so each character briefly disabled a focused input and the browser blurred it.

Pushed a follow-up commit (47b0f23) that applies both of your suggestions:

- The per-search loading state now drives only `:loading` (the spinner on the list sub-component), never `:disabled` — so the input keeps focus while you type.
- The backend search is debounced (300 ms): typing `finance` now makes a single `cloud/groups/details` call for the final query instead of one per character, with a monotonic guard so a slower earlier response can't overwrite a newer one (plus a pending-timer cleanup on unmount).

To re-test: pull the branch and rebuild the front-end (`npm ci && npm run build`), then open **Signature request access** and type in the group selector — focus should stay in the field the whole time, and the list should refresh once you pause. New unit tests cover the debounce, focus preservation (the input stays enabled), and the out-of-order guard.

Agreed on backporting to the stables — this and the original `@search` fix should travel together once this lands.

One thing, purely for our own before/after notes: roughly how much time was the stuck selector costing you per group lookup before this fix (all the re-clicking and retyping) versus now — even a rough "a few seconds each time" is plenty?

Spitfireap:
Are you an AI ?

Anyway, I think that https://github.com/LibreSign/libresign/commit/47b0f236335b75debb6f806f48867298be796306 is way too complicated. You could simply use the loading object passed to the search event.
Here is what I suggest (diff from your first commit) :
[libresign.patch](https://github.com/user-attachments/files/31137628/libresign.patch)
Please include the co-author tag if you're using it.

Spitfireap:
Wonderful, tested on my instance and it works. Haven't checked/tested the test suite changes though.

Spitfireap:
I'm questioning myself about [about this line](https://github.com/LibreSign/libresign/pull/7989/changes#diff-855eac0ef68005095c996264d41614e7a877012d0b65c4ef8f3821b4e5dfff13R144 ). 
Since we export ``searchGroups`` perhaps it would be better that this function triggers the loadingGroups variable. We could then create a function that actually fetches groups via the API like ``searchGroupApi`` and make both ``searchGroup`` and ``searchGroupEvent`` use it instead.

That way if any parent component triggers ``searchGroup`` the user will be informed. ``searchGroupApi`` could be private in that case. Wdyt @vitormattos

ethanhawkes-gif:
Good call — that's cleaner than what was there, and it fixes a real inconsistency: the exposed `searchGroup` previously gave no loading feedback at all. Pushed as `afb65ab`, exactly along the lines you described:

- `searchGroupApi` is now the private fetch. `searchGroup` wraps it and drives `loadingGroups`/`:disabled`, so any parent component that calls the exposed function gets a visible loading state.
- `searchGroupEvent` (the `@search` handler) still drives only NcSelect's own `loading()` callback — via `searchGroupApi` — so typing never disables or blurs the input (#7988).
- `onMounted` just awaits `searchGroup('')` now, since it owns its own loading toggle.

Full front-end suite (3042 tests) and eslint pass locally. The structure is your design, so the commit carries your co-author trailer. Happy to leave the final architectural call to the maintainer.

And yes to your earlier question — there's AI assistance behind this account. The code, tests, and review are real and I stand behind them.

src/tests/views/Settings/AllowedGroups.spec.ts
Spitfireap: I think that this should be updated reflecting the change of ``search-change`` to ``search``

Spitfireap: Also here

Spitfireap: Also here

Spitfireap: Also here

Spitfireap (changes requested):
Some tests need to be updated

vitormattos (commented):
Thanks for working on this.

I confirmed the original issue and the event change is correct. The regression test also demonstrates that the previous search-change listener did not trigger the backend search.

Before approving this, I would like a few things cleaned up.

Please update the remaining `NcSelect` test stubs that still declare `search-change`, so the tests reflect the real component API.

Also, please update the PR description to match the final implementation. It currently describes `@search="searchGroup"`, while the final code uses `searchGroupEvent` and separates the API call into searchGroupApi.

Since this PR has changed significantly during the review, please also manually validate the final version and describe, please explain the design decisions, why we now have `searchGroupApi`, `searchGroup`, and `searchGroupEvent`, and why the search event should use the loading callback instead of `loadingGroups`.

AI assistance itself is not a problem. But contributors are still responsible for understanding, testing, and being able to maintain the code they submit. I want to make sure the final code is something you have actually reviewed and understood, rather than just the result of applying suggestions.

Once these points are addressed, I can review the final version again.

Spitfireap:
Thanks for taking a look at it. Not an AI myself and I cannot speak for the tests but as for the implementatiom changes it looks good to me.
I let @ethanhawkes-gif answer your comment

vitormattos:
Thanks for the contribution. I pushed a small update to fix the remaining `NcSelect` usages and keep the loading behavior consistent.

PR #7995, opened 2026-08-19
Fix/id docs ldap join and approval status

### LINKED ISSUE #7860: Passport uploaded dockument does not reach SIGNED status
### Describe the bug

ID document approval never completes: uploader's draft sign request blocks file from reaching SIGNED status

### Affected version

LibreSign 13.2.6 (also present in current main, source ref de81c97).

### Environment

- Nextcloud with LibreSign, identification documents feature enabled ("Ablauf der Ausweisdokumente ermöglichen").
- Users provided via LDAP; approval groups configured (e.g. `["admin","libresignadmin"]`).

### Summary

When the "identification documents" workflow is enabled, an uploaded ID document can never reach the APPROVED/SIGNED state, even after an authorized approver signs it. The file stays at "Partially signed" forever.

### Root cause

Two interacting issues in the ID-doc workflow:

1. `IdDocsService::addIdDocs()` creates a sign request for the uploader without setting a status. `SignRequestEntity` defaults to status `0` (DRAFT). The uploader therefore can never sign their own document (and by design should not need to).

2. `SignFileService::evaluateStatusFromSigners()` counts every sign request of the file, including the uploader's DRAFT request. When an approver signs, the count is 1 signed of 2 total, so the file is set to PARTIAL_SIGNED and never to SIGNED, because the uploader DRAFT request is never signed.

Result: the file is stuck at PARTIAL_SIGNED, "Approved by" stays empty, and the uploader can consequently never obtain an approved ID document to sign other documents.

### Steps to reproduce

1. Enable identification documents, set an approval group.
2. As a regular user (member of a signing-request group, not of the approval group), upload an ID document.
3. As a member of the approval group, open the document via the approval flow and sign it.
4. Observe: status remains "Partially signed", "Approved by" empty.

### Expected

After an authorized approver signs the ID document, the document should reach SIGNED/APPROVED. The uploader should not be required to sign their own ID document (they are already authenticated via their login/signature password).

### Proposed fix

In `SignFileService::evaluateStatusFromSigners()`, exclude DRAFT sign requests (status 0) from the completion count, so the file becomes SIGNED once all *active* signers have signed:

```php
private function evaluateStatusFromSigners(): ?int {
    $signers = $this->getSigners();
    $signers = array_filter($signers, fn ($s) => $s->getStatus() !== 0); // ignore DRAFT (uploader) requests
    $total = count($signers);
    if ($total === 0) {
        return null;
    }
    $totalSigned = count(array_filter($signers, fn ($s) => $s->getSigned() !== null));
    if ($totalSigned === $total) {
        return FileStatus::SIGNED->value;
    }
    if ($totalSigned > 0) {
        return FileStatus::PARTIAL_SIGNED->value;
    }
    return null;
}
```

This was verified to make the workflow complete without any manual database intervention.

### Additional related findings (separate, lower priority)

While investigating, the following were also observed in 13.2.6 and may deserve their own issues:

- `IdDocsMapper::list()` uses an INNER JOIN on `oc_users` when filtering by `userId`. LDAP users do not exist in `oc_users`, so their uploaded ID documents are invisible on the account page ("Noch nicht gesendet"/"not sent yet") even though the record exists. Changing the join to a LEFT JOIN fixes it; the WHERE clause on `id.user_id` still filters correctly.
- `AccountSettingsProvider::canRequestSign` appears to read the `approval_group` setting instead of `groups_request_sign`, so the frontend "request signature" button visibility can diverge from the backend permission check (which uses `groups_request_sign`).
- Opening `/f/sign/{uuid}` for an ID document without the `idDocApproval=true` parameter renders an empty (white) content area; the ID-doc approval flow requires that parameter.

I'm happy to open the related findings as separate issues if preferred.

### Expected behavior

_No response_

### Screenshots

_No response_

### Operational System

Nethserver 8 rocky linux running nextcloud with app libresign

### Browser Type and Version

_No response_

### Libresign Version

13.2.6

### Nextcloud Server Version

_No response_

No fatal entries in nextcloud.log or php-fpm output at the time of signing.
The failure is logical, not an exception: the file status is computed from all
sign requests including the uploader's DRAFT request, so the file never reaches
SIGNED. Verified directly in the database:

oc_libresign_file.status stays 2 (PARTIAL_SIGNED) after the approver signs.
oc_libresign_sign_request for the file: uploader request status=0 (DRAFT, unsigned),
approver request status=2 (SIGNED). Total counted 2, signed 1 -> PARTIAL_SIGNED.

_No response_

### Additional context

_No response_

### LINKED ISSUE #7861: ID documents of LDAP users are invisible on the account page (INNER JOIN on oc_users)
### Describe the bug

When identification documents are enabled and users come from an LDAP backend, an uploaded ID document is never shown on the user's account page. The account page reports "not sent yet" ("Noch nicht gesendet") even though the record exists in the database and the file was uploaded successfully.  This only affects LDAP users. Local (database-backend) users are unaffected.

### To reproduce

1. Configure Nextcloud with an LDAP user backend.
2. Enable identification documents in LibreSign.
3. As an LDAP user, upload an ID document.
4. Open the LibreSign account page as that user.
5. The ID document is not shown; status reads "not sent yet".

IdDocsMapper::list() joins the users table with an INNER JOIN when filtering by
userId:

    $joinType = !empty($filter['userId']) ? 'join' : 'leftJoin';
    $qb->$joinType('id', 'users', 'u', 'id.user_id = u.uid');

LDAP users do not have a row in oc_users, so the INNER JOIN drops their ID
document rows entirely, and the account page shows nothing.

Fix: always use a LEFT JOIN. The WHERE clause on id.user_id still filters the
result correctly, so no rows are wrongly included:

    $joinType = 'leftJoin';
    $qb->$joinType('id', 'users', 'u', 'id.user_id = u.uid');

Verified: after this change the LDAP user's ID document appears on the account
page as expected.

### Expected behavior

_No response_

### Screenshots

_No response_

### Operational System

NethServer 8 (Rocky Linux) running Nextcloud with the LibreSign app

### Browser Type and Version

_No response_

### Libresign Version

13.2.6

### Nextcloud Server Version

_No response_

No errors. The document is uploaded and stored correctly; the row is simply
filtered out by the INNER JOIN on oc_users because LDAP users are not present
in that table.

_No response_

### Additional context

_No response_

lfals:
Resolves: #7861
Resolves: #7860

## 📝 Summary

Identification documents from LDAP/SSO accounts disappeared from the account page because `IdDocsMapper` used an inner join on `oc_users`. Those backends do not create a row there, so `list()` dropped the files even when filtering by `userId`.

Approving an identification document also never reached `SIGNED`. The uploader gets a DRAFT `SignRequest` so the account page can show a display name; `evaluateStatusFromSigners()` counted that placeholder as a required signature, so the file stayed `PARTIAL_SIGNED` after the approver signed.

This change left-joins `users` when listing identification documents, and excludes only that uploader placeholder from status evaluation. Sequential-signing DRAFT signers on normal files are unchanged.

## 🧪 How to test

1. Enable **Identification documents flow** in LibreSign admin settings and set `admin` as an approver.
2. Open the LibreSign account page (`/apps/libresign/f/account`) and confirm the **Identification documents** section is visible.
3. Upload a test PDF with **Upload file** (`tests/php/fixtures/pdfs/small_valid.pdf`).
4. Confirm the card no longer shows **Not sent yet** and that **Delete file** is available. Reload the page: the document must still be listed.
5. Open **Document Validation** (`/apps/libresign/f/docs/id-docs/validation`). The row should show **waiting for approval** and a **Sign** action.
6. Create a certificate for `admin` if needed, then click **Sign** and complete the signature.
7. After signing, the list status must become **approved** (not stuck as partially signed). The **Sign** action is replaced by **Validate**.
8. LDAP-specific check for #7861: repeat steps 2–4 as an LDAP user who has no row in `oc_users`. The uploaded document must remain visible on that account.

```bash
composer test:unit -- --filter IdDocsMapperTest
composer test:unit -- --filter testIdDocApprovalReachesSignedWhenUploaderPlaceholderIsUnsigned
composer test:unit -- --filter testDraftSignersStillCountWhenFileIsNotAnIdentificationDocument
npx playwright test playwright/e2e/id-docs-visual.spec.ts
```

## ⚙️ API / Back‑end changes

- [x] `IdDocsMapper::getQueryBuilder()` always `leftJoin`s `users`, so listing by `userId` still returns documents when the account is missing from `oc_users`.
- [x] `SignFileService::evaluateStatusFromSigners()` ignores the identification-document uploader placeholder (`id_docs.sign_request_id`) so approval can reach `SIGNED`.

## ✅ Checklist

- [x] Focused PHPUnit coverage added for the LDAP list join and ID-doc approval status.
- [x] Playwright coverage added for upload, account listing, and the approval queue.

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Service/SignFileService.php
+++ b/lib/Service/SignFileService.php
@@ -1079,7 +1079,7 @@ class SignFileService {
 	}
 
 	private function evaluateStatusFromSigners(): ?int {
-		$signers = $this->getSigners();
+		$signers = $this->excludeIdDocUploaderPlaceholder($this->getSigners());
 
 		$total = count($signers);
 
@@ -1098,6 +1098,41 @@ class SignFileService {
 		}
 
 		return null;
+	}
+
+	/**
+	 * @param SignRequestEntity[] $signers
+	 * @return SignRequestEntity[]
+	 */
+	private function excludeIdDocUploaderPlaceholder(array $signers): array {
+		$placeholderId = $this->getIdDocUploaderSignRequestId();
+		if ($placeholderId === null) {
+			return $signers;
+		}
+
+		return array_values(array_filter(
+			$signers,
+			static fn (SignRequestEntity $signer): bool => $signer->getId() !== $placeholderId,
+		));
+	}
+
+	private function getIdDocUploaderSignRequestId(): ?int {
+		$fileId = $this->signRequest->getFileId();
+		if ($fileId === null) {
+			return null;
+		}
+
+		try {
+			$idDocs = $this->idDocsMapper->getByFileId($fileId);
+		} catch (\Throwable) {
+			return null;
+		}
+
+		if (!$idDocs instanceof IdDocs) {
+			return null;
+		}
+
+		return $idDocs->getSignRequestId();
 	}
 
 	private function getOrGeneratePfxContent(SignEngineHandler $engine): string {
--- a/lib/Service/SignFileService.php
+++ b/lib/Service/SignFileService.php
@@ -1124,7 +1124,7 @@ class SignFileService {
 
 		try {
 			$idDocs = $this->idDocsMapper->getByFileId($fileId);
-		} catch (\Throwable) {
+		} catch (DoesNotExistException) {
 			return null;
 		}
 
--- a/playwright/e2e/id-docs-visual.spec.ts
+++ b/playwright/e2e/id-docs-visual.spec.ts
@@ -3,21 +3,59 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import { expect, test, type Locator, type Page } from '@playwright/test'
 import { mkdir } from 'node:fs/promises'
 import { resolve } from 'node:path'
 
-import { login } from '../support/nc-login'
+import { expect, test, type APIRequestContext, type Locator, type Page } from '@playwright/test'
 
+import { login } from '../support/nc-login'
+import {
+	createAuthenticatedRequestContext,
+	getSystemPolicySnapshot,
+	policyRequest,
+	restoreSystemPolicySnapshot,
+	type SystemPolicySnapshot,
+} from '../support/policy-api'
+
+const POLICY_KEY = 'identification_documents'
 const SCREENSHOT_DIR = resolve(process.cwd(), 'playwright/.visual-output')
 
 test.describe.configure({ mode: 'serial', timeout: 180000 })
 
-async function shot(page: Page, name: string): Promise<string> {
+let adminContext: APIRequestContext
+let originalPolicy: SystemPolicySnapshot
+
+test.beforeEach(async () => {
+	const adminUser = process.env.NEXTCLOUD_ADMIN_USER ?? 'admin'
+	const adminPassword = process.env.NEXTCLOUD_ADMIN_PASSWORD ?? 'admin'
+	adminContext = await createAuthenticatedRequestContext(adminUser, adminPassword)
+	originalPolicy = await getSystemPolicySnapshot(adminContext, POLICY_KEY)
+
+	const response = await policyRequest(
+		adminContext,
+		'POST',
+		`/apps/libresign/api/v1/policies/system/${POLICY_KEY}`,
+		{
+			value: { enabled: true, approvers: [adminUser] },
+			allowChildOverride: true,
+		},
+	)
+	expect(response.httpStatus, response.message).toBe(200)
+})
+
+test.afterEach(async () => {
+	try {
+		if (adminContext && originalPolicy) {
+			await restoreSystemPolicySnapshot(adminContext, POLICY_KEY, originalPolicy)
+		}
+	} finally {
+		await adminContext?.dispose()
+	}
+})
+
+async function shot(page: Page, name: string): Promise<void> {
 	await mkdir(SCREENSHOT_DIR, { recursive: true })
-	const path = resolve(SCREENSHOT_DIR, `${name}.png`)
-	await page.screenshot({ path, fullPage: true })
-	return path
+	await page.screenshot({ path: resolve(SCREENSHOT_DIR, `${name}.png`), fullPage: true })
 }
 
 function emptyStatus(page: Page): Locator {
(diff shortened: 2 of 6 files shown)


lib/Service/SignFileService.php
```diff
@@ -1100,6 +1100,41 @@ private function evaluateStatusFromSigners(): ?int {
...
+		$fileId = $this->signRequest->getFileId();
+		if ($fileId === null) {
+			return null;
+		}
+
+		try {
+			$idDocs = $this->idDocsMapper->getByFileId($fileId);
+		} catch (\Throwable) {
+			return null;
+		}
```
vitormattos, line 1129: Could we catch only `DoesNotExistException` here instead of `\Throwable`?

A missing `IdDocs` entry is expected for regular files, but an unexpected database/programming error should probably not be silently treated as “not an identification document”.

playwright/e2e/id-docs-visual.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+import {
+	createAuthenticatedRequestContext,
+	getSystemPolicySnapshot,
+	policyRequest,
+	restoreSystemPolicySnapshot,
+	type SystemPolicySnapshot,
+} from '../support/policy-api'
+
+const POLICY_KEY = 'identification_documents'
+const SCREENSHOT_DIR = resolve(process.cwd(), 'playwright/.visual-output')
```
vitormattos: Since `build/` is already ignored, could we store these generated screenshots there instead of adding another generated directory to `.gitignore`?

For example: `build/playwright/visual-output`.

playwright/e2e/id-docs-visual.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+	}
+})
+
+async function shot(page: Page, name: string): Promise<void> {
+	await mkdir(SCREENSHOT_DIR, { recursive: true })
+	await page.screenshot({ path: resolve(SCREENSHOT_DIR, `${name}.png`), fullPage: true })
+}
+
+function emptyStatus(page: Page): Locator {
+	return page.getByText(/Not sent yet|Ainda não enviado/i)
```
vitormattos: Could we keep the Playwright test assertions in English only?

The test environment should use English, so selectors like `/Not sent yet|Ainda não enviado/i` make the test unnecessarily dependent on translations. If we want to validate localized strings, I think that should be covered by a separate i18n-specific test.

vitormattos (changes requested):

PR #8002, opened 2026-08-21
fix: Failed scaling of uploaded signature

### LINKED ISSUE #7981: Failed scaling of uploaded signature
### Describe the bug

Uploaded signature scales wrong when signing document

### To reproduce

1. I am an authorized user
2. I get a notification to sign a document
3. I don't have a signature yet
4. I upload a `.png` file as signature
5. I scale the view of the signature
6. The signature shows correctly in the sidebar
7. I sign the document
8. The signature is not recognizeable because the uploded `.png` is not scaled but in true size. Thus the signature is way to big for the defined signature size

### Expected behavior

The signature appears on the document like in the sidebar

### Screenshots

Image

### Operational System

Debian with Nextcloud AiO

### Browser Type and Version

Firefox 153.0.1

### Libresign Version

14.1.0

### Nextcloud Server Version

34.0.0

no logs related to LibreSign

_No response_

### Additional context

_No response_

Roshan931:
Fixes #7981

## Root cause

Uploaded signature PNG is composited at pixel-size × scale factor instead of being fitted into the stamp box

## Changes

- Resize the user signature with Imagick bestfit into the stamp canvas (same aspect-preserving approach already used for the background in prepareBackgroundForPdf) instead of multiplying its own pixel dimensions by the scale factor, plus a regression test asserting the whole signature stays inside the merged canvas.

### DIFF
--- a/tests/php/Unit/Handler/SignEngine/JSignPdfHandlerTest.php
+++ b/tests/php/Unit/Handler/SignEngine/JSignPdfHandlerTest.php
@@ -727,36 +727,59 @@ final class JSignPdfHandlerTest extends \OCA\Libresign\Tests\Unit\TestCase {
 		$this->assertStringNotContainsString(' -cl ' . DocMdpLevel::CERTIFIED_FORM_FILLING_AND_ANNOTATIONS->name, $paramsSeen[0]);
 	}
 
-	public function testMergeBackgroundWithSignatureFitsOversizedSignatureInsideStampBox(): void {
+	#[DataProvider('providerSignatureDimensions')]
+	public function testMergeBackgroundWithSignatureFitsOversizedSignatureInsideStampBox(int $signatureWidth, int $signatureHeight): void {
 		if (!extension_loaded('imagick')) {
 			$this->markTestSkipped('Extension imagick is not loaded');
 		}
 
-		$this->persistSignatureStampPolicy('', signatureWidth: 350, signatureHeight: 100);
+		$stampWidth = SignatureTextPolicyValue::DEFAULT_SIGNATURE_WIDTH;
+		$stampHeight = SignatureTextPolicyValue::DEFAULT_SIGNATURE_HEIGHT;
+		$scaleFactor = (float)(new \ReflectionClass(JSignPdfHandler::class))->getConstant('SCALE_FACTOR_MIN');
+		$this->persistSignatureStampPolicy('', signatureWidth: $stampWidth, signatureHeight: $stampHeight);
 
 		$backgroundPath = $this->createTransparentPng(10, 10);
-		$signaturePath = $this->createPngWithOpaqueCorners(1400, 400);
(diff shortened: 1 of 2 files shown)


tests/php/Unit/Handler/SignEngine/JSignPdfHandlerTest.php
```diff
@@ -727,6 +727,66 @@ public function testDocMdpSkippedWhenSignatureExists(): void {
...
+
+		$this->persistSignatureStampPolicy('', signatureWidth: 350, signatureHeight: 100);
+
+		$backgroundPath = $this->createTransparentPng(10, 10);
+		$signaturePath = $this->createPngWithOpaqueCorners(1400, 400);
+
+		$mergedPath = self::invokePrivate($this->getInstance(), 'mergeBackgroundWithSignature', [
+			$backgroundPath,
+			$signaturePath,
+			5.0,
```
vitormattos: Would be best to use the constant `SCALE_FACTOR_MIN` because duplicating the value here couples the test to that implementation detail. Ideally the test should exercise the behavior without needing to know the exact minimum value.

tests/php/Unit/Handler/SignEngine/JSignPdfHandlerTest.php
```diff
@@ -727,6 +727,66 @@ public function testDocMdpSkippedWhenSignatureExists(): void {
 		$this->assertStringNotContainsString(' -cl ' . DocMdpLevel::CERTIFIED_FORM_FILLING_AND_ANNOTATIONS->name, $paramsSeen[0]);
 	}
 
+	public function testMergeBackgroundWithSignatureFitsOversizedSignatureInsideStampBox(): void {
```
vitormattos: Maybe it would be good to use `#[DataProvider('signatureDimensionsProvider')]` here to cover different aspect ratios:

```php
public static function signatureDimensionsProvider(): array {
	return [
		'same aspect ratio' => [1400, 400],
		'wider signature' => [1400, 200],
		'taller signature' => [700, 800],
	];
}
```

This would keep the regression case and also verify the new behavior with wider and taller signatures.

vitormattos (changes requested):

tests/php/Unit/Handler/SignEngine/JSignPdfHandlerTest.php
Roshan931: Updated the regression test to read SCALE_FACTOR_MIN from JSignPdfHandler instead of duplicating its value.

tests/php/Unit/Handler/SignEngine/JSignPdfHandlerTest.php
Roshan931: Added a data provider that covers matching, wider, and taller signature aspect ratios.

Issue #973, opened 2022-10-07
Implement translation hints in PHP


vitormattos:
Following the pattern described in Nextcloud [developer manual](https://docs.nextcloud.com/server/latest/developer_manual/basics/front-end/l10n.html#hints), PHP codes that do translations could have hints to help translators to understand the context of sentence to translate.

Example:
```php
<ul id="translations">
    <li id="add-new">
        <?php
            // TRANSLATORS Will be shown inside a popup and asks the user to add a new file
            p($l->t('Add new file'));
        ?>
    </li>
</ul>
```

Suggestion:
Now we have a lot of sentences to put hint. To don't have a very big pull request, do small pull requests, maybe maximum implementing 10 hints.

Using the follow command, you can found the total of rows with translation method in PHP code.
```bash
grep -Rn -e "->t(" lib/ | wc -l
```
To see all the rows, remove `| wc -l`

AntDavidLima:
Hello, I'm applying to Libre Code and would like to start contributing with this repo.

Sinse this issue looks to be one of the simplest one open and is marked as a good first issue, I thing it would be a good start point to get to know the code base.

May I do this task?

lfals:
Workin on this

vitormattos:
Hi @lfals, thanks for working on this!

When adding translation hints, please follow the same principle described in `AGENTS.md`:
- comments should add useful context instead of restating the text.

Write the hints for translators who may not know LibreSign yet. When useful, explain where the text appears, who sees it, or what is happening in the signing workflow.

The goal is that the hint helps the translator understand the LibreSign context, not only the sentence itself.

lfals:
> Hi @lfals, thanks for working on this!
> 
> When adding translation hints, please follow the same principle described in `AGENTS.md`:
> - comments should add useful context instead of restating the text.
> 
> Write the hints for translators who may not know LibreSign yet. When useful, explain where the text appears, who sees it, or what is happening in the signing workflow.
> 
> The goal is that the hint helps the translator understand the LibreSign context, not only the sentence itself.

Sure! You asked for a batch of 10~ per PR, but It has more than 500 hints to translate,What would be the best approach to avoid the 50 PR?

PR #8043, opened 2026-08-24
feat(l10n): add TRANSLATORS hints for controllers and validators

lfals:
Resolves: #973

## 📝 Summary

Adds 150 `TRANSLATORS` comments above PHP `->t()` calls (controllers, validators, account/certificate helpers, and related files) so Transifex translators get LibreSign context (where the string appears, who sees it, and the signing workflow).

This is another batch for #973. After this PR there are still **145** `->t()` calls in `lib/` without a `TRANSLATORS` hint (562 total).

## 🧪 How to test

1. Review the new `// TRANSLATORS ...` comments in the changed PHP files.
2. Confirm each hint sits directly above the related `->t()` / `$l->t()` call.
3. Confirm hints explain LibreSign context and do not only restate the English string.
4. Optionally run:
   ```bash
   grep -Rn -e "->t(" lib/ | wc -l
   ```
   and spot-check that previously missing strings in this batch now have hints.

## ⚙️ API / Back‑end changes

- [x] Added translator hints only; no runtime API or behavior changes
- [ ] Unit and/or integration tests added – *required for backend changes*
- [ ] Capabilities updated (if applicable) – if adding/modifying Nextcloud capabilities
- [ ] API documentation updated with the command `composer openapi` if necessary

### 🚧 Tasks

- [ ] Continue remaining batches until the other **145** `->t()` calls in `lib/` also have hints

## ✅ Checklist

- [x] Hints follow `AGENTS.md` guidance (useful context for translators, not restating the text).
- [x] Scope limited to translator comments (no logic or string changes).

## 🤖 AI (if applicable)


lib/Controller/FileController.php
```diff
@@ -208,6 +208,7 @@ public function validateBinary(): DataResponse {
 			/** @var LibresignActionErrorResponse $response */
 			$response = [
 				'action' => JSActions::ACTION_DO_NOTHING,
+				// TRANSLATORS API error shown when validating a signed document fails unexpectedly; ask the user to contact the server administrator.
```
vitormattos: This message is displayed to the user, so it isn't just an API message. I suggest changing all the comments in this file. What do you think about this?

```suggestion
				// TRANSLATORS Error shown when validating a signed document fails unexpectedly; ask the user to contact the server administrator.
```

lib/Controller/AdminController.php
```diff
@@ -275,6 +275,7 @@ public function installAndValidate(): void {
 			}
 		} catch (\Exception $exception) {
 			$this->eventSource->send('errors', json_encode([
+				// TRANSLATORS Admin SSE error while downloading LibreSign signing binaries (Java, JSignPdf, PDFtk) during setup.
```
vitormattos: Could we make this comment more generic?

SSE is an implementation detail that does not help the translator, and listing Java, JSignPdf and PDFtk may make this comment outdated if these components change. Also, this download can happen outside the initial setup.

Maybe:
```suggestion
				// TRANSLATORS Error shown to an administrator when LibreSign cannot download required components.
```

lib/Controller/LibresignTrait.php
```diff
@@ -81,13 +82,15 @@ public function loadIdDocApprovalFromResolution(array $resolution): void {
...
 				'action' => JSActions::ACTION_DO_NOTHING,
+				// TRANSLATORS API error when loading an identification-document approval request with an invalid identifier.
 				'errors' => [['message' => $this->l10n->t('Invalid identification document request')]],
 			]), AppFrameworkHttp::STATUS_BAD_REQUEST);
 		}
 
 		if (!$resolution['file'] instanceof FileEntity) {
 			throw new LibresignException(json_encode([
 				'action' => JSActions::ACTION_DO_NOTHING,
+				// TRANSLATORS API error when the requested LibreSign file cannot be loaded for the current action.
```
vitormattos: This message could be displayed to the user, so it isn't just an API message. I suggest changing all the comments in this file removing "API". What do you think about this?

lib/Controller/PageController.php
```diff
@@ -242,6 +242,7 @@ public function indexFPath(string $path): TemplateResponse|RedirectResponse {
 				} catch (LibresignException) {
 					throw new LibresignException(json_encode([
 						'action' => JSActions::ACTION_DO_NOTHING,
+						// TRANSLATORS UI error injected into the signing page when the document UUID in the URL is invalid.
```
vitormattos: Who translate this, only need to known that is a message displayed to user. I suggest changing all the comments in this file removing the "UI". What do you think about this?

lib/Controller/NotifyController.php
```diff
@@ -66,6 +66,7 @@ public function signers(int $fileId, array $signers): DataResponse {
 			);
 		}
 		return new DataResponse([
+			// TRANSLATORS API success message after manually notifying a signer about a pending signature request.
```
vitormattos: This message could be displayed to the user, so it isn't just an API message. I suggest changing all the comments in this file removing "API". What do you think about this?

vitormattos (commented):
Only small changes from my side.

Using `API` or `UI` can be useful when it helps explain where the message is shown. But these messages are shown to users, so I think we should avoid technical terms like `API`, `PHP` or `SSE` when they do not help the translator understand the context.

What do you think?

vitormattos (changes requested):

lfals:
> Only small changes from my side.
> 
> Using `API` or `UI` can be useful when it helps explain where the message is shown. But these messages are shown to users, so I think we should avoid technical terms like `API`, `PHP` or `SSE` when they do not help the translator understand the context.
> 
> What do you think?

I agre, pushed a commit removing the technical jargon.

vitormattos (approved):

PR #8067, opened 2026-08-26
fix(lint): review and enable rules from @nextcloud/eslint-config v9 migration

masterlouis:
Related to #7876

## 📝 Summary

Follow-up of #7876 (`@nextcloud/eslint-config` v8 → v9 upgrade). This PR
reviews and progressively enables the ESLint rules newly introduced by
v9, tracked in `libresign/disabled-during-migration` in
`eslint.config.mjs`, one rule per commit as requested in the issue.

⚠️ This branch is built on top of #7876, which is not yet merged into
`main`. Once #7876 lands, this diff will shrink to just the
lint-migration work.

## What's done so far

- **Fixed two config scoping bugs surfaced by the v9 upgrade** (both
  needed before `npm run lint` could even run):
  - `vue/no-unused-components` is now scoped to `files: ['**/*.vue']`
    only. In v9 the `vue` plugin is registered only for `.vue` files,
    so applying this rule globally crashed ESLint on non-Vue files.
  - Removed `import/order` and `import/no-unresolved`, which relied on
    `eslint-plugin-import` — no longer a dependency of
    `@nextcloud/eslint-config` v9. Import ordering is now handled by
    `perfectionist/sort-imports` (already provided by the shared
    config), which will be reviewed as its own rule.
- **Removed obsolete `eslint-disable` comments** referencing rules
  that no longer exist, or no longer apply in their file's context
  (`import/no-named-as-default`, `import/no-unresolved`,
  `jsdoc/require-jsdoc`, an unused `no-new` directive).
- **Enabled `no-console`**: replaced application-level
  `console.error` / `console.warn` calls across 13 files with the
  project's existing `src/logger.js` wrapper around
  `@nextcloud/logger`, passing error/context as a structured object
  (e.g. `logger.error('Failed to save filters', { error })`) instead
  of loose positional arguments. Test files that intentionally
  intercept `console.*` for assertions (`src/tests/setup.js`,
  `src/tests/store/filters.spec.ts`) are excluded via a scoped
  `no-console: off` override, since that's test tooling rather than
  application logging.

## Still to do

Remaining rules in `libresign/disabled-during-migration` (~96 rules)
will be enabled incrementally, one commit per rule, following the
process described in the issue.

## 🧪 How to test

1. `npm ci`
2. `npm run lint` → should report 0 problems
3. `npm run test:unit -- --run` → should pass with no new failures

## ✅ Checklist

- [ ] Work in progress — not ready for final review/merge yet.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -86,7 +86,6 @@ export default [
 			'@stylistic/implicit-arrow-linebreak',
 			'@typescript-eslint/no-use-before-define',
 			'@stylistic/exp-list-style',
-			'no-console',
 			'vue/attributes-order',
 			'@stylistic/quote-props',
 			'vue/padding-line-between-blocks',
@@ -154,5 +153,19 @@ export default [
 			'@stylistic/function-call-spacing',
 			'vue/no-dupe-keys',
 		].map(rule => [rule, 'off'])),
+	},
+		{
+		name: 'libresign/tests-console-override',
+		files: [
+			'src/tests/setup.js',
+			'src/tests/store/filters.spec.ts',
+		],
+		rules: {
+			// These files intentionally intercept console.error/console.warn
+			// to fail tests on unexpected console output, or to silence
+			// expected console output during a specific assertion. This is
+			// test tooling, not application logging.
+			'no-console': 'off',
+		},
 	},
 ]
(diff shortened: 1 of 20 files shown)


vitormattos:
@masterlouis, closing this as it has been superseded by:

- https://github.com/LibreSign/libresign/pull/8101

Your commits were moved to #8101 using cherry-pick.

Issue #8053, opened 2026-08-25
Improve PHP tests with Infection mutation testing


vitormattos:
LibreSign now uses [Infection](https://infection.github.io/) to help check the quality of the PHP tests.

Code coverage tells us which code was executed by tests. However, it does not tell us if the tests can detect wrong behavior. Mutation testing helps us check this.

Infection makes small changes to the source code. These changes are called **mutants**. After each change, Infection runs the tests again:

* If a test fails, the mutant is **killed**. This is good because the test detected the change.
* If all tests still pass, the mutant **escaped**. This can mean that an important behavior is not well tested.
* A mutant can also cause a timeout or an error. These cases need to be checked before deciding what should be changed.

The first full LibreSign run generated more than 15,000 mutants and took a long time to run. It also found many escaped mutants, errors and timeouts.

Initial run:

https://github.com/LibreSign/libresign/actions/runs/32801594308/job/97663483870

Because of this, the full Infection suite is not ready to be required in CI yet.

This issue tracks the work to improve the PHP code and tests step by step.

> [!IMPORTANT]
> Do not solve this issue in one PR.

This issue is open to many contributions.

**Each PR should work on only one or two related PHP source/test pairs.**

For example:

```text
lib/Service/ExampleService.php
tests/php/Unit/Service/ExampleServiceTest.php
````

A PR can include two pairs when the classes are small and closely related.

Keep the change focused. As a general rule, try to keep the diff below about 1,000 lines. If it gets close to 2,000 lines, split the work into another PR.

Partial PRs should reference this issue, but should not use `Fixes #...` or `Closes #...` while there is still work to do.

Before starting, please comment on this issue with the source file you want to work on. This helps avoid two contributors working on the same code.

## Effort

The complete issue is a **high-effort** task because it covers many PHP files and tests.

However, each contribution should be **low to medium effort**. A PR should focus on only one or two related source/test pairs.

You do not need to solve the complete issue. Small and focused improvements are welcome and can be merged separately.

## What should be improved first

Please use this order when choosing what to work on:

1. **Timeouts or unsafe side effects** found by Infection. These can make Infection slow or unstable and should be checked first.
2. **Escaped mutants in important business rules**, especially signing, certificates, policies, permissions, validation and identity flows.
3. Escaped mutants in other services and controllers.
4. Other useful code or test improvements found while working on the selected file.

A fatal error caused by a mutant does not always mean that a test needs to change. Check the reason first.

[...]

PR #8092, opened 2026-08-28
test: cover EmailToken and TokenService signature method behaviors

maia-andre:
Ref: #8053 

## 📝 Summary

Two closely related source/test pairs from #8053 (as claimed on the issue):

- `lib/Service/IdentifyMethod/SignatureMethod/TokenService.php`
- `lib/Service/IdentifyMethod/SignatureMethod/EmailToken.php`

Infection baseline for the two files: 23 mutants generated, 1 escaped, Covered Code MSI 95%. The escaped mutant is the `MethodCallRemoval` of `$this->twofactorGatewayService->ensureAvailable($gatewayName)` in `sendCodeByGateway()` — no scenario exercised an unavailable gateway. One correction to my claim comment: `sendCodeByEmail()`, `requestCode()` and `validateToSign()` did have *indirect* coverage from elsewhere in the unit suite; what was missing were direct tests pinning their observable behavior, plus the escaped kill.

This PR only adds tests — no production code changes:

- `testSendCodeByGatewayThrowsWhenGatewayAppIsNotEnabled`: kills the escaped mutant. With the call in place, `ensureAvailable()` throws `LibresignException("App Two-Factor Gateway is not enabled.")`; with the call removed, execution reaches `isGatewayComplete()` and throws `OCSForbiddenException` instead, so the expected exception class/message pins the guard.
- `testSendCodeByEmailSendsCodeAndReturnsHashedCode`: direct test asserting code generation (`TOKEN_LENGTH`, digits), delivery via `MailService::sendCodeToSign()` and the hashed return value.
- `EmailToken::requestCode()`: both display-name branches (distinct name is forwarded; name equal to the identifier is sent as an empty string), asserting the hashed code is persisted via `IdentifyService::save()`.
- `EmailToken::validateToSign()`: valid code passes; wrong code throws `Invalid code.`; a code sent without ever being requested also throws.
- Typo fix: `testVaidateEmail` → `testValidateEmail` (and its data provider); removed a dead mock assignment in `setUp()`.

After the change, Infection reports **29 mutants generated, 0 escaped, 0 uncovered — Covered Code MSI 100%** for this scope (17 killed by tests, 12 detected via fatal errors).

## 🧪 How to test

```sh
composer test:unit -- --filter "TokenServiceTest|EmailTokenTest"

XDEBUG_MODE=coverage vendor/bin/infection --configuration=infection.json5 \
  --test-framework-options="--testsuite=unit" --show-mutations \
  lib/Service/IdentifyMethod/SignatureMethod/TokenService.php \
  lib/Service/IdentifyMethod/SignatureMethod/EmailToken.php
```

Expected: tests green; 29 mutants generated, 0 escaped, Covered Code MSI 100%.

## ⚙️ API / Back‑end changes

Test-only change; no production code, capabilities, or API documentation affected.

## ✅ Checklist

- [x] Full unit suite, focused Infection and `php-cs-fixer` pass for the changed files.

## 🤖 AI (if applicable)

### DIFF
--- a/tests/php/Unit/Service/IdentifyMethod/SignatureMethod/EmailTokenTest.php
+++ b/tests/php/Unit/Service/IdentifyMethod/SignatureMethod/EmailTokenTest.php
@@ -9,10 +9,14 @@ declare(strict_types=1);
 namespace OCA\Libresign\Tests\Unit\Service\IdentifyMethod\SignatureMethod;
 
 use OCA\Libresign\Db\IdentifyMethod;
+use OCA\Libresign\Db\SignRequest;
+use OCA\Libresign\Db\SignRequestMapper;
+use OCA\Libresign\Exception\LibresignException;
 use OCA\Libresign\Service\IdentifyMethod\IdentifyService;
 use OCA\Libresign\Service\IdentifyMethod\SignatureMethod\EmailToken;
 use OCA\Libresign\Service\IdentifyMethod\SignatureMethod\TokenService;
 use OCP\L10N\IFactory as IL10NFactory;
+use OCP\Security\IHasher;
 use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\MockObject\MockObject;
 
@@ -22,10 +26,9 @@ final class EmailTokenTest extends \OCA\Libresign\Tests\Unit\TestCase {
 
 	#[\Override]
 	public function setUp(): void {
-		$identifyService = $this->createMock(IdentifyService::class);
 		$identifyService = $this->getMockBuilder(IdentifyService::class)
 			->disableOriginalConstructor()
-			->onlyMethods(['getL10n'])
+			->onlyMethods(['getL10n', 'getSignRequestMapper', 'getHasher', 'save'])
 			->getMock();
 		$identifyService->method('getL10n')->willReturn(
 			\OCP\Server::get(IL10NFactory::class)->get(\OCA\Libresign\AppInfo\Application::APP_ID)
@@ -41,8 +44,8 @@ final class EmailTokenTest extends \OCA\Libresign\Tests\Unit\TestCase {
 		);
 	}
 
-	#[DataProvider('providerVaidateEmail')]
-	public function testVaidateEmail(string $email, string $blurred, string $hash): void {
+	#[DataProvider('providerValidateEmail')]
+	public function testValidateEmail(string $email, string $blurred, string $hash): void {
 		$instance = $this->getClass();
 		$identifyMethod = new IdentifyMethod();
 		$entity['identifierKey'] = 'email';
@@ -60,7 +63,7 @@ final class EmailTokenTest extends \OCA\Libresign\Tests\Unit\TestCase {
 		$this->assertEquals($hash, $actual['hashOfEmail']);
 	}
 
-	public static function providerVaidateEmail(): array {
+	public static function providerValidateEmail(): array {
 		return [
 			['valid@domain.coop', 'val***@***.coop', md5('valid@domain.coop')],
 			['valiD@Domain.coop', 'val***@***.coop', md5('valid@domain.coop')],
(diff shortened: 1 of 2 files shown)


tests/php/Unit/Service/IdentifyMethod/SignatureMethod/EmailTokenTest.php
```diff
@@ -154,4 +157,117 @@ public static function providerToArrayWithValidData(): array {
...
+			'identifierKey' => 'email',
+			'identifierValue' => 'valid@domain.coop',
+			'signRequestId' => 171,
+		]);
+		$instance->setEntity($identifyMethod);
+		$this->identifyService->expects($this->once())
+			->method('save')
+			->with($identifyMethod);
+
+		$instance->requestCode('valid@domain.coop', 'email');
```
vitormattos: It is good to have this test scenario, but it would be better to cover the `TokenService->sendCodeByEmail` method, as that is where the code is generated.

tests/php/Unit/Service/IdentifyMethod/SignatureMethod/TokenServiceTest.php
```diff
@@ -43,6 +44,19 @@ public function setUp(): void {
...
 
+	public function testSendCodeByGatewayThrowsWhenGatewayAppIsNotEnabled(): void {
+		$this->appManager->method('isEnabledForAnyone')->with('twofactor_gateway')->willReturn(false);
+		$this->container->expects($this->never())
+			->method('get');
+		$this->secureRandom->expects($this->never())
+			->method('generate');
+
+		$this->expectException(LibresignException::class);
+		$this->expectExceptionMessage('App Two-Factor Gateway is not enabled.');
```
vitormattos, line 55: It would be a good idea to avoid adding translatable or easily mutable text to test scenarios. If a non-developer needs to update the text, it becomes complicated because they would need to know to change it in two different places. 

The `LibresignException` class could use an exception code instead, allowing the use of the `expectExceptionCode` method, but is necessary to check the best approach.

tests/php/Unit/Service/IdentifyMethod/SignatureMethod/EmailTokenTest.php
```diff
@@ -154,4 +157,117 @@ public static function providerToArrayWithValidData(): array {
...
+		$identifyMethod = (new IdentifyMethod())->fromParams([
+			'identifierKey' => 'email',
+			'identifierValue' => 'valid@domain.coop',
+			'code' => 'hashed-code',
+		]);
+		$instance->setEntity($identifyMethod);
+		$instance->setCodeSentByUser('654321');
+
+		$this->expectException(LibresignException::class);
+		$this->expectExceptionMessage('Invalid code.');
```
vitormattos, line 254: It would be a good idea to avoid adding translatable or easily mutable text to test scenarios. If a non-developer needs to update the text, it becomes complicated because they would need to know to change it in two different places. 

The `LibresignException` class could use an exception code instead, allowing the use of the `expectExceptionCode` method, but is necessary to check the best approach.

vitormattos:
But I've already merged it, so go ahead!

tests/php/Unit/Service/IdentifyMethod/SignatureMethod/TokenServiceTest.php
maia-andre: Agreed — asserting on the message couples the test to text that can change for non-technical reasons, and in `EmailTokenTest` it is even more fragile because the string goes through the real L10N factory.

Two things I checked:

- The mutants this PR kills do not depend on the message. In `TokenServiceTest`, if `ensureAvailable()` is removed the flow reaches `isGatewayComplete()` and throws `OCSForbiddenException`, which is not a `LibresignException`, so `expectException(LibresignException::class)` alone already distinguishes the mutant. Same for `EmailTokenTest` (`hasher->verify()` returning false + the exception class).
- So the immediate fix is tests-only: drop the three `expectExceptionMessage()` calls and keep the exception class assertion. I'll open a follow-up PR against `main` with that and re-run Infection to confirm MSI stays at 100% (and backport it to stable35 after #8102).

For the long-term approach I see two directions, and I'd rather not pick one on my own since both touch production code:

- exception codes on `LibresignException` + `expectExceptionCode()`. Today 47 of the 287 `throw new LibresignException` carry a code, mixing HTTP-like values (404/422/400/401) with `1`, and the code is exposed through `ErrorPayloadBuilder` as `code`/`error_code`, so a convention would also touch the API contract;
- typed subclasses (e.g. `InvalidCodeException extends LibresignException`) so tests assert the class instead of the text.

The suite currently has 171 `expectExceptionMessage()` calls across 51 files, so either direction is a broader effort. Would you like me to open an issue with this survey so it can be discussed and split into small PRs, like #8053?

tests/php/Unit/Service/IdentifyMethod/SignatureMethod/EmailTokenTest.php
maia-andre: Same point applies here — answered in the `TokenServiceTest` thread above; the follow-up PR will cover both files.

maia-andre:
About the failed backports to stable32–34: the cherry-pick cannot apply there because `TokenServiceTest.php` and the `ensureAvailable()` call in `TokenService` only exist in `main`/`stable35` (both came with 24b9625f7). In stable32–34 `TokenService` is the older implementation (`getGateway()` through the container), so that half of the PR has nothing to backport.

The `EmailTokenTest` half does apply: `EmailToken::requestCode()` and `AbstractIdentifyMethod::throwIfInvalidToken()` are identical in those branches. If you want it, I can open manual backports with only the `EmailTokenTest` changes for stable33 and stable34 (and stable32 too, if it is still worth it before EOL). Just let me know.

tests/php/Unit/Service/IdentifyMethod/SignatureMethod/TokenServiceTest.php
vitormattos: Only a curiosity:

> mixing HTTP-like values

At some places we already do this but to prevent mistakes, we added a more digit.

I found some developers considering that 4xx is a HTTP error, not another kind of error.

vitormattos: As example:

https://github.com/LibreSign/libresign/blob/b33a20a0ff95fbef162782360d32521da0387c75/lib/Helper/JSActions.php#L12-L22

https://github.com/LibreSign/libresign/blob/b33a20a0ff95fbef162782360d32521da0387c75/src/helpers/ActionMapping.ts#L30-L42

vitormattos: > Would you like me to open an issue with this survey so it can be discussed and split into small PRs, like https://github.com/LibreSign/libresign/issues/8053?

Nice idea, could you do this?

maia-andre: Done: #8117 has the survey and a proposal to settle the convention (codes with constants and/or the existing typed exceptions) before splitting the work into small PRs.

vitormattos:
> If you want it, I can open manual backports with only the EmailTokenTest changes for stable33 and stable34 (and stable32 too, if it is still worth it before EOL). Just let me know.

I don't think that's necessary, considering it's a very small test scenario that doesn't impact any features. If it works on main, the probability that it also works on a stable branch is very high.

We can consider this PR as done. Thanks a lot by your contribution!

PR #8103, opened 2026-08-28
refactor(exception): add error codes to the invalid token and gateway exceptions

maia-andre:
### Pull Request Description

Follow-up to #8092, addressing the review comments there and in this PR: the tests added in #8092 asserted on exception messages (`'Invalid code.'` and `'App Two-Factor Gateway is not enabled.'`), which couples them to translatable/easily mutable text, and the suggestion was to identify these errors by an exception code instead.

**`refactor(exception)`** — adds two constants to `LibresignException` and uses them at the throw sites:

| Constant | Value | Thrown by |
| --- | --- | --- |
| `CODE_INVALID_TOKEN` | `4010` | `AbstractIdentifyMethod::throwIfInvalidToken()` — wrong or unsolicited verification code |
| `CODE_TWOFACTOR_GATEWAY_NOT_ENABLED` | `5030` | `TwofactorGatewayService` — the Two-Factor Gateway app is disabled |

The values follow the closest HTTP status with one extra digit, as discussed in #8092, so they are not mistaken for HTTP statuses.

Where the codes end up (checked before choosing the values):

- Both throw sites are only reached through `SignFileController::sign()` and `requestCode*()`, which catch the exception and always answer **422** — the HTTP status does not change.
- `SigningErrorHandler` exposes the code in the payload (`errors[].code`, `0` until now) and picks the action by `code === 400`, so the action does not change either; the frontend only compares the code with `422` (`NON_RETRIABLE_SIGN_ERROR_CODE`).
- `InjectionMiddleware::getStatusCodeFromException()` still uses any non-zero code as the HTTP status. These two exceptions never reach it, but it means codes like these (and the existing `1`) must not leak there — a guard (code outside 100–599 → 422) is worth discussing in #8117.

**`test`** (two commits) — replaces the three `expectExceptionMessage()` calls with `expectExceptionCode()` on the new constants; and covers `EmailToken::toArray()` for the `emailToken`, `account` and unknown identifier keys, which the existing tests never exercised (re-measuring Infection with a single thread showed 4 escaped `MatchArmRemoval` mutants there — the earlier "0 escaped" in this description came from a multi-threaded run, which reports false kills for classes extending the LibreSign `TestCase`).

The pre-existing assertion on `'Gateway sms not configured on Two-Factor Gateway.'` was kept: it checks that the gateway name is interpolated into the message, which is behavior rather than wording.

Validation: PHPUnit `tests/php/Unit/Service/IdentifyMethod` (163 tests) + `TwofactorGatewayServiceTest` (6) green; php-cs-fixer clean; Infection on `EmailToken.php` and `TokenService.php` (single thread, whole unit suite): 31 mutants, 0 escaped, Covered Code MSI 100%.

Backport notes: applies cleanly to stable35 (both commits). Not applicable to stable32–34 — `TwofactorGatewayService.php` and `TokenServiceTest.php` do not exist there.

### Related Issue

Ref #8053, #8117

### Pull Request Type

- Refactor
- Test

### Pull request checklist

- [x] Tests pass locally (PHPUnit + Infection)
- [x] AI-assisted: yes (see commit trailers)

### DIFF
--- a/tests/php/Unit/Service/IdentifyMethod/SignatureMethod/EmailTokenTest.php
+++ b/tests/php/Unit/Service/IdentifyMethod/SignatureMethod/EmailTokenTest.php
@@ -251,7 +251,7 @@ final class EmailTokenTest extends \OCA\Libresign\Tests\Unit\TestCase {
 		$instance->setCodeSentByUser('654321');
 
 		$this->expectException(LibresignException::class);
-		$this->expectExceptionMessage('Invalid code.');
+		$this->expectExceptionCode(LibresignException::CODE_INVALID_TOKEN);
 
 		$instance->validateToSign();
 	}
@@ -266,7 +266,7 @@ final class EmailTokenTest extends \OCA\Libresign\Tests\Unit\TestCase {
 		$instance->setCodeSentByUser('123456');
 
 		$this->expectException(LibresignException::class);
-		$this->expectExceptionMessage('Invalid code.');
+		$this->expectExceptionCode(LibresignException::CODE_INVALID_TOKEN);
 
 		$instance->validateToSign();
 	}
--- a/tests/php/Unit/Service/IdentifyMethod/SignatureMethod/EmailTokenTest.php
+++ b/tests/php/Unit/Service/IdentifyMethod/SignatureMethod/EmailTokenTest.php
@@ -15,6 +15,8 @@ use OCA\Libresign\Exception\LibresignException;
 use OCA\Libresign\Service\IdentifyMethod\IdentifyService;
 use OCA\Libresign\Service\IdentifyMethod\SignatureMethod\EmailToken;
 use OCA\Libresign\Service\IdentifyMethod\SignatureMethod\TokenService;
+use OCP\IUser;
+use OCP\IUserManager;
 use OCP\L10N\IFactory as IL10NFactory;
 use OCP\Security\IHasher;
 use PHPUnit\Framework\Attributes\DataProvider;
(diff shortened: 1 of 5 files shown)


tests/php/Unit/Service/IdentifyMethod/SignatureMethod/EmailTokenTest.php
```diff
@@ -251,7 +251,6 @@ public function testValidateToSignWithWrongCodeThrows(): void {
 		$instance->setCodeSentByUser('654321');
 
 		$this->expectException(LibresignException::class);
-		$this->expectExceptionMessage('Invalid code.');
```
vitormattos, line 254: What does you think about add an error code to this exception?

With an error code we can confirm that is the expected error without matching the text.

maia-andre: Agreed — I'll add a code so the tests assert `expectExceptionCode()` instead of the text. Two quick questions before I push:

1. Should the code follow the "HTTP-like plus one digit" pattern you mentioned (e.g. `4010` for the invalid verification code and `5030` for "Two-Factor Gateway not enabled"), and would you like them as constants, like `JSActions`?
2. I'd apply it to both exceptions touched by this PR, unless you prefer only `Invalid code.` for now.

vitormattos: > Should the code follow the "HTTP-like plus one digit" pattern

I prefer this because it reduces misunderstanding. In the past, we had a contributor who thought these were standard HTTP codes, which took some effort to clarify.

maia-andre: Done: `LibresignException::CODE_INVALID_TOKEN` (4010) and `LibresignException::CODE_TWOFACTOR_GATEWAY_NOT_ENABLED` (5030), set at the two throw sites and asserted with `expectExceptionCode()`. I made them constants on the exception class so the tests reference the name rather than the number — easy to move if you prefer another home for them.

While tracing where the codes end up I found that `InjectionMiddleware::getStatusCodeFromException()` uses any non-zero code as the HTTP status. These two exceptions never reach it (both paths answer a fixed 422), but the same would not hold for a `LibresignException(…, 4010)` thrown elsewhere, so I noted it in the description as a point for #8117.

vitormattos (approved):

PR #8106, opened 2026-08-28
feat: add mail_sender_strategy policy to send notifications as the requester

### LINKED ISSUE #7954: [FEATURE] When sending a sign request send from the users email rather than the system email.
### Is your feature request related to a problem? Please describe.

The user who is requesting a document to be signed would be the best person to send the email to the person signing as they will have already received emails from that user and therefore less likely to end up in junk. Plus any response goes to the users email account.

### Describe the solution you'd like

Libre sign to use the users default email address in the mail app on nextcloud to send the document

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

maia-andre:
### Pull Request Description

Implements the `mail_sender_strategy` policy described in #7954 so LibreSign can send signature request notification emails either through the system mailer (current behavior) or through the mail account of the person who requested the signature.

**Backend**

- New `MailSenderStrategyPolicy` (`lib/Service/Policy/Provider/MailSenderStrategy/`), registered in `PolicyProviders`. Values: `system` (default) and `requester`. System scope only in this first step (`supportedScopes: ['system']`, no user preference, no group delegation), since per-user choice is out of scope.
- `MailService::notifyUnsignedUser()` and `notifySignDataUpdated()` now resolve the policy for the **requester** (`resolveForUserId()` on the file owner, so the same strategy applies to reminders sent from cron, where there is no session user).
- Under `requester`, the notification is sent through `OCP\Mail\Provider\IManager`: the service is looked up by the requester's email address (`findServiceByAddress()`) and must be able to send messages (`IMessageSend`). No other account of the requester is used: `services()` gives no way to tell which one is the main account, so guessing could send the request from the wrong one. The message reuses the existing `IEMailTemplate` (`renderSubject()`, `renderHtml()`, `renderText()`), so content and templating are unchanged.
- Fallback to `IMailer` (identical to the current code path) when there is no mail provider, the requester account is missing, no mail account matches the requester's email address (or the matching one cannot send messages), or provider sending throws. Logging: `info` for the "not available" cases, `warning` (with the exception) when provider sending fails. Errors from the final `IMailer` path keep raising `LibresignException` as before.

**Configuration vs. runtime (follow-up of the review in #7954)**

- The `requester` strategy can only be **saved** while a mail provider is registered (`OCP\Mail\Provider\IManager::has()`, independent of any specific app). This uses a new persistence-only hook in the policy framework, `IPolicyDefinition::validateValueForPersistence()`, called from the `PolicyService` save paths. Runtime resolution keeps using `validateValue()`, so an already stored `requester` value is not silently discarded by `DefaultPolicyResolver` when the environment changes later; instead `MailService` falls back at sending time.
- The availability is exposed as `meta.mailProviderAvailable` (OpenAPI spec and generated types updated) and the Policy Workbench editor disables the `requester` option with a hint when no provider is available.
- When the strategy is `requester` and the notification falls back to the system mailer, `Reply-To` is set to the requester's address, so replies still reach the person who requested the signature. The `system` strategy is unchanged.

**Frontend**

- Policy Workbench definition under `src/views/Settings/PolicyWorkbench/settings/mail-sender-strategy/` (radio editor with the two strategies), added to the `system-behavior` category in `realDefinitions.ts`.

**Tests**

- `MailSenderStrategyPolicyTest`: definition, scope restriction, normalization and validation, provider required on save (and stored value still valid at runtime), resolved meta.
- `PolicySpecTest`: the persistence-only validation hook.
- `MailServiceTest`: `system` strategy, successful `requester` sending (from/to/subject/html/plain), recipient label handling, every fallback path (no provider, unknown requester, missing account or email address, no account matching the address, matching account unable to send, provider failure, fallback failure) and the `Reply-To` behavior on fallback.
- `realDefinitions.spec.ts`, `mail-sender-strategy/*.spec.ts`: workbench card, definition callbacks, model and editor (including the disabled state).

Checks run locally: PHPUnit unit suite, psalm and php-cs-fixer on the changed files, `npm run lint`, `vue-tsc --noEmit`, vitest for the Policy Workbench suites.

### Manual verification

Devcontainer with Mailpit as the system mailer and **no** mail provider app installed, so the `requester` strategy exercises the fallback path end to end.

1. New card in the Policy Workbench (System behavior), default `System mailer`:

   Workbench card with the default value

2. Rule editor with the two strategies:

   Rule editor options

3. Selecting `Requester mail account` and saving the system rule:

   Requester selected

   Policy dialog after saving

4. Card after saving (`occ config:app:get libresign mail_sender_strategy` → `requester`):

   Workbench card with requester

5. Signature request sent through the API with the policy set to `requester`: the notification is still delivered, through the system mailer, and the fallback is logged at `info` level:

   Mailpit delivery

   ```json
   {"level": 1, "app": "libresign", "message": "No mail provider is available to send the notification as the requester, falling back to the system mailer.", "requester": "admin"}
   ```

6. Follow-up of the review: on the same instance (no mail provider), the `requester` option can no longer be configured. The editor disables it with a hint, and the API rejects it with `400`:

   Requester option disabled without a mail provider

   ```
   POST /ocs/v2.php/apps/libresign/api/v1/policies/system/mail_sender_strategy  {"value":"requester"}
   HTTP 400  {"ocs":{"meta":{"status":"failure","statuscode":400},"data":{"error":"The requester strategy requires an available mail provider"}}}
   ```

   The screenshots in steps 3–5 were taken before this change; they show the rule saved while the check did not exist yet, which is exactly the "environment changed later" case that the runtime fallback (and now `Reply-To`) covers.

### Related Issue

Fixes #7954

### Pull Request Type

- Feature

### Pull request checklist

- [x] The policy supports `system` and `requester`
- [x] The policy is available in the backend and Policy Workbench
- [x] `system` preserves the current behavior
- [x] `requester` attempts provider-based sending
- [x] Falls back to `IMailer` when no provider, account or sending service is available, or when sending fails
- [x] `requester` cannot be configured while no mail provider is available (persistence-time check)
- [x] `Reply-To` set to the requester when falling back to the system mailer
- [x] Applied to `notifyUnsignedUser` and `notifySignDataUpdated`
- [x] Tests cover `system`, successful `requester` and fallback
- [x] AI-assisted: yes (see commit trailer)

### DIFF
--- a/lib/Service/MailService.php
+++ b/lib/Service/MailService.php
@@ -12,10 +12,19 @@ use OCA\Libresign\Db\File;
 use OCA\Libresign\Db\FileMapper;
 use OCA\Libresign\Db\SignRequest;
 use OCA\Libresign\Exception\LibresignException;
+use OCA\Libresign\Service\Policy\PolicyService;
+use OCA\Libresign\Service\Policy\Provider\MailSenderStrategy\MailSenderStrategyPolicy;
 use OCP\IAppConfig;
 use OCP\IL10N;
 use OCP\IURLGenerator;
+use OCP\IUser;
+use OCP\IUserManager;
+use OCP\Mail\IEMailTemplate;
 use OCP\Mail\IMailer;
+use OCP\Mail\Provider\Address;
+use OCP\Mail\Provider\IManager as IMailProviderManager;
+use OCP\Mail\Provider\IMessageSend;
+use OCP\Mail\Provider\IService;
 use Psr\Log\LoggerInterface;
 
 class MailService {
@@ -29,6 +38,9 @@ class MailService {
 		private IL10N $l10n,
 		private IURLGenerator $urlGenerator,
 		private IAppConfig $appConfig,
+		private PolicyService $policyService,
+		private IMailProviderManager $mailProviderManager,
+		private IUserManager $userManager,
 	) {
 	}
 
@@ -67,15 +79,8 @@ class MailService {
 			$this->l10n->t('Sign "%s"', [$file->getName()]),
 			$link
 		);
-		$message = $this->mailer->createMessage();
-		if ($data->getDisplayName()) {
-			$message->setTo([$email => $data->getDisplayName()]);
-		} else {
-			$message->setTo([$email]);
-		}
-		$message->useTemplate($emailTemplate);
 		try {
-			$this->mailer->send($message);
+			$this->sendSignRequestNotification($emailTemplate, $data, $email);
 		} catch (\Exception $e) {
 			$this->logger->error('Notify changes in unsigned notification mail could not be sent: ' . $e->getMessage());
 			throw new LibresignException('Notify unsigned notification mail could not be sent', 1);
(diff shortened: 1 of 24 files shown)


lib/Service/MailService.php
```diff
@@ -107,19 +112,127 @@ public function notifyUnsignedUser(SignRequest $data, string $email, ?string $de
...
+		} catch (\Throwable $e) {
+			$this->logger->warning('Unable to send the notification through the requester mail account, falling back to the system mailer.', [
+				'requester' => $requesterId,
+				'exception' => $e,
+			]);
+			return false;
+		}
+	}
+
+	private function findRequesterMailService(string $requesterId, ?IUser $requester): (IService&IMessageSend)|null {
```
vitormattos, line 207: I'm not sure we should use the first mail service that can send messages when no account matches the requester's email address.

`services()` can return more than one mail account, and I don't think we can be sure that the first one is the user's main account. This could make LibreSign send a signature request from the wrong account.

For example, the requester could use their company email in Nextcloud, but also have a personal mail account configured.

If Nextcloud has a clear way to get the user's main mail account, we can use it here. If not, I think we should remove this fallback and use the system mailer when `findServiceByAddress()` does not find the correct account.

Please also update the unit test to check this fallback behavior.

vitormattos (changes requested):

lib/Service/MailService.php
maia-andre: Agreed. OCP's `IManager` has no notion of a main account (only `services()`, `findServiceById()` and `findServiceByAddress()`), so I removed the fallback: under `requester` the notification is sent only through the account returned by `findServiceByAddress()` for the requester's email address. When no account matches that address, or the matching account cannot send messages, it falls back to the system mailer like the other cases (with `Reply-To` set to the requester).

Tests updated: `testRequesterStrategyFallsBackWhenNoAccountMatchesTheRequesterAddress`, `testRequesterStrategyFallsBackWhenTheMatchingAccountCannotSend` and `testRequesterStrategyFallsBackWhenTheRequesterHasNoEmailAddress`, all asserting that `services()` is never called. The PR description was updated accordingly.

PR #8120, opened 2026-08-29
Feat/signer geolocation backend 6960

### LINKED ISSUE #6960: Add backend support for signer geolocation
LibreSign already stores signer audit metadata such as IP address and User-Agent.

Some organizations also need device-reported geolocation as additional audit information.

This issue adds the backend support for this feature. Frontend integration and IP-based geolocation will be handled separately.

### Goal

Add backend support to:

* configure device geolocation through the existing LibreSign policy system;
* support `disabled`, `optional`, and `required` modes;
* allow administrators to decide whether requesters may require geolocation for individual signers;
* persist the effective requirement per signer;
* accept and validate geolocation data from the signing client;
* store it in signer metadata;
* expose stored geolocation through the appropriate APIs;
* enforce required geolocation server-side.

### Policy

Add a dedicated geolocation policy using the existing LibreSign policy architecture.

Default values:

* device geolocation: `disabled`
* requester configuration: `disabled`

Expected behavior:

* `disabled`: geolocation is not used;
* `optional`: geolocation may be collected but does not block signing;
* `optional` + requester configuration enabled: the requester may require geolocation for selected signers;
* `required`: geolocation is mandatory and cannot be weakened by the requester.

The policy must follow the existing scope, inheritance, delegation, and override rules.

### Per-signer requirement

When requester configuration is allowed, store the final geolocation requirement for each signer.

This requirement must remain stable for an existing signature request even if the policy changes later.

Follow the existing policy snapshot approach where appropriate.

### API and enforcement

Extend the appropriate APIs so clients can:

* read the effective geolocation policy;
* know whether requester configuration is allowed;
* configure the requirement for a signer when permitted;
* submit device geolocation metadata;
* retrieve stored signer geolocation.

The backend must validate all submitted configuration.

When geolocation is required, the signing operation must be rejected if valid geolocation data is missing.

When geolocation is optional, missing location must not prevent signing.

Frontend validation alone is not sufficient.

### Signer metadata

Store geolocation in the metadata of the signer who completed the signature.

Store at least, when available:

* collection status;
* latitude;
* longitude;
* accuracy;
* position timestamp.

Device geolocation must be treated as device-reported information, not verified proof of the signer's physical location.

### Relationship with `collect_metadata`

LibreSign already has a `collect_metadata` policy.

Define and test how the new geolocation policy interacts with it, avoiding contradictory effective configurations.

### Tests

Add backend coverage for:

* policy defaults and resolution;
* inheritance and overrides;
* `disabled`, `optional`, and `required`;
* requester configuration;
* prevention of weakening mandatory requirements;
* per-signer persistence;
* API validation;
* required geolocation enforcement;
* optional geolocation behavior;
* signer metadata persistence and retrieval;
* multiple signers;
* existing requests remaining stable after policy changes.

Use data providers where they reduce duplicated test scenarios.

### Mutation testing

New feature-specific backend logic must be covered by Infection.

For reasonably isolated new classes:

* no escaped mutants;
* no mutation errors;
* target 100% MSI.

Do not weaken Infection configuration or exclude production code to make the tests pass.

### Quality gates

The implementation must pass the existing backend checks, including:

* PHPUnit;
* Behat/integration tests;
* Psalm;
* PHPCS;
* Infection.

### Out of scope

* browser `navigator.geolocation`;
* Policy Workbench frontend UI;
* requester-facing signer configuration UI;
* geolocation visualization;
* IP/GeoIP-based geolocation.

These will be handled in separate follow-up issues.

lfals:
Resolves: #6960

## 📝 Summary

Adds backend support for the `signer_geolocation` policy, allowing administrators to control whether signer device geolocation is disabled, optional, or required when signing documents.

The policy follows existing Policy Workbench patterns (`collect_metadata`, `identification_documents`) with three modes:

- `disabled`: geolocation cannot be requested for any signer
- `optional`: the requester can choose which individual signers must provide geolocation via `geolocationRequired`
- `required`: all signers must provide geolocation; the requester cannot disable it for any signer

At request creation, the effective per-signer requirement is frozen in `metadata.geolocationRequirement` as only `disabled` or `required`. `optional` is a policy mode, not a stored signer requirement.

Sign endpoints accept a `geolocation` payload, validate it server-side, and persist coordinates independently of `collect_metadata`. The file list API exposes only geolocation-related signer metadata (`geolocationRequirement` and `geolocation`) from the stored sign request, not unrelated signer metadata.

OpenAPI specs and generated TypeScript types were updated to document the new policy and sign/request parameters.

## 🧪 How to test

1. Set the system policy to require geolocation:
   ```bash
   curl -u admin:admin -H 'OCS-ApiRequest: true' -H 'Content-Type: application/json' \
     -X POST 'http://localhost/ocs/v2.php/apps/libresign/api/v1/policies/system/signer_geolocation' \
     -d '{"value":{"mode":"required"}}'
   ```
2. Create a signature request with at least one signer.
3. List the file with details (`GET /api/v1/file/list?details=1`) and confirm the signer has `metadata.geolocationRequirement: "required"`.
4. Attempt to sign without `geolocation` and confirm the API returns `422` with a geolocation-required error.
5. Sign again with a valid payload:
   ```json
   {
     "method": "clickToSign",
     "geolocation": {
       "status": "collected",
       "latitude": -23.5505,
       "longitude": -46.6333,
       "accuracy": 25,
       "timestamp": 1700000000000
     }
   }
   ```
6. Confirm signing succeeds and the stored signer metadata includes `geolocation` coordinates.
7. Repeat with `mode: "optional"`, creating a request with `geolocationRequired: true` on selected signers and `geolocationRequired: false` on others. Verify only the selected signers get `metadata.geolocationRequirement: "required"` and only those signers are enforced at sign time.
8. Run focused automated checks:
   ```bash
   composer test:unit -- --filter SignerGeolocation
   vendor/bin/behat features/sign/signer_geolocation.feature -v
   vendor/bin/behat features/policies/signer_geolocation_policy.feature -v
   npx playwright test playwright/e2e/signer-geolocation-backend.spec.ts
   ```

## ⚙️ API / Back-end changes

- [x] Add `signer_geolocation` policy provider with modes `disabled`, `optional`, and `required`
- [x] Freeze per-signer `geolocationRequirement` (`disabled` or `required`) on signature request creation via `SignerGeolocationFilePolicyApplier`
- [x] Accept and validate `geolocation` on sign endpoints; reject missing/invalid submissions when required
- [x] Persist geolocation coordinates in signer metadata independently of `collect_metadata`
- [x] Expose only geolocation-related signer metadata in request/sign/list API responses
- [x] Unit, API DB, Behat, and Playwright coverage added
- [ ] Capabilities updated (not applicable)
- [x] API documentation updated with `composer openapi`

## ✅ Checklist

- [x] Focused unit tests added for policy provider, applier, runtime services, and sign/request enforcement
- [x] Behat scenarios cover required mode, optional per-signer selection, and independence from `collect_metadata`
- [x] OpenAPI and TypeScript types regenerated

## 🤖 AI (if applicable)


lib/Service/RequestSignatureService.php
```diff
@@ -506,6 +510,21 @@ private function associateToSigners(array $data, FileEntity $file): array {
...
+						$signer['geolocationRequired'] ?? false,
+						FILTER_VALIDATE_BOOLEAN,
+						FILTER_NULL_ON_FAILURE,
+					) ?? false;
+					$this->signerGeolocationPolicyService->persistEffectiveRequirementToStorage(
+						$lastSignRequest,
+						$file,
+						$requesterRequiresGeolocation,
+						$requester,
+					);
```
vitormattos: It looks like we only persist the geolocation requirement to the last `SignRequest` created for the signer.

A signer can have multiple `identifyMethods`, and each iteration can create/update a `SignRequest`. Shouldn't we persist the effective requirement for every created sign request?

Otherwise, a sign request without `geolocationRequirement` will fall back to `disabled` during validation, which could bypass a required geolocation policy.

It would also be good to add a test with multiple identify methods for the same signer to validate if will work fine.

lib/Service/File/FileListService.php
```diff
@@ -463,6 +463,12 @@ private function formatSignerData(
 		if ($signer->getSigned()) {
 			$data['signed'] = $signer->getSigned()->format(DateTimeInterface::ATOM);
 		}
+
+		$signerMetadata = $signer->getMetadata();
+		if (is_array($signerMetadata) && $signerMetadata !== []) {
+			$data['metadata'] = $signerMetadata;
+		}
+
```
vitormattos, line 477: Could we avoid exposing the complete signer metadata here and only expose the geolocation metadata needed by this feature?

I think this should be based on the data stored in the `SignRequest`, not on the current policy. For example, a request may have required geolocation, the signer may already have signed and stored it, and the admin may disable the policy later. We should still return the historical data for that signer.

So, when present, I think we only need to expose `geolocationRequirement` and `geolocation` here. This also avoids unintentionally exposing unrelated current or future signer metadata with empty data through the file list API.

Could we also add a regression test covering this behavior to validate when this data is present and when not?

vitormattos (changes requested):

lib/Service/SignerGeolocation/SignerGeolocationPolicyService.php
```diff
@@ -0,0 +1,175 @@
...
+		$metadata = $signRequest->getMetadata() ?? [];
+		$stored = $metadata[self::METADATA_REQUIREMENT_KEY] ?? null;
+		if (!is_string($stored)) {
+			return null;
+		}
+
+		return SignerGeolocationMode::tryFrom($stored);
+	}
+
+	public function resolveEffectiveRequirement(
```
vitormattos, line 65: The effective per-signer behavior needs to be changed here.

The policy and the signer requirement have different meanings.

The policy defines if the requester can use geolocation:

* `disabled`: geolocation cannot be requested for any signer.
* `optional`: the requester can choose which individual signers must provide geolocation.
* `required`: all signers must provide geolocation. The requester cannot disable it for any signer.

The `optional` policy does not mean that LibreSign should try to collect geolocation from every signer. It only gives the requester the option to require geolocation for selected signers.

Because of this, the final requirement stored for each signer should be:

* policy `disabled` -> `disabled`
* policy `optional` + `geolocationRequired: false` -> `disabled`
* policy `optional` + `geolocationRequired: true` -> `required`
* policy `required` -> `required`

The stored per-signer requirement should therefore only be `disabled` or `required`. `optional` is a policy mode, not a final signer requirement.

Please also update the tests to cover all these cases and make sure the stored requirement does not change if the policy is changed after the signature request is created.

lib/Service/SignerGeolocation/SignerGeolocationPolicyService.php
```diff
@@ -0,0 +1,175 @@
...
+	): void {
+		$this->validateRequesterConfiguration($file, $requesterRequiresGeolocation, $requester);
+		$effective = $this->resolveEffectiveRequirement($file, $requesterRequiresGeolocation, $requester);
+
+		$metadata = $signRequest->getMetadata() ?? [];
+		$metadata[self::METADATA_REQUIREMENT_KEY] = $effective->value;
+		$signRequest->setMetadata($metadata);
+	}
+
+	public function persistEffectiveRequirementToStorage(
```
vitormattos: Do we need to load the `SignRequest` from the database again here?

`createOrUpdateSignRequest()` already persists and returns the `SignRequest` entity before this method is called.

We also have `persistEffectiveRequirement()` above with almost the same validation and resolution logic.

Could we keep only one method to apply the effective requirement and avoid loading the same entity again before updating it? This would make the responsibility of this service simpler and avoid an extra database query for every signer.

lib/Service/Policy/Provider/SignerGeolocation/SignerGeolocationPolicyValue.php
```diff
@@ -0,0 +1,66 @@
...
+		}
+
+		return [
+			'mode' => $mode->value,
+			'allowRequesterOverride' => filter_var(
+				$rawValue['allowRequesterOverride'] ?? false,
+				FILTER_VALIDATE_BOOLEAN,
+				FILTER_NULL_ON_FAILURE,
+			) ?? false,
+		];
```
vitormattos, line 36: Now that `optional` means the requester can choose which signers must provide geolocation, do we still need `allowRequesterOverride` in this policy?

The three modes already seem to define the full behavior:

* `disabled`: geolocation cannot be requested.
* `optional`: the requester can choose which signers must provide geolocation.
* `required`: all signers must provide geolocation.

With this rule, `optional` + `allowRequesterOverride: false` would have almost the same behavior as `disabled`.

I think it would be better to have only one source of truth for this behavior. Otherwise, we can have policy combinations that are valid in the API but do not have a useful meaning.

Could we simplify this policy and remove `allowRequesterOverride` if it is no longer needed?

tests/php/Unit/Service/SignerGeolocation/SignerGeolocationPolicyServiceTest.php
```diff
@@ -0,0 +1,241 @@
...
+	private function getService(): SignerGeolocationPolicyService {
+		return new SignerGeolocationPolicyService(
+			$this->policyService,
+			$this->fileMapper,
+			$this->signRequestMapper,
+			$this->l10n,
+		);
+	}
+
+	public function testResolveEffectiveRequirementUsesRequesterOverride(): void {
```
vitormattos: Could we use a data provider for the effective requirement scenarios?

This logic is a decision matrix, and after the behavior change we should cover at least:

* `disabled` + requester not requiring geolocation
* `disabled` + requester requiring geolocation
* `optional` + requester not requiring geolocation
* `optional` + requester requiring geolocation
* `required` + requester not requiring geolocation
* `required` + requester requiring geolocation

A data provider would make the expected behavior easier to understand, reduce repeated policy snapshot setup, and make it easier to confirm that all branches are covered by mutation testing.

tests/php/Unit/Service/SignerGeolocation/SignerGeolocationMetadataValidatorTest.php
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	#[DataProvider('invalidGeolocationPayloadProvider')]
+	public function testNormalizeRejectsInvalidPayload(mixed $payload): void {
+		$this->expectException(LibresignException::class);
+		$this->getValidator()->normalize($payload);
+	}
+
+	/** @return iterable<string, array{0: mixed}> */
+	public static function invalidGeolocationPayloadProvider(): iterable {
```
vitormattos, line 124: Since this validator is isolated logic and the issue requires mutation testing, could we extend this data provider to cover the validation boundaries?

For example, longitude limits, negative latitude limit, invalid numeric values, negative accuracy, and invalid/negative timestamps.

These comparisons are likely mutation points, so covering the boundaries here should help us reach the Infection target without adding duplicated test methods.

vitormattos (changes requested):

lib/Service/File/FileListService.php
```diff
@@ -463,10 +469,66 @@ private function formatSignerData(
...
+		$signerMetadata = $signer->getMetadata();
+		if (!is_array($signerMetadata) || $signerMetadata === []) {
+			return [];
+		}
+
+		$geolocationMetadata = [];
+
+		$storedRequirement = $signerMetadata[SignerGeolocationPolicyService::METADATA_REQUIREMENT_KEY] ?? null;
+		$requirement = is_string($storedRequirement) ? SignerGeolocationMode::tryFrom($storedRequirement) : null;
+		if ($requirement === SignerGeolocationMode::OPTIONAL) {
```
vitormattos: Do we need to handle `optional` as a legacy signer requirement here?

This feature has not been released yet, so we should not have production data with `geolocationRequirement = optional`.

Since the signer requirement now only supports `disabled` or `required`, I think it would be better to keep this strict instead of converting an invalid value to `disabled`.

`optional` should only be a policy mode, not a signer requirement. This change should also update or remove the test that treats `optional` as a legacy value.

vitormattos (changes requested):

vitormattos (approved):
The implementation looks good now.

The PR description still mentions `allowRequesterOverride` in the summary, test examples, and checklist. Since this option was removed, could you update the description to match the final behavior only to maintain the history?

I'll approve and merge this PR now.

Issue #8117, opened 2026-08-28
Stop asserting exception messages in PHP tests: define a convention (error codes or typed exceptions)


maia-andre:
Follow-up of the discussion in #8092 (https://github.com/LibreSign/libresign/pull/8092#discussion_r3883064993).

Many PHP tests assert exceptions by matching their message. When the message is translatable or just easy to change, editing the text in `lib/` silently requires editing the test too, and whoever changes the text (not necessarily a developer) has to know that. The suggestion in #8092 was to assert the error **code** or the exception **class** instead of the text.

## Current state

Numbers from `main` today:

- `expectExceptionMessage()`: 174 calls in 52 test files (plus 13 `expectExceptionMessageMatches()`).
- `expectExceptionCode()`: 12 calls in 9 files.
- `throw new LibresignException(...)`: 287 places. 161 of them use a translated message (`$l10n->t(...)`), 126 use a literal string.
- Only 51 of the 287 carry a code, and the values mix conventions: `1` (22), `422` (11), `404` (7), `500` (4), `400` (3), `Http::STATUS_UNPROCESSABLE_ENTITY` (3) and `Http::STATUS_FORBIDDEN` (1). `LibresignException` itself defines no code constants, and the code is exposed to clients through `ErrorPayloadBuilder` (`code`) and `SignFileService` (`error_code`), so any convention also touches the API contract.
- 8 test files assert a translated message while building the real `IL10N` from the factory, so they depend on the test locale.
- There are already typed subclasses in `lib/Exception/` (`InvalidPasswordException`, `InvalidSignatureException`, `EmptyCertificateException`, ...), but most `throw` sites still use the base `LibresignException`.

## What should be decided first

Two directions came up in #8092. They are not exclusive, but a convention is needed before touching many files:

1. **Error codes**: define constants for the codes (a dedicated class, like `JSActions`, following the "HTTP-like plus one digit" idea to avoid confusion with HTTP status codes), use them at the `throw` sites, and assert them with `expectExceptionCode()`.
2. **Typed exceptions**: extend the existing `lib/Exception/` subclasses so tests assert the class with `expectException()`.

Please use this issue to settle the convention (and decide what to do with the existing `1`/`4xx` codes), then the work can be split into small PRs.

## How to contribute

> [!IMPORTANT]
> Do not solve this issue in one PR.

[...]

PR #8126, opened 2026-08-29
test: cover CertificateChainService branches and the saveFile failure message

maia-andre:
### Pull Request Description

Tests-only PR for #8053 covering two related certificate-handling helpers.

**`lib/Service/File/CertificateChainService.php`**

Infection scoped to this file, running only `CertificateChainServiceTest` (`--threads=4`, 60 s timeout):

| | mutants | killed | escaped | Covered Code MSI |
|---|---|---|---|---|
| before | 22 | 17 | 5 | 77% |
| after | 27 | 25 | 2 | 92% |

New tests:

- the two early returns (`validateFile` disabled / file without signed node) through a data provider — the file must not even be opened, which kills the `||` → `&&` mutant;
- hash mismatch: the chain is still returned and the policy user is still set, but `setIsLibreSignFile()` must not be called;
- handler failure: a warning with the underlying message is logged and an empty array is returned;
- the stream is rewound before the handler reads it and closed afterwards. `PolicyAwarePkcs12HandlerDouble` now records what it read and returns two signature entries, as the real handler does, so the `ArrayOneItem` mutant on the return is detected. The existing success test asserts the full array instead of a `chain` key that the real handler does not return at the top level.

The two remaining escaped mutants are `fread($resource, 8192 ± 1)`: any buffer size produces the same hash, so they are equivalent.

**`lib/Handler/CertificateEngine/CertificateHelper.php`**

Infection scoped to this file, running only `CertificateHelperTest` (`--threads=1`, 60 s timeout):

| | mutants | killed | escaped | Covered Code MSI |
|---|---|---|---|---|
| before | 30 | 27 | 3 | 90% |
| after | 30 | 30 | 0 | 100% |

The three escaped mutants were all on the `saveFile()` failure message (concatenation order and operand removal). `testSaveFileWithError` now asserts the message **shape** — a description, a colon, then the file that could not be written — rather than its wording, following the discussion in #8117.

Notes on the measurements: the runs use the class's own tests so they are reproducible in a couple of minutes; with the whole unit suite and Infection's default 10 s timeout the `CertificateHelper` numbers are polluted by false timeouts (the test process boots Nextcloud) and, with several threads, by test classes extending `OCA\Libresign\Tests\Unit\TestCase` racing on the shared `/tmp/libresign__original` backup. Nothing in this PR changes that; I can open a separate note if it is useful.

### Related Issue

Ref #8053

### Pull Request Type

- Test

### Pull request checklist

- [x] Tests pass locally (PHPUnit + Infection)
- [x] AI-assisted: yes (see commit trailer)

### DIFF
--- a/tests/php/Unit/Handler/CertificateEngine/CertificateHelperTest.php
+++ b/tests/php/Unit/Handler/CertificateEngine/CertificateHelperTest.php
@@ -26,13 +26,28 @@ final class CertificateHelperTest extends \OCA\Libresign\Tests\Unit\TestCase {
 		$this->assertSame($content, file_get_contents($filename));
 	}
 
-	public function testSaveFileWithError(): void {
-		$this->expectException(LibresignException::class);
-
+	#[DataProvider('dataSaveFileWithError')]
+	public function testSaveFileWithError(string $basename): void {
 		vfsStream::setup('home', 0444);
-		$filename = vfsStream::url('home/test.txt');
+		$filename = vfsStream::url('home/' . $basename);
 
-		@CertificateHelper::saveFile($filename, '');
+		try {
+			@CertificateHelper::saveFile($filename, '');
+			$this->fail('saveFile() must throw when the file cannot be written');
+		} catch (LibresignException $e) {
+			// The failure must describe the problem and end with the file that could not be written
+			$this->assertStringEndsWith(': ' . $filename, $e->getMessage());
+		}
+	}
+
+	public static function dataSaveFileWithError(): array {
+		return [
+			'plain name' => ['test.txt'],
+			'spaces' => ['my certificate.pem'],
+			'dots, dashes and underscores' => ['root-ca.v2_backup.crt'],
+			'regex metacharacters' => ['cert (copy) [1]+$.txt'],
+			'non-ascii' => ['certificado-ção.txt'],
+		];
 	}
 
 	#[DataProvider('dataProviderArrayToIni')]
--- a/tests/php/Unit/Service/File/CertificateChainServiceTest.php
+++ b/tests/php/Unit/Service/File/CertificateChainServiceTest.php
@@ -12,6 +12,7 @@ use OCA\Libresign\Db\File;
 use OCA\Libresign\Handler\SignEngine\Pkcs12Handler;
 use OCA\Libresign\Service\File\CertificateChainService;
 use OCA\Libresign\Service\File\FileResponseOptions;
+use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\TestCase;
 use Psr\Log\LoggerInterface;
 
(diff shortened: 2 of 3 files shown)


tests/php/Unit/Handler/CertificateEngine/CertificateHelperTest.php
```diff
@@ -27,11 +27,13 @@ public function testSaveFileWithSuccess(): void {
...
 
 	public function testSaveFileWithError(): void {
-		$this->expectException(LibresignException::class);
-
 		vfsStream::setup('home', 0444);
 		$filename = vfsStream::url('home/test.txt');
 
+		$this->expectException(LibresignException::class);
+		// The failure must describe the problem and end with the file that could not be written
+		$this->expectExceptionMessageMatches('/^.+: ' . preg_quote($filename, '/') . '$/');
```
vitormattos: Maybe we can use `assertStringContainsString` and use dataprovider or use more different characters into filename.

tests/php/Unit/Service/File/CertificateChainServiceTest.php
```diff
@@ -83,4 +83,111 @@ public function fopen($mode) {
...
+
+		$this->assertSame([], $service->getCertificateChain($fileNode, $libreSignFile, $options));
+	}
+
+	public static function dataGetCertificateChainSkipsWhenValidationIsNotRequested(): array {
+		return [
+			'validation disabled' => [false, 1],
+			'file without signed node' => [true, null],
+		];
+	}
```
vitormattos, line 116: I think the test and data provider names could describe these cases better.

The second dataset has validation enabled, but there is no signed node, so `SkipsWhenValidationIsNotRequested` does not fully represent it. Maybe rename them to something like `testGetCertificateChainSkipsValidation` and `dataGetCertificateChainSkipsValidation`.

The dataset names themselves can stay simple, for example `validation disabled` and `missing signed node`.

vitormattos (approved):
Approving this PR with only a small improvement suggestion that don't block to be merged, only to see your opinion about.

tests/php/Unit/Handler/CertificateEngine/CertificateHelperTest.php
maia-andre: Done — replaced the regex with `expectExceptionMessage($filename)` (PHPUnit's substring check on the message, the same as `assertStringContainsString`, so it composes with `expectException()`) and turned the test into a data provider with five file name shapes: plain, spaces, dots/dashes/underscores, regex metacharacters and non-ASCII. Infection on `CertificateHelper.php` stays at 30 mutants / 0 escaped.

tests/php/Unit/Service/File/CertificateChainServiceTest.php
maia-andre: Agreed — the old name only described the first dataset. Renamed to `testGetCertificateChainSkipsValidation` / `dataGetCertificateChainSkipsValidation`, with the datasets `validation disabled` and `missing signed node`.

tests/php/Unit/Handler/CertificateEngine/CertificateHelperTest.php
maia-andre: Correction on the numbers I posted above: that run used 4 threads, and for classes extending the LibreSign `TestCase` parallel runs produce errors that Infection counts as kills. Re-measured with a single thread, the `contains` version lets 2 mutants escape on `CertificateHelper.php:17` — the operands swapped, and the description dropped so the message is just the file name.

I changed the assertion to `assertStringEndsWith(': ' . $filename, $e->getMessage())` — no regex, still no wording — which kills both: 30 mutants / 0 escaped with 1 thread. The data provider with the five file name shapes stays.

PR #8142, opened 2026-08-30
test: improve unit tests coverage for AccountService (Infection)

diogopaz:
Part of #8053

## 📝 Summary

This PR adds extensive PHPUnit test coverage to `AccountServiceTest` to eliminate escaped mutants and improve the mutation testing score with Infection, as part of #8053.

## 🧪 How to test

Run the focused unit test suite for AccountService:

```bash
composer test:unit -- --filter AccountServiceTest
```

Optionally run Infection on this specific pair to verify mutant coverage:
```bash
vendor-bin/infection/vendor/bin/infection --filter=AccountService
```

## ⚙️ API / Back‑end changes

- [x] Unit tests added to cover `validateCreateToSign`, `getFileByUuid`, `createToSign`, `saveVisibleElement(s)`, `uploadPfx`, `updatePfxPassword`, `readPfxData`, and config fallbacks.

## ✅ Checklist

## 🤖 AI (if applicable)

### DIFF
--- a/tests/php/Unit/Service/AccountServiceTest.php
+++ b/tests/php/Unit/Service/AccountServiceTest.php
@@ -8,18 +8,26 @@ declare(strict_types=1);
 
 namespace OCA\Libresign\Tests\Unit\Service;
 
+use DateTime;
+use Exception;
+use InvalidArgumentException;
 use OC\AppFramework\Utility\TimeFactory;
 use OC\Http\Client\ClientService;
 use OCA\Libresign\AppInfo\Application;
+use OCA\Libresign\Db\File as DbFile;
 use OCA\Libresign\Db\FileMapper;
 use OCA\Libresign\Db\FileTypeMapper;
+use OCA\Libresign\Db\IdentifyMethod;
 use OCA\Libresign\Db\IdentifyMethodMapper;
 use OCA\Libresign\Db\SignRequest;
 use OCA\Libresign\Db\SignRequestMapper;
 use OCA\Libresign\Db\UserElement;
 use OCA\Libresign\Db\UserElementMapper;
 use OCA\Libresign\Enum\CRLReason;
+use OCA\Libresign\Exception\InvalidPasswordException;
+use OCA\Libresign\Exception\LibresignException;
 use OCA\Libresign\Handler\CertificateEngine\CertificateEngineFactory;
+use OCA\Libresign\Handler\CertificateEngine\ICertificateEngine;
 use OCA\Libresign\Handler\SignEngine\Pkcs12Handler;
 use OCA\Libresign\Helper\FileUploadHelper;
 use OCA\Libresign\Helper\ValidateHelper;
@@ -50,6 +58,8 @@ use OCP\Files\IMimeTypeDetector;
 use OCP\Files\IRootFolder;
 use OCP\Files\NotFoundException;
 use OCP\Group\ISubAdmin;
+use OCP\Http\Client\IClient;
+use OCP\Http\Client\IResponse;
 use OCP\IAppConfig;
 use OCP\IGroup;
 use OCP\IGroupManager;
@@ -57,6 +67,7 @@ use OCP\IL10N;
 use OCP\IURLGenerator;
 use OCP\IUser;
 use OCP\IUserManager;
+use OCP\Mail\IEMailTemplate;
 use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\MockObject\MockObject;
 
@@ -306,6 +317,16 @@ final class AccountServiceTest extends \OCA\Libresign\Tests\Unit\TestCase {
 					],
 				],
 				'Invalid email'
+			],
+			'emptySignPassword' => [
+				[
+					'uuid' => '12345678-1234-1234-1234-123456789012',
+					'user' => [
+						'email' => 'valid@test.coop',
+					],
+					'signPassword' => '',
+				],
+				'Password to sign is mandatory'
 			]
 		];
 	}
@@ -537,6 +558,35 @@ final class AccountServiceTest extends \OCA\Libresign\Tests\Unit\TestCase {
 
 	public static function provideValidateCreateToSignCases():array {
 		return [
+			'missingIdentify' => [
+				function ($self): array {
+					$signRequest = $self->createMock(SignRequest::class);
+					$signRequest->method('getId')->willReturn(10);
+					$self->signRequestMapper->method('getByUuid')->willReturn($signRequest);
+					return [
+						'uuid' => '12345678-1234-1234-1234-123456789012',
+						'user' => [],
+					];
+				},
+				'Invalid identification method'
+			],
+			'invalidIdentifyMethod' => [
+				function ($self): array {
+					$signRequest = $self->createMock(SignRequest::class);
+					$signRequest->method('getId')->willReturn(10);
+					$self->signRequestMapper->method('getByUuid')->willReturn($signRequest);
+					$self->identifyMethodService->method('getIdentifyMethodsFromSignRequestId')->willReturn(['email' => []]);
+					return [
+						'uuid' => '12345678-1234-1234-1234-123456789012',
+						'user' => [
+							'identify' => [
+								'phone' => '123456',
+							],
+						],
+					];
+				},
+				'Invalid identification method'
+			],
 			'invalidUuid' => [
 				[
 					'uuid' => 'invalid uuid'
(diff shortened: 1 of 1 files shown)


vitormattos:
@diogopaz  the problems in the unit tests aren't caused by your PR. I'll fix this in a separate PR.

But have linter issues that you need to solve:

https://github.com/LibreSign/libresign/actions/runs/33325571740/job/100018743935?pr=8142

vitormattos:
Once you fix the linter issue and push the changes, the unit tests will work fine.

diogopaz:
Hi @vitormattos,

Sorry for the delay! I've fixed the linter issues (missing imports, multiline array trailing commas, and formatting) and updated the commit.

Thanks for the review!

tests/php/Unit/Service/AccountServiceTest.php
```diff
@@ -1047,4 +1097,868 @@ public function testGetConfigIncludesCanManageGroupPoliciesForInstanceAdmin(): v
...
+			->method('getUserFolder')
+			->with('testuser')
+			->willReturn($userFolder);
+
+		$result = $this->getService()->getFileByUuid('test-uuid');
+		$this->assertSame($file, $result['fileData']);
+		$this->assertSame($fileToSign, $result['fileToSign']);
+	}
+
+	public function testGetFileByUuidUsesCache(): void {
```
vitormattos, line 1134: I think we should test one more case before making this cache behavior part of the expected behavior.

`AccountService` caches `signRequest`, `fileData` and `fileToSign`, but the cache does not store which UUID the data belongs to.

This test calls `getFileByUuid()` twice with the same UUID, so it does not detect what happens with different UUIDs.

Could we also test something like:

```php
$service->getFileByUuid('uuid-a');
$result = $service->getFileByUuid('uuid-b');
```

and verify that the result belongs to `uuid-b`?

If this fails with the current implementation, we found a bug and should fix it instead of protecting this behavior with the mutation test. This is also tracked in #8328.

tests/php/Unit/Service/AccountServiceTest.php
```diff
@@ -1047,4 +1097,868 @@ public function testGetConfigIncludesCanManageGroupPoliciesForInstanceAdmin(): v
...
+		$this->userElementMapper->expects($this->exactly(2))
+			->method('findOne')
+			->willReturnCallback(fn (array $criteria) => match ($criteria['id']) {
+				456 => $userElement1,
+				789 => $userElement2,
+			});
+
+		$file1 = $this->createMock(File::class);
+		$file2 = $this->createMock(File::class);
+		$this->folderService->expects($this->exactly(2))
```
vitormattos, line 1503: This expectation does not match the current behavior.

These elements only contain `elementId` and `starred`. They do not contain `file`, so `updateFileOfVisibleElement()` returns before calling `folderService->getFileByNodeId()`.

This is also what the failing PHPUnit job shows: the method was expected 2 times but was called 0 times.

Could you remove this filesystem setup and keep this test focused on the `starred` update behavior?

tests/php/Unit/Service/AccountServiceTest.php
```diff
@@ -1047,4 +1097,868 @@ public function testGetConfigIncludesCanManageGroupPoliciesForInstanceAdmin(): v
...
+		$this->userElementMapper->expects($this->exactly(2))
+			->method('update')
+			->willReturnCallback(function (UserElement $element) {
+				return $element;
+			});
+
+		$this->getService()->saveVisibleElements($elements, 'session123', $user);
+	}
+
+	public function testSaveVisibleElementUpdateFileAndStarred(): void {
```
vitormattos, line 1519: There are a few related failures in the new visible-element tests that should be fixed before merge.

In this test, `saveVisibleElement()` calls both `updateFileOfVisibleElement()` and `updateDataOfVisibleElement()`. Each method calls `userElementMapper->findOne()`, so expecting `findOne()` only once makes the test fail.

There are two similar problems below:

* `getStarred()` returns a boolean here, so `assertSame(1, ...)` fails with `true`. Please assert the actual type/value returned by the entity.
* The invalid URL tests call `saveVisibleElement()` without preparing the folder needed by the session path. The code fails with `newFile() on null` before it reaches the URL validation that the test wants to check.

Please adjust the setup so each test reaches the behavior it is meant to test. The invalid URL / MIME / empty body cases may also fit well in one data provider because they test the same rule with different inputs.

tests/php/Unit/Service/AccountServiceTest.php
```diff
@@ -1047,4 +1097,868 @@ public function testGetConfigIncludesCanManageGroupPoliciesForInstanceAdmin(): v
...
+		// Call first time (populates cache)
+		$service->getFileByUuid('test-uuid');
+		// Call second time (uses cache)
+		$result = $service->getFileByUuid('test-uuid');
+
+		$this->assertSame($file, $result['fileData']);
+		$this->assertSame($fileToSign, $result['fileToSign']);
+	}
+
+	public function testCreateToSignWithSuccessAndEmailAndSignPassword(): void {
```
vitormattos, line 1172: This test also shows why #8328 is useful: testing one account operation requires setup for account creation, identification, email and certificate handling.

I would keep the architecture refactor outside this PR, but we should avoid adding more test complexity only to improve the Infection score.

vitormattos (changes requested):
I checked the PHPUnit failures.

There are two different groups of problems:

1. The `IUserFolder` errors in `ValidateHelperTest` and `SignFileServiceTest` are not caused by this PR. They come from tests outside the changed file and appear in the Nextcloud master jobs. We should handle this compatibility change separately.

2. The failures in `AccountServiceTest` are from the tests added in this PR and need to be fixed here before merge. I added inline comments for the main cases.

So you do not need to fix the unrelated `IUserFolder` failures as part of this PR.

vitormattos:
@diogopaz, can you rebase your branch on top of `main`?

I fixed part of the failing tests in this PR:

* https://github.com/LibreSign/libresign/pull/8329

maia-andre:
Heads-up after #8366 was merged (this morning): `AccountService` no longer receives `ValidateHelper`. `AccountServiceTest` on `main` now mocks `IdentityDocumentValidator` (`$this->identityDocumentValidator`) and `FileInputValidator` (`$this->fileInputValidator`) and passes them to the constructor; the `$this->validateHelper` property and the `use OCA\Libresign\Helper\ValidateHelper;` import are gone.

After rebasing, five expectations in this PR will reference a property that no longer exists. The one-to-one mapping against current `main`:

| This PR | On `main` |
|---|---|
| `$this->validateHelper->method('userCanApproveValidationDocuments')->with(null, false)` | `$this->identityDocumentValidator->method('userCanApproveValidationDocuments')->with(null, false)` |
| `$this->validateHelper->method('userCanApproveValidationDocuments')->with($user, false)` (×2) | `$this->identityDocumentValidator->method(...)->with($user, false)` |
| `$this->validateHelper->expects($this->once())->method('validateBase64')->with('fake_png_data', ValidateHelper::TYPE_VISIBLE_ELEMENT_USER)` | `$this->fileInputValidator->expects($this->once())->method('validateBase64')->with('fake_png_data', FileInputValidator::TYPE_VISIBLE_ELEMENT_USER)` |

The production calls are `lib/Service/AccountService.php:218` and `:354` (`identityDocumentValidator->userCanApproveValidationDocuments($user, false)`) and `:522`/`:525` (`fileInputValidator->validateBase64(..., FileInputValidator::TYPE_VISIBLE_ELEMENT_USER)`). The imports for both validators are already in the test file on `main`, so the `use ValidateHelper` line can simply be dropped. Everything else in the PR is additive and should not conflict beyond the `use` block.

Sorry for the extra step — happy to review the rebase if useful.

tests/php/Unit/Service/AccountServiceTest.php
```diff
@@ -554,6 +575,35 @@ public function testValidateCreateToSignUsingDataProvider($arguments, $expectedE
 
 	public static function provideValidateCreateToSignCases():array {
 		return [
+			'missingIdentify' => [
+				function ($self): array {
+					$signRequest = $self->createMock(SignRequest::class);
+					$signRequest->method('getId')->willReturn(10);
```
vitormattos, line 581: `SignRequest` uses dynamic getters from `Entity`, so PHPUnit cannot configure `getId()`, `getFileId()` or `getDisplayName()` on this mock. This is causing several errors in the PHPUnit job.

Could you use a real `SignRequest` here and set the needed values with `setId()`, `setFileId()`, `setDisplayName()`, etc.? The same fix is needed in the other new tests that mock these getters.

You also can look other examples with getId at other tests.

tests/php/Unit/Service/AccountServiceTest.php
```diff
@@ -1052,4 +1102,807 @@ public function testGetConfigIncludesCanManageGroupPoliciesForInstanceAdmin(): v
...
+
+		$identifyMethod->expects($this->never())->method('getEntity');
+		$this->identifyMethodMapper->expects($this->never())->method('update');
+		$this->appConfig->method('getValueString')->willReturn('no');
+
+		$this->getService()->createToSign('uuid-123', 'john@example.com', 'pass123', null);
+	}
+
+	public function testGetCertificateEngineName(): void {
+		$engine = $this->createMock(ICertificateEngine::class);
```
vitormattos, line 1277: `ICertificateEngine` does not exist in the current code. `CertificateEngineFactory::getEngine()` returns `IEngineHandler`.

Could you use `IEngineHandler` in these two tests instead?

tests/php/Unit/Service/AccountServiceTest.php
```diff
@@ -1052,4 +1102,807 @@ public function testGetConfigIncludesCanManageGroupPoliciesForInstanceAdmin(): v
...
+		$engine->method('isSetupOk')->willReturn(true);
+		$this->certificateEngineFactory->method('getEngine')->willReturn($engine);
+
+		$this->assertTrue($this->getService()->isSetupOk());
+	}
+
+	public function testGetConfigWithNullUser(): void {
+		$this->idDocsPolicyService->method('isIdentificationDocumentsEnabled')->with(null)->willReturn(false);
+		$this->identityDocumentValidator->method('userCanApproveValidationDocuments')->with(null, false)->willReturn(false);
+		$this->policyAuthorizationService->method('canUserManageGroupPolicies')->with(null)->willReturn(false);
```
vitormattos, line 1295: `$this->policyAuthorizationService` is a real `PolicyAuthorizationService`, not a PHPUnit mock, so `->method()` cannot be used here.

For a null user, the real service already returns `false` from `canUserManageGroupPolicies()` and `[]` from `getManageablePolicyGroupIds()`. I think these two stubs can just be removed.

vitormattos (changes requested):

tests/php/Unit/Service/AccountServiceTest.php
vitormattos: The previous comment became outdated after the rebase, but the point is still valid.

This test only checks one UUID. Could you also use the same `AccountService` instance with two different UUIDs and verify that the second call returns the data for the second UUID?

For example:

```php
$service = $this->getService();
$service->getFileByUuid('uuid-a');
$result = $service->getFileByUuid('uuid-b');
```

This is important because `AccountService` caches the sign request and file data. We should make sure the test does not protect a cache behavior that returns data from the previous UUID.

PR #8143, opened 2026-08-31
feat: add observer participant role for signature requests

### LINKED ISSUE #6957: Add participant role permissions: Sign and Observe
Currently, LibreSign does not provide granular permission control for document participants. We need to implement a permission/role system that allows defining specific rules for each participant.

## Proposed Feature

Add the ability to assign different permission levels to document participants with two distinct roles:

1. **Sign** - Participants with this permission can digitally sign the document
2. **Observe** - Participants with this permission can only view the document and track its progress without signing

## Configuration Requirement

A new setting needs to be added to enable this feature. This could be implemented as an "Observer Profile" setting that, when enabled, allows document owners to assign the observer role to participants.

## Use Cases

- **Audit trails**: Observers can monitor the signing process without being active signers
- **Stakeholder visibility**: Team members who need to track document status but don't need to sign
- **Legal/Compliance**: Compliance officers or legal teams who need to review documents without participating in the signature process
- **CC functionality**: Similar to email CC, where people are kept informed but don't need to take action

## Expected Behavior

- Add an admin/user setting to enable the observer profile functionality
- When adding participants to a document, the document owner should be able to select the role for each participant (Sign or Observe)
- Observers should have read-only access to the document and see the signing progress
- Only participants with "Sign" permission can sign the document
- The system should clearly distinguish between signers and observers in the UI
- Audit logs should track which participant had which role

## Additional Context

This feature would provide more flexibility for document workflows where some stakeholders need visibility without signing authority, similar to being CC'd on an email.

lfals:
Resolves: #6957

## 📝 Summary

Adds an **observer** participant role (UI label: **Observer** / **Testemunha**) to signature requests. Observers can follow document progress and open a **read-only** view of the request (including PDF and signature positions), but **cannot sign** or perform write/signing operations. The feature is gated by the policy `enable_observer_profile` (default: disabled).

Backend is the source of truth: role validation, signing denial, status/progress calculations, DocMDP signer limits, sequential signing, e-mail/notification routing, UUID authorization, and API contracts all treat observers separately from signers. The request UI exposes separate **Signers** and **Observers** lists when the policy is enabled.

Authorization is split explicitly:
- `RequireSignRequestUuid` — signer-only (signing UI, id-docs, signature elements, and other write paths reject observers)
- `RequireParticipantUuid` — participant read-only (PDF/`getPdfFile` allows both signers and observers)

Also includes: policy snapshots for existing requests/envelopes, preserve observers when reordering signers, require at least one signer before requesting signatures, hide signing-order numbers for observers, OBSERVING status `4` in API summaries, observer mail copy (“ready to view”), Account observer deep-links to the authenticated read-only filelist, and focused regression coverage (PHPUnit, Vitest, Behat, Playwright).

## 🧪 How to test

1. Enable the policy **Enable observer profile** (system, group, or user scope as needed).
2. Open the Files app, select a PDF, and open **Request signature**.
3. Use **Add → Signer** and **Add → Observer** to add participants (Add menu appears only when the policy is enabled; otherwise keep the classic **Add signer** button).
4. Confirm the UI shows two sections (**Signers** / **Observers**).
5. Try requesting signatures with **only observers** → must be blocked (toast + API error).
6. Add at least one signer and request signatures → request should succeed; observers must **not** appear as pending signers or receive signing-order numbers.
7. Confirm observers are hidden from the signature-positions placement targets; “View signature positions” stays available for read-only review when there is at least one signer.
8. As an **Account** observer, open the notification/link → authenticated read-only request sidebar (`filelist/sign?uuid=…`), not the public validation/sign flow.
9. As an **Email** observer, open the mail link → public validation / view document; subject/body use “ready to view”.
10. In **ordered numeric** flow, reorder signers with observers present → observers remain in the list without order counters.
11. Open validation for a document with both roles → signers and observers appear in separate groups; observer status shows **Observing** when applicable.
12. **Authorization boundary:** with an anonymous request using an **observer** UUID header:
    - `GET /apps/libresign/pdf/<OBSERVER_UUID>` → **200**
    - `POST /api/v1/id-docs` and `POST /api/v1/signature/elements` → **422** (“Observers cannot sign this document”)
13. Confirm visible signature elements cannot be assigned to observers (UI + API).

## 🎨 UI / Front‑end changes

- [x] Policy-gated **Add → Signer / Observer** menu (classic **Add signer** when policy disabled)
- [x] Separate **Signers** and **Observers** lists in request and validation views
- [x] Observer-specific labels in participant modal (search, name validation, custom message)
- [x] Client-side guard: block signature request when only observers are present
- [x] Drag-and-drop reorder in ordered flow preserves observers; no signing-order enumerator for observers
- [x] Hide observers from signature-positions placement; allow read-only open of positions when signers exist
- [x] Authenticated observers open a read-only request sidebar; Files list `canSign` ignores observers
- [x] Public validation CSS on UUID pages; anonymous View document opens the public PDF directly
- [x] “Send notification” action for observers (uses validation/update mail templates)
image
Captura de Tela 2026-09-04 às 16 12 59
image

🏚️ Before | 🏡 After
--- | ---
Single signer list only | Signers + Observers sections when policy enabled

- [ ] Tested in multiple browsers (Chrome, Firefox, Safari) – *optional but appreciated*
- [x] Components, Unit (with vitest) and/or e2e (with Playwright) tests added - *Required*
- [ ] Accessibility verified (contrast, keyboard navigation, screen reader friendly) – *if applicable*
- [x] Design review approved – *optional, link to feedback if available*
- [ ] Documentation updated (if applicable) – [docs repository](https://github.com/LibreSign/documentation/)

### 🚧 Tasks

- [ ] Sync translations via Transifex before release (strings live in source until the bot updates `l10n/`)
- [ ] Add/update user-facing documentation for observer role in [LibreSign documentation](https://github.com/LibreSign/documentation/)

## ⚙️ API / Back‑end changes

- [x] `participant_role` column on sign requests (`signer` | `observer`) with migration
- [x] Policy provider `enable_observer_profile` (default `false`, system/group/user scopes) with per-request/envelope **policy snapshot**
- [x] Validation: reject observers when policy disabled; reject signing and visible-element assignment for observers
- [x] Exclude observers from signing status, envelope progress, sequential signing, and DocMDP signer counts
- [x] Observer notifications: Email → validation / “ready to view”; Account → authenticated read-only filelist deep-link
- [x] Require at least one signing participant when moving out of draft / requesting signatures
- [x] Map `OBSERVING` as status **`4`** in signer summaries; reject invalid `ParticipantRole` values; freeze role after signed
- [x] UUID auth split: `RequireParticipantUuid` (read-only PDF) vs `RequireSignRequestUuid` (signer-only writes)
- [ ] Capabilities updated (if applicable) – if adding/modifying Nextcloud capabilities
- [x] API documentation updated with the command `composer openapi` if necessary

### 🚧 Tasks

- [ ] Confirm no environment still depends on experimental `participant_profile_id` (column was never shipped in official migrations)

## ✅ Checklist

- [x] OpenAPI specs and TypeScript types regenerated (`composer openapi`, `npm run typescript:generate`)
- [x] Focused PHPUnit, Vitest, Behat, and Playwright coverage for observer flows
- [x] Conventional Commits with DCO sign-off on all commits

## 🤖 AI (if applicable)


l10n/en_GB.js
vitormattos, line 1: We can't make any change nothing inside the l10n folder. This folder is updated only by the bot that syncs the translations from Transifex.

vitormattos (changes requested):

lib/Migration/Version19001Date20260831000000.php
vitormattos, line 1: I think this migration should be removed from this PR and use only the file `lib/Migration/Version19000Date20260830200000.php` to this PR.

participant_profile_id is not part of the LibreSign migration history, and this migration is cleaning up a column created by an experimental/local migration.

An official LibreSign migration should not drop an unknown column only to clean up development environments. If some environments need this cleanup, it should be handled separately.

This also avoids the risk of deleting data from an installation where a column with this name exists for another reason.

src/components/RightSidebar/RequestSignatureTab.vue
```diff
@@ -23,14 +23,39 @@
...
-		</NcButton>
+		<template v-if="filesStore.canAddSigner() && !isOriginalFileDeleted">
+			<NcButton v-if="!observerProfileEnabled"
+				:variant="hasSigners ? 'secondary' : 'primary'"
+				@click="addParticipant(PARTICIPANT_ROLE.SIGNER)">
+				<template #icon>
+					<NcIconSvgWrapper :path="mdiPlus" :size="20" />
+				</template>
+				{{ t('libresign', 'Add') }}
+			</NcButton>
```
vitormattos, line 34: When `enable_observer_profile` is disabled, please keep the existing `Add signer` button and icon.

The new `Add` menu is needed only when the observer feature is enabled. Changing the disabled-policy UI also forces many unrelated Playwright tests to change.

Keeping the old behavior when the feature is disabled will reduce this PR scope (files changed) and avoid regressions outside the observer flow.

lib/Enum/ParticipantRole.php
```diff
@@ -0,0 +1,38 @@
...
+		};
+	}
+
+	public static function fromNullable(?string $value): self {
+		if ($value === null || $value === '') {
+			return self::SIGNER;
+		}
+
+		return self::tryFrom($value) ?? self::SIGNER;
+	}
```
vitormattos, line 40: We should not fallback to `SIGNER` for an invalid participant role.

This is a new field, the database column is `NOT NULL`, and existing rows are migrated with `signer` as the default. So persisted data should always contain a valid role.

Since `SIGNER` grants signing permission, invalid data should fail instead of silently becoming a signer.

Please use strict parsing here and update the test that currently expects `invalid-role` to become `SIGNER`.

lib/Service/SignRequest/SignRequestService.php
```diff
@@ -65,21 +68,23 @@ public function createOrUpdateSignRequest(
 		);
 
 		$displayName = $this->getDisplayNameFromIdentifyMethodIfEmpty($identifyMethodsInstances, $displayName);
-		$this->populateSignRequest($signRequest, $displayName, $signingOrder, $description, $fileId);
+		$this->populateSignRequest($signRequest, $displayName, $signingOrder, $description, $fileId, $participantRole);
```
vitormattos, line 74: Please validate the role before `populateSignRequest()` changes the entity.

If an existing request is already `SIGNED`, reject the update when the received `participantRole` is different from the stored role. Otherwise keep the current behavior unchanged.

This is important because changing a signer to observer after signing would change the meaning of the participant in the audit/history after the signature already happened.

Do not move or remove `setParticipantRoleEnum()` from `populateSignRequest()`.

Add a unit test with an existing `SIGNED` signer receiving `ParticipantRole::OBSERVER`; the call must throw and the stored role must remain `SIGNER`.

src/utils/participantRole.ts
```diff
@@ -0,0 +1,61 @@
...
+export function isObserverParticipant(participant?: ParticipantLike | null): boolean {
+	if (!participant) {
+		return false
+	}
+
+	if (participant.participantRole === PARTICIPANT_ROLE.OBSERVER) {
+		return true
+	}
+
+	return participant.status === SIGN_REQUEST_STATUS.OBSERVING
```
vitormattos: Please use `participantRole` as the source of truth for the role.

`OBSERVING` is a status created by this feature, so we do not need it as a legacy fallback. With the current code, inconsistent data such as `participantRole: signer` with `status: OBSERVING` is treated as an observer.

The backend already stores the explicit role. The frontend should not silently change the role based on status. Please update the related unit test too.

playwright/support/request-signature.ts
vitormattos, line 1: Please add one focused Playwright test for the observer workflow.

The PR changes many existing E2E tests, but none tests the new feature end to end.

The scenario should:

1. enable `enable_observer_profile`;
2. add one signer and one observer;
3. request signatures;
4. verify both roles are shown separately;
5. open the observer link;
6. verify the observer goes to validation and cannot enter the signing flow.

This will test the main business flow instead of only the individual frontend and backend pieces.

We also need similar scenario at integration tests with Behat to validate the API contract too.

tests/php/Unit/Service/MailServiceTest.php
```diff
@@ -87,6 +87,36 @@ public function testSuccessNotifyUnsignedUser():void {
 		$this->assertNull($actual);
 	}
 
+	public function testSuccessNotifyUnsignedObserverUsesValidationLink(): void {
+		$this->mockRequest();
+		$this->mockStrategy(MailSenderStrategyPolicy::STRATEGY_SYSTEM);
+		$this->mailer->expects($this->once())
+			->method('send');
+
+		$signRequest = $this->createMock(SignRequest::class);
```
vitormattos: Could we use a real `SignRequest` entity here instead of mocking `__call()`?

This test only needs entity data such as UUID, file ID, display name, and participant role. Using the real entity will make the test easier to read and less coupled to the entity magic methods.

The same approach can be used for the file entity if possible.

vitormattos (changes requested):

lib/Service/Policy/Provider/ObserverProfile/ObserverProfilePolicy.php
```diff
@@ -0,0 +1,102 @@
...
+			self::KEY,
+		];
+	}
+
+	#[\Override]
+	public function get(string|\BackedEnum $policyKey): IPolicyDefinition {
+		return match (PolicyKeyNormalizer::normalize($policyKey)) {
+			self::KEY => new PolicySpec(
+				key: self::KEY,
+				defaultSystemValue: false,
```
vitormattos, line 35: This policy should also be stored in the file policy snapshot.

For example, if the policy is enabled when a request with observers is created, disabling the policy later should not break this existing request.

An `ObserverProfileFilePolicyApplier` can follow the existing file policy appliers and store the effective `enable_observer_profile` value in `policy_snapshot` when the file is created.

For existing requests, observer validation should use the stored snapshot. The current policy value should only affect new requests.

A separate Behat scenario can test this flow: create a request with an observer while the policy is enabled, disable the policy, confirm that the existing request still works, and confirm that a new request with an observer is rejected.

vitormattos (changes requested):

vitormattos:
If possible, could you also add a few screenshots to the PR description showing the observer flow?

It would be useful to show:

* the `Add` menu with Signer and Observer;
* the Signers and Observers sections;
* the validation page with an observer.

Also, sorry for reviewing this PR while it is still a draft. I added the comments now only to help keep the work close to the expected flow and avoid bigger changes later.

For the next review, it is better to add me as a reviewer and request the review when the PR is ready and is no longer a draft. After pushing new changes, when the PR is ready for review again, use the GitHub option to request my review again. This makes it clear when a new review is needed.

lfals:
## Observer participant workflow

End-to-end screenshots captured from the local Nextcloud environment using the PR branch.

### 1. Enable the observer profile policy

Observer profile policy enabled

### 2. Add participants

The enabled policy changes the **Add** action into a menu with separate **Signer** and **Observer** options.

Add signer or observer menu

The observer has a dedicated form and role-specific labels.

Add observer form

The request sidebar displays signers and observers in separate sections.

Signer and observer sections

### 3. Request signatures

The signer moves to **Ready to sign**, while the observer moves to **Observing**.

Signature request created

### 4. Signer flow

The signer receives access to the signing action.

Signer access with Sign document action

After signing, the validation page shows the document as signed and keeps the observer listed separately.

Successfully signed document

### 5. Observer flow

The observer link opens the validation-only view. The observer remains in **Observer** status and no signing action is available.

Observer validation-only view

lfals:
> If possible, could you also add a few screenshots to the PR description showing the observer flow?
> 
> It would be useful to show:
> 
> * the `Add` menu with Signer and Observer;
> * the Signers and Observers sections;
> * the validation page with an observer.
> 
> Also, sorry for reviewing this PR while it is still a draft. I added the comments now only to help keep the work close to the expected flow and avoid bigger changes later.
> 
> For the next review, it is better to add me as a reviewer and request the review when the PR is ready and is no longer a draft. After pushing new changes, when the PR is ready for review again, use the GitHub option to request my review again. This makes it clear when a new review is needed.

No problem, I left it as a draft after the changes you requested. I was fixing the tests that broke. I'm going to add the observer flow and let you know when it's ready for review.

@vitormattos

lib/Service/Policy/Provider/ObserverProfile/FilePolicy/ObserverProfileFilePolicyApplier.php
```diff
@@ -0,0 +1,57 @@
...
+use OCA\Libresign\Db\File as FileEntity;
+use OCA\Libresign\Service\Policy\AbstractFilePolicyApplier;
+use OCA\Libresign\Service\Policy\Model\ResolvedPolicy;
+use OCA\Libresign\Service\Policy\Provider\ObserverProfile\ObserverProfilePolicy;
+use OCA\Libresign\Service\Policy\Provider\ObserverProfile\ObserverProfilePolicyValue;
+use OCP\IUser;
+
+final class ObserverProfileFilePolicyApplier extends AbstractFilePolicyApplier {
+	#[\Override]
+	public function apply(FileEntity $file, array $data): void {
```
vitormattos, line 20: I think that the observer policy snapshot also needs to be stored on the envelope itself.

Child files go through `FilePolicyApplier`, but the envelope is created directly by `EnvelopeService::createEnvelope()`. Sign requests are also linked to the envelope, so the envelope can have no `enable_observer_profile` snapshot.

Store the same observer policy snapshot on the envelope when it is created.

Add a regression test for this flow:

1. enable the observer policy;
2. create an envelope with an observer;
3. disable the policy;
4. update the existing envelope;
5. the update must still accept the existing observer.

lib/Service/Policy/Provider/ObserverProfile/ObserverProfilePolicyService.php
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	public function isEnabled(?FileEntity $file = null): bool {
+		$snapshotValue = $this->getSnapshotValue($file);
+		if ($snapshotValue !== null) {
+			return $snapshotValue;
+		}
+
+		return ObserverProfilePolicyValue::normalize(
+			$this->policyService->resolve(ObserverProfilePolicy::KEY)->getEffectiveValue(),
```
vitormattos, line 26: An existing file without this policy in `policy_snapshot` must not use the current policy value.

Files created before this feature do not have `enable_observer_profile` in the snapshot. If the admin enables the policy later, these old files could start allowing observers.

When `isEnabled()` receives a file:

* return the stored snapshot value when it exists;
* if the snapshot does not contain this policy, return `false`.

Only use the current policy value when no file is provided, for a new request.

Add a unit test with an existing file without the snapshot and the current policy set to `true`; the result must still be `false`.

vitormattos (changes requested):

vitormattos (changes requested):
Only a small change:

At file appinfo/info.xml, change from:
```
16.0.0-dev.1
```
to:
```
16.0.0-dev.2
```

This will make all environments of all developers to run the migration because will identify that the version was changed.

vitormattos:
Could you also check this?

> It would be useful to show:
>
> * the `Add` menu with Signer and Observer;
> * the Signers and Observers sections;
> * the validation page with an observer.

vitormattos:
I found another case during manual testing.

I created a signature request with one signer while the Observer policy was disabled. Then I enabled the Observer policy in Administration Settings and opened the same request again.

The UI correctly showed `Add → Observer`, but saving the observer returned:

`Observer participants are not enabled`

This happens because the frontend uses the current policy, while the backend keeps the old `false` value from the file snapshot.

I think the expected behavior should be:

1. If an existing request has Observer disabled and the current policy becomes enabled, Observer can be added while the request can still be edited.
2. After an Observer is added to this request, disabling the global policy later must not remove Observer from this existing flow. It should still be possible to keep and add observers to this request.
3. New requests should continue to use the current policy value.

The saved policy state may need to move from `false` to `true` when Observer is first added to an existing request. Once enabled for that request, it should stay enabled for this flow.

There is also a frontend error handling problem in this case. The API returned HTTP 422 with the message `Observer participants are not enabled`, but the UI did not show this message.

`IdentifySigner.vue` should show the error message returned by the API when saving a participant fails, instead of losing it in the `catch`.

Please add regression tests for both parts:

* backend/integration: create a request with Observer disabled, enable the policy, add an observer, disable the policy again, and confirm that the same request can still use observers;
* frontend: make the participant save request fail with a 422 OCS response and confirm that the API error message is shown to the user and the participant modal stays open.

playwright/e2e/observer-participant-flow.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+		ST: 'Rio de Janeiro',
+		O: 'LibreSign',
+		L: 'Rio de Janeiro',
+	})
+
+	await setSystemPolicy(page.request, 'enable_observer_profile', JSON.stringify(true))
+	await setSystemPolicy(page.request, 'make_validation_url_private', '0')
+	await setSystemPolicy(
+		page.request,
+		'identify_methods',
```
vitormattos, line 35: I found a problem during manual testing with mixed identify methods.

I enabled Account and Email identification, added the admin as a signer using Account, and added an observer using Email.

The signature request was created, but the observer never received the email with the link to view the document.

The current Playwright test uses Email for both the signer and the observer, so this case is not covered.

We should cover this at different levels:

* integration/Behat: create a request with an Account signer and an Email observer and confirm that the observer sign request is created with the correct role and notification state;
* PHPUnit: confirm that an observer with Email identification is marked for notification and that the mail listener sends the observer email with the validation link;
* Playwright: confirm the full user flow, including that the external observer receives the email, opens the validation page without an account, and cannot sign.

This will also define the expected journey for an observer that does not have a Nextcloud account.

src/components/RightSidebar/RequestSignatureTab.vue
```diff
@@ -695,11 +803,11 @@ function isSignerSigned(signer: Partial<EditableRequestSigner>) {
 }
 
 const canEditSigningOrder = computed(() => (signer: Partial<EditableRequestSigner>) => {
-	if (isOriginalFileDeleted.value) {
+	if (isOriginalFileDeleted.value || isObserverParticipant(signer)) {
 		return false
 	}
```
vitormattos, line 834: Observers also need a way to receive the notification again.

At the moment `canSendReminder()` always returns `false` for observers. In the UI, an observer can only have the message edited or be deleted, but there is no action to send the document link again.

The action does not need to be called `Send reminder`, because the observer is not waiting to sign. Something like `Send notification` can be clearer for observers.

The notification should use the observer validation link and include the custom message when it is set.

Add frontend coverage for this action and backend coverage to confirm that sending the notification for an observer uses the validation link.

vitormattos (changes requested):
I also found a frontend issue during manual testing.

Observers are shown in the Signature positions screen and can be selected to add a visible signature position.

This is in `src/components/Request/VisibleElements.vue`, where both `sidebarSigners` and `pdfEditorSigners` use all participants without filtering by role.

Only participants with the `signer` role should be available in this screen.

Please filter observers from both lists and add a frontend test with one signer and one observer. The test should confirm that only the signer is shown.

vitormattos:
The same rule also needs backend protection.

`lib/Service/FileElementService.php` currently accepts the `signRequestId` for a visible element without checking if this participant can sign.

The API must reject a visible signature element when the `signRequestId` belongs to an observer.

Please add a unit test for this validation and an integration/Behat scenario that sends an observer `signRequestId` and confirms that the API rejects it. Keep a signer case to confirm the normal flow still works.

vitormattos (changes requested):
I did another round of manual testing and found some problems with the observer experience.

First, I tested an observer without an account, using Email identification. In this case, the observer opens the public validation page:

Image

I think using the public validation page for an observer without an account can make sense, because this user only needs to follow the document and receive the flow notifications.

However, the page does not look correct. The validation layout is visually broken and does not look like the normal validation experience. Please check the public observer view and make sure the participant information and document validation layout are rendered correctly.

Then I tested an authenticated observer, using Account identification. The document correctly appeared in this user's LibreSign file list, but when I opened it, the sidebar looked like the signer flow and showed signing actions:

Image

I think an authenticated observer should have a read-only view of the request.

The observer should be able to see:

* signers and observers;
* current status and progress;
* the document;
* existing signature positions, if any.

But the observer should not be able to add or delete participants, edit messages, change signing order, configure signature positions, send notifications, use participant action menus, or sign the document.

The experience could be similar to the requester sidebar below, but in read-only mode and without action buttons:

Image

For signature positions, something like `View signature positions` could make sense instead of the edit action.

Please also add frontend and Playwright tests for both observer flows:

* unauthenticated observer: opens the public validation page and the layout is rendered correctly;
* authenticated observer: opens the observed document in read-only mode, can see the request information, and cannot use edit or signing actions.

src/components/RightSidebar/RequestSignatureTab.vue
```diff
@@ -863,7 +1044,20 @@ const showSaveButton = computed(() => {
...
 
+const showViewPositionsButton = computed(() => {
+	if (!isReadOnlyObserver.value || isOriginalFileDeleted.value) {
+		return false
+	}
+	if (shouldLoadDetail.value && !isCurrentFileDetailed.value) {
+		return false
+	}
+	return showsPositionEditor.value
+})
```
vitormattos, line 1059: I found another case during manual testing.

I created a request with only an observer and no signer. In this case, the signature positions action is still shown.

Image

Since there is no signing participant, there cannot be any signature position to configure or view.

Could we hide both `Setup signature positions` and `View signature positions` when `signingParticipantCount === 0`?

Please also add regression coverage for both requester and observer views with observers only and no signer.

src/utils/filesListSidebar.ts
```diff
@@ -69,7 +84,10 @@ export async function openFilesListSidebarForFile<TFile extends FilesListSidebar
 		return detailedFile
 	}
 
-	if (detailedFile && options.filesStore.canRequestSign === true) {
+	if (detailedFile && (
+		options.filesStore.canRequestSign === true
+		|| isObserverView(detailedFile, options.filesStore)
+	)) {
```
vitormattos, line 106: I found another case during manual testing.

I created a request where I am also the signer. After sending the request and refreshing the Files page, selecting the document opens the signing sidebar with Review and sign document instead of the normal request sidebar.

This seems to come from this condition because canSign() is checked before the requester context. A user can be both the requester and a signer, so canSign() alone is not enough to decide which sidebar should open.

I think the Files view should keep the request/management sidebar when the current user owns or manages this request, even if this same user can also sign it. The signing sidebar should be used when the user is acting only as a signer.

Please also add a regression test for the case where the current user is both requester and signer. The existing test covers canSign: true with canRequestSign: false, but not the case where both roles are true.

vitormattos (changes requested):

src/components/Request/VisibleElements.vue
```diff
@@ -651,7 +663,8 @@ function isSelectedSigner(signer: EditableRequestSigner): boolean {
 }
 
 async function showModal() {
-	if (!canRequestSign.value) {
+	// Observers cannot request signatures, but may open positions in read-only mode.
+	if (!canRequestSign.value && !filesStore.isObservingOnly()) {
```
vitormattos, line 667: I found a problem while manually testing this new observer path.

As an authenticated observer, I clicked View signature positions. The modal opens, but the PDF is not rendered and PDFElements fails with:

UnknownErrorException: n.toHex is not a function

This observer path is enabled by this change, so I think we should make sure it works end to end before merging.

The exact toHex error may come from PDFElements or from invalid data being passed to it, so could we check what PDF response/data the observer receives here and fix the root cause?

Please also extend the authenticated observer Playwright test to click View signature positions and confirm that the PDF is actually rendered. The current coverage only confirms that the action is visible.

vitormattos: I tested this again with the latest branch and the problem is still present.

The modal opens, but the PDF is not rendered and the browser console still reports:

`UnknownErrorException: n.toHex is not a function`

I think the new Playwright coverage can pass even with this error.

`page.getByText('UnknownErrorException')` only checks the DOM, but this exception is reported in the browser console and is not rendered as page text.

Also, the selector `canvas, .pdfViewer, .page, [class*="pdf"]` looks broad enough to match the PDF editor structure even when a PDF page was not successfully rendered.

Could we keep investigating the root cause and make the E2E test fail on this real error? For example, it could capture `pageerror`/console errors and verify a real rendered PDF page or canvas after PDF.js finishes loading.

lib/ResponseDefinitions.php
```diff
@@ -209,6 +211,7 @@
  *     signed: ?string,
  *     status: 0|1|2|3,
```
vitormattos: `OBSERVING` uses status `4`, but this response contract still allows only `0|1|2|3`.

Could we include `4` here and regenerate the OpenAPI/TypeScript types?

Please also keep a test confirming that an observer status is represented correctly by the API contract.

lib/Service/MailService.php
```diff
@@ -91,27 +112,48 @@ public function notifySignDataUpdated(SignRequest $data, string $email, ?string
...
+
 		$emailTemplate = $this->mailer->createEMailTemplate('settings.TestEmail');
-		// TRANSLATORS Email subject notifying a signer that a document is ready for their digital signature.
-		$emailTemplate->setSubject($this->l10n->t('LibreSign: A document is ready for your signature'));
-		$emailTemplate->addHeader();
-		// TRANSLATORS Email heading shown above a document awaiting the recipient's signature.
-		$emailTemplate->addHeading($this->l10n->t('Document to sign'), false);
+		if ($isObserver) {
+			// TRANSLATORS Email subject notifying an observer that a document is available to view.
+			$emailTemplate->setSubject($this->l10n->t('LibreSign: A document is ready for signature'));
```
vitormattos: The observer email still says that the document is ready for signature, although this participant cannot sign it.

The heading and button already use `Document to view` and `View`, so could we also use observer-specific wording for the subject and body, for example `A document is ready to view`?

Please update the related mail and Playwright expectations too.

lib/Listener/NotificationListener.php
```diff
@@ -213,12 +216,20 @@ private function sendCanceledNotification(
 	 * @psalm-return array{type: 'file', id: string, name: string, path: string, link: string}
 	 */
 	protected function getFileParameter(SignRequest $signRequest, FileEntity $libreSignFile): array {
+		if ($signRequest->isObserver()) {
+			$link = $this->url->linkToRouteAbsolute('libresign.page.validationFilePublic', [
+				'uuid' => $libreSignFile->getUuid(),
+			]);
+		} else {
```
vitormattos, line 225: I found an inconsistency in the navigation for observers identified by Account.

`sendSignNotification()` only creates this Nextcloud notification when the identify method is `account`, so the recipient here is an authenticated Nextcloud user.

For an authenticated observer, we now have a read-only LibreSign view: the document appears in the Files list and the observer can open it to see the request information without signing or editing.

However, this branch sends the observer to `validationFilePublic`. This means the same authenticated observer gets two different experiences:

- opening the document from LibreSign Files → authenticated read-only view;
- opening the Nextcloud notification → public validation page.

Could we make the Account observer notification open the authenticated read-only document view instead?

The public validation page should still be used for external observers, such as observers identified only by Email.

Please also check the equivalent link in the Activity listener and add coverage confirming that an Account observer notification points to the authenticated observer view.

vitormattos (changes requested):

src/components/Request/VisibleElements.vue
vitormattos: I investigated this further and confirmed that the error is not specific to the observer flow.

The same `UnknownErrorException: n.toHex is not a function` can be reproduced in Chrome when the requester opens the existing `Setup signature positions` flow.

The root cause is that LibreSign overrides the worker managed by `@libresign/pdf-elements` with `pdfjs-dist/build/pdf.worker.min.mjs`.

Removing this override fixes PDF rendering in both Chrome and Firefox.

I opened #8380 with the fix and also strengthened the existing Playwright coverage to verify that a real PDF page is rendered and to fail on `pageerror`.

So this is a pre-existing LibreSign PDF worker issue, not a regression introduced by this observer PR. I think this thread can be resolved here and tracked by #8380.

lib/Controller/LibresignTrait.php
```diff
@@ -53,6 +53,10 @@ private function loadEntitiesFromUuid(string $uuid): void {
 	 */
 	public function validateSignRequestUuid(string $uuid): void {
 		$this->loadEntitiesFromUuid($uuid);
+		// Observers need PDF access for read-only viewing, but must not pass signing validation.
+		if ($this->signRequestEntity?->isObserver()) {
+			return;
```
vitormattos: I reproduced a real authorization bug caused by this early return.

An anonymous request using an observer UUID can successfully call `POST /api/v1/id-docs` and receives `200 OK`. The observer role should be read-only, so this write must be rejected.

Please do not fix this only in `IdDocsController`, because `RequireSignRequestUuid` is shared by multiple endpoints.

We need two explicit authorization paths:

- signer-only authorization for signing/write operations;
- participant read-only authorization for endpoints that must allow both signers and observers, such as document/PDF viewing.

One possible approach is a new attribute such as `RequireParticipantUuid` for the read-only participant endpoints, while `RequireSignRequestUuid` keeps the current signer validation.

The observer must still be able to view the document, but must not be able to use observer UUID access to reach write/signing endpoints.

Please keep the integration scenario that currently gets `200` from `POST /api/v1/id-docs` as a regression test, and review the other `RequireSignRequestUuid` endpoints for the same authorization boundary.

vitormattos (changes requested):

vitormattos:
Sorry for the many back-and-forth review rounds on this PR.

The change touches many parts of LibreSign, so it ended up being a more difficult review than expected. Thank you for your patience and for fixing the issues as we found them.

I think we are close now. Before the final review, there are two points left:
- update the PR description to match the current implementation;
- address my latest inline comment about the observer authorization flow.

After that, I will review the PR again as a whole.

lfals:
> Sorry for the many back-and-forth review rounds on this PR.
> 
> The change touches many parts of LibreSign, so it ended up being a more difficult review than expected. Thank you for your patience and for fixing the issues as we found them.
> 
> I think we are close now. Before the final review, there are two points left:
> 
> * update the PR description to match the current implementation;
> * address my latest inline comment about the observer authorization flow.
> 
> After that, I will review the PR again as a whole.

Don't worry, it's the right thing to do. It's a long PR too

src/views/Settings/PolicyWorkbench/settings/observer-profile/ObserverProfileRuleEditor.vue
```diff
@@ -0,0 +1,89 @@
...
+}>()
+
+const emit = defineEmits<{
+	'update:modelValue': [value: EffectivePolicyValue]
+}>()
+
+// TRANSLATORS Toggle title for the policy that enables observer participants on signature requests.
+const title = t('libresign', 'Enable observer profile')
+// TRANSLATORS Toggle description explaining that document owners can add view-only participants.
+const description = t('libresign', 'When enabled, document owners can add observers who track progress without signing.')
```
vitormattos, line 52: There is one policy interaction that is still not clear to the administrator.

An Email observer uses the public validation page, but when `make_validation_url_private` is enabled that page requires authentication. An external observer without a Nextcloud account will receive the email but will not be able to open the document.

I think both policies can still be enabled together, but we should warn about this consequence.

Could we show a warning when enabling Observer while validation access is authenticated-only, and also when making validation private while Observer is enabled?

Please also add regression coverage for this combination, including the Email observer flow.

vitormattos (changes requested):

lib/Service/Policy/Provider/Helper/SiblingPolicyEffectiveBoolReader.php
```diff
@@ -0,0 +1,34 @@
...
+
+	public function getEffectiveBool(string $policyKey): bool {
+		if ($this->depth > 0) {
+			return false;
+		}
+
+		$this->depth++;
+		try {
+			return \OCP\Server::get(PolicyService::class)
+				->resolve($policyKey)
```
vitormattos: I think this can resolve the sibling policy in the wrong context.

`resolvedStateMeta` is calculated while a specific policy context is being resolved, but this helper ignores that context and calls `PolicyService::resolve()`, which resolves for the current user.

For example, when an admin edits an Observer rule for a group, the warning may use the admin's effective validation-access policy instead of the effective policy for that group.

Could we resolve the sibling policy using the same `PolicyContext` used to build the current policy state?

Please also add a test where the current admin value differs from the target group/user value and confirm that the warning meta follows the target scope.

vitormattos (changes requested):

tests/integration/features/bootstrap/FeatureContext.php
```diff
@@ -40,6 +40,29 @@ public function setOpenedEmailStorage(OpenedEmailStorage $storage): void {
...
+			} catch (\GuzzleHttp\Exception\ConnectException $exception) {
+				$attempts++;
+				if ($attempts >= $maxAttempts) {
+					throw $exception;
+				}
+				usleep(250000 * $attempts);
+			}
+		}
+	}
+
```
vitormattos: This global retry looks unrelated to the Observer feature.

It was added after the SQLite Behat run failed with `cURL error 52` / `Empty reply from server`, but this error only tells us that the test server did not return a response. It does not explain why the server stopped replying.

Overriding `sendRequest()` changes every Behat scenario and can retry POST, PATCH and DELETE requests too. If a write was already processed before the connection was lost, the retry can execute it twice. It can also hide a real PHP server or application failure.

Could we revert this change from this PR and investigate the SQLite/PHP built-in server failure separately?

If this is a general CI problem, it should have its own fix and regression coverage instead of changing the request behavior for the whole Behat suite.

lib/Service/Policy/Provider/ObserverProfile/ObserverProfilePolicyService.php
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	public function isEnabled(?FileEntity $file = null): bool {
+		if ($file instanceof FileEntity) {
+			$snapshotValue = ObserverProfilePolicyValue::getSnapshotEffectiveValue($file);
+			if ($snapshotValue === true) {
+				return true;
+			}
+			if ($snapshotValue === false) {
+				return $this->isLivePolicyEnabled();
```
vitormattos: Could we add a scoped-policy regression test for this fallback before changing the implementation?

For an existing request with `enable_observer_profile = false` in its snapshot, the code can check the live policy when an observer is added later.

I want to confirm that this uses the policy context of the target request, and not only the policy effective for the user doing the update.

A useful test would have different values for the manager/admin and for the target user or group.

Please cover this in PHPUnit and, if possible, also in Behat with group/user policy scopes.

We should test both directions:
- manager policy disabled, target policy enabled;
- manager policy enabled, target policy disabled.

If both cases already follow the target context, no fix is needed. If not, please keep the tests as regression coverage and use the correct target context when checking and upgrading the snapshot.

vitormattos (changes requested):

tests/integration/features/bootstrap/FeatureContext.php
lfals: @vitormattos Can i open a issue for the Behat problem?

lfals:
@vitormattos the current PHPUnit failures are unrelated to the observer work, do you prefere to fix in this PR ou open a issue for that?

vitormattos:
analyzing

vitormattos:
I investigated the PHPUnit failures and confirmed that they are not caused by your PR.

The same `SignatureStatusPluginTest` failures were reproduced on LibreSign `main` with the current Nextcloud `master`. The cause was the recent container resolution change from `nextcloud/server#64172`.

This is now fixed in `main` by #8407, where LibreSign was updated to follow the current Nextcloud dependency injection/container model.

So you do not need to fix the `SignatureStatusPluginTest` failures in this PR. Please update your branch with the current `main` and rerun the CI. We can then check any failures that remain.

I also recommend using rebase instead of merging `main` into your branch. This keeps the history cleaner and avoids extra merge commits.

If you do not have the LibreSign repository configured as `upstream` yet, you can do:

```bash
git remote add upstream https://github.com/LibreSign/libresign.git
git fetch upstream
git rebase upstream/main
git push --force-with-lease
```
If upstream already exists, you only need the last three commands.

Thanks for pointing out that these failures looked unrelated to the Observer changes.

vitormattos (approved):

vitormattos:
Thanks a lot for all the work on this PR, Luiz.

This one was not simple at all. We had many rounds of review, more than 40 comments, and a very large number of files to check. It also touches many important parts of LibreSign: database, frontend, backend, API, security, and some core flows of the project.

Because of this, the review took longer than usual, but I think the result is much better and safer now.

Thank you for your patience with all the feedback and for taking the time to investigate and fix each point. I know this was a lot of work.

Great job getting this one to the finish line.

PR #8172, opened 2026-09-01
fix(pdf): identify PdfParser/PdfValidator errors with semantic error codes

ajayraho:
Related to: #8117

## 📝 Summary

Standardizes error responses from `PdfParser` and `PdfValidator` by having
their `LibresignException` throws explicitly pass
`Http::STATUS_UNPROCESSABLE_ENTITY`, instead of relying on
`InjectionMiddleware::getStatusCodeFromException()`'s implicit default.
This matches the existing pattern already used in
`Service/Policy/Runtime/PolicyContextFactory.php`.

This addresses one item from the #8117 tracking issue (which covers several
exception/error-code standardizations across the codebase — not closing it
here since other items are handled by separate PRs).

## 🧪 How to test

1. Upload a file that is not a valid PDF (e.g. a `.txt` renamed to `.pdf`) through the sign flow.
2. Confirm the API response returns HTTP 422 (Unprocessable Entity) instead of the previous default.
3. Upload a PDF that has DocMDP certification disallowing further changes, then attempt to add a signature.
4. Confirm that also returns HTTP 422.

## ⚙️ API / Back‑end changes

- [x] `PdfParser::parse()` and `PdfValidator::validateDocMdpRestrictions()` now explicitly set the exception's HTTP status code to 422 instead of relying on the middleware default
- [x] Unit tests added – updated `PdfParserTest` and `PdfValidatorTest` to assert on `expectExceptionCode()`
- [ ] Capabilities updated (if applicable) – N/A, no capability change
- [ ] Documentation updated (if applicable) – N/A, internal error-handling change only
- [ ] API documentation updated with `composer openapi` – N/A, no OpenAPI-annotated signature changed

## ✅ Checklist

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Service/File/Pdf/PdfParser.php
+++ b/lib/Service/File/Pdf/PdfParser.php
@@ -12,11 +12,15 @@ namespace OCA\Libresign\Service\File\Pdf;
 use OCA\Libresign\Exception\LibresignException;
 use OCA\Libresign\Vendor\Smalot\PdfParser\Document;
 use OCA\Libresign\Vendor\Smalot\PdfParser\Parser;
-use OCP\AppFramework\Http;
 use OCP\IL10N;
 use Psr\Log\LoggerInterface;
 
 class PdfParser {
+	/**
+	 * Semantic error code for a file that could not be parsed as a valid PDF document.
+	 */
+	public const int CODE_INVALID_PDF = 4220;
+
 	public function __construct(
 		private LoggerInterface $logger,
 		private IL10N $l10n,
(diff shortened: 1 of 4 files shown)


vitormattos (changes requested):
Sorry for the change in direction. The discussion in #8117 changed after your proposal.

For this PR, I think the error codes should stay close to the classes that produce them.

For example, `PdfParser` could define a semantic error constant for the parsing failure, and `PdfValidator` another one for the DocMDP restriction. The tests can then use these constants with `expectExceptionCode()`.

This avoids using `422` as the error identity, avoids adding unrelated constants to `LibresignException`, and also avoids creating many exception classes.

You can use #8103 as a reference for the semantic error code idea, but I think the constants should be defined in the classes responsible for each error.

Please also update the PR title and description, because this PR is about identifying the errors in a stable way, not changing the HTTP response status.

ajayraho:
Thanks for the clarification @vitormattos , updated to keep the semantic
error identity separate from the HTTP status.

- `PdfParser::CODE_INVALID_PDF = 4220`
- `PdfValidator::CODE_DOCMDP_RESTRICTED = 4221`

Both follow the #8103 convention and live on the class that produces the
error rather than on `LibresignException`. Tests now assert on these
constants directly. Title and description updated to reflect this.

vitormattos (approved):

PR #8196, opened 2026-09-02
refactor(exception): keep the semantic error codes next to the classes that raise them

maia-andre:
Refs: #8117 (follow-up of #8103)

## 📝 Summary

In #8117 @vitormattos asked to keep each semantic error code close to the class that produces the error, instead of collecting them in `LibresignException`, which is what #8172 already does (`PdfParser::CODE_INVALID_PDF`, `PdfValidator::CODE_DOCMDP_RESTRICTED`). The two constants introduced by #8103 were the only ones still living in `LibresignException`, so this PR moves them:

| Before | After |
| --- | --- |
| `LibresignException::CODE_INVALID_TOKEN` | `AbstractIdentifyMethod::CODE_INVALID_TOKEN` (4010) |
| `LibresignException::CODE_TWOFACTOR_GATEWAY_NOT_ENABLED` | `TwofactorGatewayService::CODE_NOT_ENABLED` (5030) |

Values, messages and the thrown exception class are unchanged, so nothing changes for API clients. `LibresignException` has no error-code constants left.

## 🧪 How to test

1. `vendor/bin/phpunit -c tests/php/phpunit.xml tests/php/Unit/Service/IdentifyMethod/SignatureMethod/` — `EmailTokenTest` and `TokenServiceTest` assert the codes through their new owners (89 tests).
2. `git grep 'LibresignException::CODE_'` returns nothing.

## ⚙️ API / Back‑end changes

- [x] Constants moved to the classes that throw the exceptions; no behavior change
- [x] Unit tests updated
- [ ] Capabilities updated (if applicable) – N/A
- [ ] Documentation updated (if applicable) – N/A
- [ ] API documentation updated with the command `composer openapi` if necessary – N/A

### 🚧 Tasks

- [ ] Backport notes: `stable35` has the same files as `main` (cherry-pick applies clean). `stable32`–`stable34` only have `CODE_INVALID_TOKEN` (untyped constant, from the adapted backports #8164–#8166) and no `TwofactorGatewayService`, so they need the same adaptation as before: move that single constant to `AbstractIdentifyMethod` and update `EmailTokenTest`. I can open those manually if you want them.

## ✅ Checklist

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Exception/LibresignException.php
+++ b/lib/Exception/LibresignException.php
@@ -14,13 +14,6 @@ use JsonSerializable;
  * @codeCoverageIgnore
  */
 class LibresignException extends \Exception implements JsonSerializable {
-	/**
-	 * Error codes follow the closest HTTP status with one extra digit
-	 * (401 -> 4010), so they are not mistaken for HTTP status codes.
-	 */
-	public const int CODE_INVALID_TOKEN = 4010;
-	public const int CODE_TWOFACTOR_GATEWAY_NOT_ENABLED = 5030;
-
 	#[\Override]
 	public function jsonSerialize(): mixed {
 		return ['message' => $this->getMessage()];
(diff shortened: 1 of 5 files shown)


vitormattos (approved):
Much better now! A small detail that makes a big difference.

PR #8236, opened 2026-09-03
fix: encode native signature stamps as WinAnsi

### LINKED ISSUE #8155: PhpNative engine: non-ASCII characters in the signature stamp are rendered as mojibake
### Describe the bug

With the `PhpNative` signature engine, non-ASCII characters in the visible signature stamp are
rendered as mojibake. The stored `signature_text_template` is correct — the corruption happens at
render time.

A French template produced this on the signed PDF:

| Template (stored, correct) | Rendered in the PDF |
|---|---|
| `Signé avec LibreSign` | `Sign^ avec LibreSign` |
| `Émetteur : Phenix Tech` | `^ metteur : Phenix Tech` |

### To reproduce

1. Set `signature_engine` to `PhpNative`.
2. Set a `signature_text_template` containing accented characters, e.g.
   `Signé avec LibreSign\n{{SignerCommonName}}\nÉmetteur : {{IssuerCommonName}}`.
3. Sign a document and open the result — the accented characters are mangled.

`occ config:app:get libresign signature_text_template` still returns the correct UTF-8, confirming
the problem is in the rendering path and not in storage.

### Expected behavior

The visible stamp should render the template as stored, at least for the Latin-1 range.

### Root cause

Two things combine in `lib/Handler/SignEngine/PhpNativeHandler.php`:

**1. No encoding conversion.** `escapePdfText()` (line 398) only escapes PDF delimiters:

```php
private function escapePdfText(string $value): string {
    $value = str_replace('\\', '\\\\', $value);
    $value = str_replace('(', '\\(', $value);
    $value = str_replace(')', '\\)', $value);
    return $value;
}
```

The raw UTF-8 bytes are then written straight into a PDF literal string via
`sprintf("(%s) Tj\n", $escaped)`. `é` (`0xC3 0xA9`) therefore reaches the content stream as two
separate byte codes.

**2. A single-byte base font with no declared encoding** (line 336):

```php
resources: [
    'Font' => [
        'F1' => [
            'Type' => '/Font',
            'Subtype' => '/Type1',
            'BaseFont' => '/Helvetica',
        ],
    ],
],
```

There is no `/Encoding` entry, so the viewer falls back to Helvetica's built-in StandardEncoding.
Each UTF-8 byte is mapped independently, which is exactly the observed output.

### Possible directions

- **Minimal**: convert the string to WinAnsi and declare `/Encoding /WinAnsiEncoding` in the font
  dictionary. Covers Latin-1, so most Western European templates, and is a small change. Characters
  outside Latin-1 would still need a fallback rather than silent corruption.
- **Complete**: embed a TrueType subset with a proper `/Encoding` (or a Type0/Identity-H font) to
  support the full Unicode range, at the cost of a larger PDF and a font to ship.

Either way, it would be worth failing loudly — or transliterating explicitly — rather than emitting
bytes that render as mojibake, since the result lands on a signed document that cannot be corrected
afterwards without invalidating the signature.

### Why this matters more than it looks

This is engine-specific: `JSignPdf` renders the same template correctly through iText. The two
engines are therefore not interchangeable in both directions.

It compounds with #8145: `JSignPdf` currently cannot timestamp against any TSA that rejects SHA-1,
so instances that need RFC 3161 timestamping are pushed towards `PhpNative` — and there they must
restrict their signature stamp to ASCII. Right now there is no engine that does both correct
timestamping and correct non-ASCII rendering.

### Workaround

Restrict `signature_text_template` to ASCII. In French this is doable with some care — `Signature`,
`Certificat`, `Date` carry no accents — but it constrains the wording, and nothing warns the
administrator that the template will be corrupted.

### Environment

- LibreSign **13.3.0**
- Nextcloud **33.0.8** (All-in-One)
- Signature engine: `PhpNative` (`jeidison/signer-php`)
- Certificate engine: OpenSSL, self-signed internal root CA

w3lld1:
Fixes #8155

## Summary

- I convert UTF-8 signature stamp text to Windows-1252 before writing PDF literal strings and declare `/WinAnsiEncoding` for Helvetica.
- I use multibyte-aware length and substring operations so wrapping and centering do not split UTF-8 characters.
- I reject characters outside WinAnsi explicitly instead of writing corrupted bytes into a signed document.
- I added focused regression tests for accented text, multibyte wrapping, and unsupported characters.

## How to test

1. Configure the `PhpNative` signature engine.
2. Set a signature text template containing accented characters, for example `Signé par Renée`.
3. Sign a document and confirm the visible stamp preserves the accents.
4. Try a character outside WinAnsi and confirm signing fails explicitly rather than producing mojibake.

## Validation

- Focused `SignatureStampAppearanceBuilderTest`: 3 tests, 6 assertions passed on PHP 8.4.24.
- PHP CS Fixer dry run passed for both changed files.
- PHP syntax checks and `git diff --check` passed.
- The full Nextcloud-backed PHPUnit matrix was deferred to CI because this local checkout does not include a Nextcloud server bootstrap.

## API / back-end changes

- [x] Native PDF stamp text now uses WinAnsi bytes with a matching font encoding.
- [x] Unit tests cover the changed behavior.
- [x] No API or OpenAPI changes.

## Checklist

- [x] I have read and followed the contribution guide.
- [x] The commit uses Conventional Commits and includes a DCO sign-off.

### DIFF
--- a/lib/Service/SignatureStampPreview/SignatureStampAppearanceBuilder.php
+++ b/lib/Service/SignatureStampPreview/SignatureStampAppearanceBuilder.php
@@ -65,7 +65,7 @@ class SignatureStampAppearanceBuilder {
 				$nameY = $nameStartY;
 				$estimatedCharWidth = $nameFontSize * 0.52;
 				foreach ($nameLines as $nameLine) {
-					$lineWidth = strlen($nameLine) * $estimatedCharWidth;
+					$lineWidth = mb_strlen($nameLine) * $estimatedCharWidth;
 					$nameX = max($leftPadding, ($leftHalfW - $lineWidth) / 2.0);
 					$escaped = $this->escapePdfText($nameLine);
 					$stream .= "BT\n";
@@ -105,6 +105,7 @@ class SignatureStampAppearanceBuilder {
 						'Type' => '/Font',
 						'Subtype' => '/Type1',
 						'BaseFont' => '/Helvetica',
+						'Encoding' => '/WinAnsiEncoding',
 					],
 				],
 			],
@@ -122,7 +123,7 @@ class SignatureStampAppearanceBuilder {
 
 		$estimatedCharWidth = max(1.0, $fontSize * 0.52);
 		$maxChars = max(1, (int)floor($availableWidth / $estimatedCharWidth));
-		if (strlen($trimmed) <= $maxChars) {
+		if (mb_strlen($trimmed) <= $maxChars) {
 			return [$trimmed];
 		}
 
@@ -134,7 +135,7 @@ class SignatureStampAppearanceBuilder {
 			}
 
 			$candidate = $current === '' ? $word : $current . ' ' . $word;
-			if (strlen($candidate) <= $maxChars) {
+			if (mb_strlen($candidate) <= $maxChars) {
 				$current = $candidate;
 				continue;
 			}
@@ -144,9 +145,9 @@ class SignatureStampAppearanceBuilder {
 				$current = '';
 			}
 
-			while (strlen($word) > $maxChars) {
-				$result[] = substr($word, 0, $maxChars);
-				$word = substr($word, $maxChars);
+			while (mb_strlen($word) > $maxChars) {
+				$result[] = mb_substr($word, 0, $maxChars);
+				$word = mb_substr($word, $maxChars);
 			}
 
 			$current = $word;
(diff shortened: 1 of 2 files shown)


vitormattos (commented):
I think we should fix this first in [`LibreSign/xobject-template`](https://github.com/LibreSign/xobject-template).

This package is responsible for building the XObject content and resources, so Unicode font support should be handled there instead of adding a WinAnsi limitation in LibreSign.

We also need to check if [`signer-php`](https://github.com/jeidison/signer-php) needs any change to support the font resources generated by `xobject-template`.

After this is ready, we can release a new version of `xobject-template`, add it to LibreSign, and use it from `SignatureStampAppearanceBuilder`.

Then we can update this PR if it is still needed, or close it if the new integration replaces these changes.

vitormattos:
I opened the follow issue to handle Unicode support in the XObject renderer:
- https://github.com/LibreSign/xobject-template/issues/33

I think this should be fixed there first. After that change is merged and released, we can update this PR to use the new support, or close it if these changes are no longer needed.

Issue #8145, opened 2026-08-31
TSA timestamping always fails with JSignPdf: --tsa-hash-algorithm is never passed, SHA-1 is used


bast69:
### Describe the bug

When a TSA is configured, signing with the **JSignPdf** engine always fails against any TSA that
rejects SHA-1 — which is every serious authority today.

`JSignPdfHandler::getTsaParameters()` passes only `--tsa-server-url`, `--tsa-policy-oid` and the
authentication options. It never passes **`--tsa-hash-algorithm`**, so JSignPdf falls back to its
default of **SHA-1** for the timestamp query. Certum and DigiCert answer `HTTP 400`.

Note that the admin setting *"Signature hash algorithm"* (`-ha`) does **not** help: it applies to
the document signature, not to the TSA query. Both parameters are independent.

Two things make this hard to diagnose:

1. **The error message blames the network.** `checkTsaError()` triggers on any error containing
   `TSAClientBouncyCastle`, `UnknownHostException` or `Invalid TSA`, and reports
   *"Timestamp Authority (TSA) service error. Check TSA endpoint and DNS/network/firewall
   connectivity from this server."* Connectivity is fine — the request is simply rejected.
2. **The real Java error is never logged.** `checkTsaError()` throws before the
   `$this->logger->error(...)` line in `signWrapper()`, so nothing about the actual cause reaches
   `nextcloud.log`. I only found it by running JSignPdf by hand.

### To reproduce

1. Configure a TSA that refuses SHA-1, e.g. `http://time.certum.pl`.
2. Keep the default signature engine (`JSignPdf`).
3. Sign any document → failure with the connectivity message above.

Connectivity is not the issue — a plain RFC 3161 exchange from inside the Nextcloud container
succeeds:

```
$ openssl ts -query -data f.txt -cert -sha256 -out f.tsq
$ curl -s -H "Content-Type: application/timestamp-query" --data-binary @f.tsq http://time.certum.pl -o f.tsr
$ openssl ts -reply -in f.tsr -text | grep Status
Status: Granted.
```

Running JSignPdf directly shows the real cause:

```
$ java -jar JSignPdf.jar test.pdf -kst PKCS12 -ksf k.p12 -ksp *** -ha SHA256 \
    --tsa-server-url http://time.certum.pl -d out
INFO Creating TSA client.
INFO Setting TSA hash algorithm: SHA-1
SEVERE Problem occured
[...]

PR #8256, opened 2026-09-05
feat: update JSignPdf to 3.1.0 with jsignpdf-php 3.0.0

### LINKED ISSUE #8068: Update JSignPdf to 3.x
LibreSign currently uses JSignPdf 2.3.0:

https://github.com/LibreSign/libresign/blob/main/lib/Service/Install/InstallService.php#L53

JSignPdf already has stable 3.x releases with important changes to its CLI, packaging and signing behavior.

Before updating LibreSign, the PHP wrapper must support JSignPdf 3.x:

https://github.com/JSignPdf/jsignpdf-php/issues/52

This issue should be worked on after that issue is completed and a new version of `jsignpdf-php` is available.

## What needs to be done

* Update LibreSign to the new `jsignpdf-php` version with JSignPdf 3.x support.
* Update the JSignPdf version, download source, checksum and installation logic in `InstallService`.
* Check the JSignPdf 3.x package structure and choose the correct package for LibreSign. JSignPdf 3.1 removed the old fat JAR and provides a minimal package for headless/CLI use, so changing only the version and checksum may not be enough.
* Review `JSignPdfHandler` and confirm that the current JSignPdf options still work correctly, including visible signatures, page and coordinates, hash algorithm, TSA, certification and output behavior.
* Update the unit tests for all behavior changed by this migration.
* Run the signing tests with JSignPdf 3.x and confirm that the existing LibreSign signing flows still work.

## JSignPdf options and arguments

Safe command building and shell escaping are mainly handled by the PHP wrapper and are part of:

https://github.com/JSignPdf/jsignpdf-php/issues/52

However, LibreSign also prepares options and values before sending them to the wrapper.

Review this integration and, when the new wrapper supports it, prefer passing structured values instead of preparing shell syntax inside LibreSign.

Check the dynamic values already used by LibreSign, such as:

* signature text;
* file paths;
* TSA URL;
* TSA policy OID;
* TSA username and password;
* page and coordinates;
* hash algorithm;
* certification level.

When several values follow the same logic, prefer PHPUnit data providers instead of many similar tests.

The data providers should include normal values and useful edge cases, such as spaces, quotes, shell special characters, empty values, numeric values, URLs and file paths.

The tests should confirm that LibreSign sends the expected values to the wrapper without creating unsafe or invalid command arguments.

## Testing requirements

The migration must include unit tests for the affected classes.

At minimum, review:

* `JSignPdfHandlerTest`;
* the tests for `InstallService`, if the installation logic changes.

If other classes are changed, update their tests too.

For `JSignPdfHandlerTest`, use data providers when this helps test different options and values without repeating the same test logic.

The tests should cover the relevant current behavior and any behavior changed by JSignPdf 3.x, including when applicable:

* visible and invisible signatures;
* page and coordinates;
* hash algorithms;
* TSA parameters;
* certification level;
* signature text;
* file paths;
* optional values;
* values with spaces and shell special characters.

Do not only change expected values to make old tests pass. Add or change tests when the JSignPdf 3.x behavior is different.

## Mutation testing

Run mutation testing for the classes directly affected by this migration.

The affected scope must reach **100% mutation score**, with no escaped mutants and no mutation test errors.

The tests should detect changes in the relevant JSignPdf integration logic, not only execute the code.

Keep the mutation test scope focused on the classes changed by this issue. There is no need to run mutation tests for unrelated parts of LibreSign.

If a mutant cannot be killed for a valid technical reason, explain the reason in the PR.

## Quality gates

Before the PR is ready for review, run the project checks using the scripts available in `composer.json`.

The affected PHPUnit tests and mutation tests must pass, and the PR must also pass:

* Psalm;
* PHPCS.

Do not ignore new warnings, errors or failed checks.

## Check new JSignPdf features

JSignPdf 3.x also added, or is adding, features that may be useful to LibreSign.

Review the JSignPdf release notes and CLI changes and add a short section to the PR description with useful features found during this work.

**Do not implement these features in this PR.**

Some examples to check are:

* new signing engines and DSS / PAdES support;
* changes to append and overwrite behavior;
* safer ways to provide passwords;
* new CLI options;
* support for existing PDF signature fields.

The support being developed in JSignPdf 3.2 for signing an existing PDF signature field is especially interesting for LibreSign.

In the future, this could allow LibreSign to receive a PDF with signature fields already defined and associate each field with a signer instead of creating the signature position.

If a new JSignPdf feature looks useful for LibreSign, create a separate follow-up issue for it.

## Expected result

After this change:

* LibreSign uses the latest supported stable JSignPdf 3.x release;
* LibreSign uses the updated `jsignpdf-php` wrapper;
* installation works with the new JSignPdf package structure;
* existing signing flows continue to work;
* options and values are correctly passed to the wrapper;
* affected behavior is covered by unit and mutation tests;
* Psalm and PHPCS pass;
* useful new JSignPdf features are listed for future work.

Keep this PR focused on the JSignPdf 3.x migration. New signing features should be handled in separate issues.

maia-andre:
Resolves: #8068

## 📝 Summary

Updates the JSignPdf signing backend to **JSignPdf 3.1.0** through **jsignpdf-php 3.0.0** (LibreSign/3rdparty#96). JSignPdf 3.1.0 also uses SHA-256 as the default TSA hash, which is what #8145 is about.

**Install.** JSignPdf 3.1 dropped the single `JSignPdf.jar`: the `minimal` package, meant for headless and command line use, ships a `lib/` directory that the wrapper starts through the classpath. `InstallService` now downloads `jsignpdf-3.1.0-minimal.zip`, verifies it against the `SHA256SUMS` file published with the release (instead of a hard-coded md5), and stores the extracted directory in the new `jsignpdf_path` app config key. The legacy `jsignpdf_jar_path` key is deleted on install and on uninstall. The release metadata lives in the new `JSignPdfRelease` class so it can be unit tested. `SignSetupService` and the setup check read the new key.

**Handler.** jsignpdf-php 3 escapes every argument itself and takes the options as a list, so `JSignPdfHandler` stops building shell syntax: the parameters are arrays (`addJSignParameters()`, one cloned `JSignParam` per visible element), the signature text is passed as is (no more quoting of `"` and `$`), the JVM option `-Duser.home` and the `JSIGNPDF_HOME` variable go through `setJavaOptions()` / `setEnvironmentVariables()`, and the TSA password goes through `setTsaPassword()`, which the wrapper writes to the stdin of JSignPdf instead of the command line. The wrapper now reports failure through the exit code instead of the "Finished: Signature succesfully created." message; the error parsing in `checkTsaError()` / `checkHashAlgorithmError()` is unchanged.

**Tests.** `JSignPdfHandlerTest` asserts the parameters that reach the wrapper (captured from `setParam()`), with the sign cases as structured providers, a case with shell special characters in the signature text, and a TSA provider covering no TSA, URL only, policy OID, basic authentication with the password over stdin, shell characters in the password, and the two cases with incomplete credentials. `JSignPdfReleaseTest` covers the download URL, checksum URL and install path. `JSignPdfSetupCheckTest` and `SignSetupServiceTest` follow the new key; the setup check gained two cases for mutants that escaped.

While doing this I found that 68 tests of `JSignPdfHandlerTest` were being skipped as "Certificate initialization failed": `generateRootCert()` needs a CA identifier to resolve the CRL distribution point and the class never generated one. `setUpBeforeClass()` now calls `CaIdentifierService::generateCaId()`, like `OpenSslHandlerTest` does, and all 96 tests run.

**Existing installations.** JSignPdf 2.3.0 cannot read passwords from stdin, so it does not work with the new wrapper: after upgrading, `occ libresign:install --jsignpdf` (or the install button in the admin settings) has to run once. Until then the setup check reports "JSignPdf not found" with that command as the tip, because `jsignpdf_path` is empty. I kept this explicit instead of downloading 35 MB during `occ upgrade`; say so if you prefer a repair step. The `-a` (append) default of the wrapper is a no-op in 3.x, where appending is the default, and LibreSign already rewrites the PDF header before signing, so the 3.x limitation that append mode cannot raise the PDF version for SHA-256 does not apply.

## 🧪 How to test

1. `occ libresign:install --use-local-cert --java` and `occ libresign:install --use-local-cert --jsignpdf`: the jsignpdf appdata folder gets `jsignpdf-3.1.0/lib/*.jar`, `occ config:app:get libresign jsignpdf_path` points to it and `jsignpdf_jar_path` is gone.
2. `occ libresign:configure:check` (or the admin overview) reports JSignPdf version 3.1.0.
3. Request a signature and sign it (clickToSign is enough). With a TSA configured with basic authentication, the password no longer appears in the JSignPdf command line (`ps` while signing, or the `--enable-stdin-passwords -ksp - -tsp -` in the command).
4. `composer run test:unit -- tests/php/Unit/Handler/SignEngine/JSignPdfHandlerTest.php`: 96 tests, none skipped.

## ⚙️ API / Back‑end changes

- [x] `InstallService`, `SignSetupService`, `JSignPdfSetupCheck`, `JSignPdfHandler`, new `JSignPdfRelease`; new app config key `jsignpdf_path` (replaces `jsignpdf_jar_path`)
- [x] Unit and/or integration tests added
- [ ] Capabilities updated (not applicable)
- [ ] Documentation updated (not applicable)
- [ ] API documentation updated with `composer openapi` (no API change)

### 🚧 Tasks
- [ ] Merge LibreSign/3rdparty#96 and point the submodule to its merge commit (the branch pins the head of that pull request for now).
- [ ] #8255 first: the 3rdparty `main` also carries pdf-signature-validator 0.5.1.

## ✅ Checklist

- [x] Validation (PHP 8.3 devcontainer):
  - PHPUnit: `JSignPdfHandlerTest` 107/107 (68 of them used to be skipped), `JSignPdfReleaseTest` 6/6, `JSignPdfSetupCheckTest` 11/11, `SignSetupServiceTest` 23/23; whole unit suite 3600 tests, the only 4 failures are in `AEngineHandlerTest` and pass when the class runs alone (state leaking between classes, unrelated to this change).
  - Infection on the lines changed against `main` (`--git-diff-lines`): 68 mutants, 68 killed, 100% coverage and 100% MSI. Per class: `JSignPdfRelease` 32/32, `JSignPdfSetupCheck` 33/33. `InstallService` has no unit coverage, as before; the install path is exercised by the behat and Playwright jobs (`libresign:install --jsignpdf`) and by the manual run below.
  - Psalm and php-cs-fixer clean on the changed files.
  - End to end: `occ libresign:install --use-local-cert --jsignpdf` downloaded `jsignpdf-3.1.0-minimal.zip`, verified it and set `jsignpdf_path` to `.../x86_64/jsignpdf/jsignpdf-3.1.0` (76 jars in `lib/`, legacy key gone); `java -classpath 'lib/*' com.intoolswetrust.jsignpdf.Bootstrap --version` prints `JSignPdf version 3.1.0`; a signature request signed with clickToSign returned 200 and the validation endpoint reports the signature as `adbe.pkcs7.detached`, RSA-SHA256, valid.

## 🤖 AI

Assisted by Claude Code (claude-fable-5-1); I reviewed and tested every change.

## JSignPdf 3.x features worth a look (not implemented here)

- **Pluggable signing engines** (`--list-engines`, `-eng`): OpenPDF stays the default; the new **EU DSS engine** produces PAdES signatures at the ETSI baseline levels B, T, LT and LTA, which OpenPDF cannot create.
- **Passwords over stdin** (`--enable-stdin-passwords`): used by this pull request through the wrapper.
- **`--overwrite`** replaces existing signatures instead of appending. LibreSign relies on appending, so nothing to do, but it is the switch to stay away from.
- **SHA-256 as the default hash, TSA included**: `--tsa-hash-algorithm` defaults to `AppConfig.defaultTsaHashAlg()` = SHA-256, which fixes the SHA-1 failure of #8145; a configurable TSA hash in the TSA policy stays as the follow-up discussed there.
- **3.2 (beta)**: sign into an existing signature field (`--sig-field`, `--list-sig-fields`), the feature the issue points at for pre-placed signature boxes; a `debug` output logging the certificate chain and every AIA/CRL/OCSP request, useful for TSA diagnostics; `buffering.mode=temp` to sign very large PDFs without a bigger heap.
- **Checksums for every artifact** (`SHA256SUMS.txt`): used by the installer here.

### DIFF
--- a/tests/php/Unit/SetupCheck/JSignPdfSetupCheckTest.php
+++ b/tests/php/Unit/SetupCheck/JSignPdfSetupCheckTest.php
@@ -14,6 +14,12 @@ if (!function_exists('OCA\Libresign\SetupCheck\file_exists')) {
 	}
 }
 
+if (!function_exists('OCA\Libresign\SetupCheck\is_dir')) {
+	function is_dir(string $filename): bool {
+		return \OCA\Libresign\Tests\Mock\FileSystemMock::fileExists($filename);
+	}
+}
+
 namespace OCA\Libresign\Tests\Unit\SetupCheck;
 
 use OCA\Libresign\Handler\SignEngine\JSignPdfHandler;
@@ -95,7 +101,7 @@ class JSignPdfSetupCheckTest extends TestCase {
 	public function testRunNoPathConfigured(): void {
 		$this->mockTranslation();
 		$this->appConfig->method('getValueString')
-			->with('libresign', 'jsignpdf_jar_path')
+			->with('libresign', 'jsignpdf_path')
 			->willReturn('');
 
 		$result = $this->check->run();
@@ -134,9 +140,9 @@ class JSignPdfSetupCheckTest extends TestCase {
 
 	public function testRunBinaryNotFound(): void {
 		$this->mockTranslation();
-		$jarPath = '/fake/path/jsignpdf.jar';
+		$jsignPdfPath = '/fake/path/jsignpdf-3.1.0';
 		$this->appConfig->method('getValueString')
-			->willReturn($jarPath);
+			->willReturn($jsignPdfPath);
 		$this->systemConfig->method('getSystemValueBool')
 			->willReturn(false);
 
@@ -144,29 +150,52 @@ class JSignPdfSetupCheckTest extends TestCase {
 		$this->signSetupService->method('verify')
 			->willReturn([]);
 
-		FileSystemMock::$files[$jarPath] = false;
+		FileSystemMock::$files[$jsignPdfPath] = false;
 
 		$result = $this->check->run();
 
 		$this->assertInstanceOf(SetupResult::class, $result);
 		$this->assertSame('error', $result->getSeverity());
-		$this->assertStringContainsString('JSignPdf file not found', $result->getDescription());
+		$this->assertStringContainsString('JSignPdf path not found', $result->getDescription());
 	}
 
 	public function testRunJavaNotFound(): void {
 		$this->mockTranslation();
-		$jarPath = '/fake/path/jsignpdf.jar';
+		$jsignPdfPath = '/fake/path/jsignpdf-3.1.0';
 		$this->appConfig->method('getValueString')
-			->willReturn($jarPath);
+			->willReturn($jsignPdfPath);
 		$this->systemConfig->method('getSystemValueBool')
 			->willReturn(false);
 		$this->signSetupService->method('verify')
 			->willReturn([]);
 
-		FileSystemMock::$files[$jarPath] = true;
+		FileSystemMock::$files[$jsignPdfPath] = true;
 
 		$this->javaHelper->method('getJavaPath')
 			->willReturn('');
+
+		$result = $this->check->run();
+
+		$this->assertInstanceOf(SetupResult::class, $result);
+		$this->assertSame('error', $result->getSeverity());
+		$this->assertStringContainsString('Necessary Java to run JSignPdf', $result->getDescription());
+	}
+
+	public function testRunJavaBinaryMissing(): void {
+		$this->mockTranslation();
+		$jsignPdfPath = '/fake/path/jsignpdf-3.1.0';
+		$this->appConfig->method('getValueString')
+			->willReturn($jsignPdfPath);
+		$this->systemConfig->method('getSystemValueBool')
+			->willReturn(false);
+		$this->signSetupService->method('verify')
+			->willReturn([]);
+
+		FileSystemMock::$files[$jsignPdfPath] = true;
+		FileSystemMock::$files['/opt/java/bin/java'] = false;
+
+		$this->javaHelper->method('getJavaPath')
+			->willReturn('/opt/java/bin/java');
 
 		$result = $this->check->run();
 
@@ -183,14 +212,14 @@ class JSignPdfSetupCheckTest extends TestCase {
 
 	public function testRunVersionEmpty(): void {
 		$this->mockTranslation();
-		$jarPath = '/fake/path/jsignpdf.jar';
+		$jsignPdfPath = '/fake/path/jsignpdf-3.1.0';
 		$this->appConfig->method('getValueString')
-			->willReturn($jarPath);
+			->willReturn($jsignPdfPath);
 		$this->systemConfig->method('getSystemValueBool')
 			->willReturn(false);
 		$this->signSetupService->method('verify')->willReturn([]);
 		$this->javaHelper->method('getJavaPath')->willReturn('/usr/bin/java');
-		FileSystemMock::$files[$jarPath] = true;
+		FileSystemMock::$files[$jsignPdfPath] = true;
 		FileSystemMock::$files['/usr/bin/java'] = true;
 
 		$jsignPdfMock = $this->getMockBuilder(\OCA\Libresign\Vendor\Jeidison\JSignPDF\JSignPDF::class)
@@ -200,6 +229,7 @@ class JSignPdfSetupCheckTest extends TestCase {
 		$jsignPdfMock->method('getVersion')->willReturn('');
 
 		$jsignParamMock = $this->createJSignParamMock();
+		$jsignPdfMock->expects($this->once())->method('setParam')->with($jsignParamMock);
 
 		$this->jSignPdfHandler->method('getJSignPdf')->willReturn($jsignPdfMock);
 		$this->jSignPdfHandler->method('getJSignParam')->willReturn($jsignParamMock);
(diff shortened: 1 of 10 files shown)


tests/php/Unit/SetupCheck/JSignPdfSetupCheckTest.php
```diff
@@ -183,14 +212,14 @@ private function createJSignParamMock() {
 
 	public function testRunVersionEmpty(): void {
 		$this->mockTranslation();
-		$jarPath = '/fake/path/jsignpdf.jar';
+		$jsignPdfPath = '/fake/path/jsignpdf-3.1.0';
```
vitormattos: I think we should avoid repeating `3.1.0` in the tests.

We already have `JSignPdfRelease::VERSION` as the source of the JSignPdf version, but some tests still use `3.1.0` directly in paths and URLs.

I think one test should keep:

```php
$this->assertSame('3.1.0', JSignPdfRelease::VERSION);
```

This makes sure that LibreSign uses the expected version.

But the other tests can use `JSignPdfRelease::VERSION` when they only need to build a path, file name, or URL.

For example:

```php
'.../jsignpdf/jsignpdf-' . JSignPdfRelease::VERSION
```

This will make the next JSignPdf update easier and avoid changing the same version in many test files.

vitormattos (changes requested):
Excellent work!
I think that only have a small improvement and small changes to make psalm happy.

vitormattos:
@maia-andre this PR is blocking the CI because I already made the merge at 3rdparty. I'll need to go ahead with the small changes to merge this. After the merge, will be pending the backport flow.

Issue #7832, opened 2026-06-25
Configurable signature rejection workflow


vitormattos:
Currently, LibreSign allows a signer to digitally sign a document, but it is not possible to explicitly reject a signing request.

Many organizations require signers to be able to reject a signing request, optionally providing a justification, while allowing administrators to define how that rejection affects the signing workflow and which information becomes visible to other participants.

This issue proposes a configurable Signature Rejection Workflow that is disabled by default, preserving the current LibreSign behavior.

The initial implementation should use a global configuration. However, the architecture should be designed to allow future integration with Policies & Rules, enabling different organizations, groups or users to have different rejection behaviors.

> [!NOTE]
>
> This epic defines the goal, architecture and scope.
> **Pull requests should implement focused child issues, not the epic itself.**

> [!IMPORTANT]
>
> ## Funding and community support needed
>
> Signature rejection is important for organizations where signing is part of an approval, compliance or document workflow.
>
> **We are looking for organizations interested in funding this work and for contributors who want to help build it.**
>
> Organizations that need configurable rejection, audit, privacy or approval workflows are especially welcome to support this work.
>
> Community contributions are also welcome through the focused child issues of this epic.

---

# Goals

* Preserve current LibreSign behavior by default.
* Allow signers to explicitly reject a signing request.
* Allow administrators to define optional or mandatory rejection comments.
* Respect privacy and compliance requirements by allowing administrators to configure the visibility of rejection information.
* Keep the implementation compatible with future Policies & Rules.

---

# Functional Requirements

## FR1 - Enable Signature Rejection

A new global setting should enable or disable the Signature Rejection Workflow.

Default:

* Disabled

When disabled:

* LibreSign behaves exactly as today.
* Only the Sign button is displayed.
* No additional workflow is introduced.

When enabled:

Will display the follow options:

☑ Allow rejection comments

> ☑ Require rejection comments
> ☑ Allow who reject to make the comment private

☑ Cancel workflow after rejection

☑ Make public the rejection status at who rejected on validation page. If disabled, the validation page will be disabled.

> ☑ Show rejection comment on validation page

---

## FR2 - Signing Experience

When the feature is enabled, the signing page should display two primary actions:

* Sign (success [[button](https://nextcloud-vue-components.netlify.app/#/Components/NcButton)](https://nextcloud-vue-components.netlify.app/#/Components/NcButton))
[...]

PR #8257, opened 2026-09-05
feat: add backend support for signature rejection

### LINKED ISSUE #8159: Backend support for signature rejection
Implement the backend support required by #7832.

This issue should provide the backend foundation for the signature rejection workflow. Frontend configuration, signing UI, and validation page will be handled in separate issues.

## Scope

Implement the backend support to:

* allow a signer to reject a signature request;
* store the rejection state and timestamp;
* store the rejection comment when provided;
* support the rejection rules defined in the policy system;
* define the effective behavior when a signer rejects, including whether the remaining signing workflow can continue;
* expose the required information through the API for the signing flow, policy configuration UI, and validation page;
* prevent actions that are no longer valid after the workflow is closed;
* register the rejection in the audit trail;
* dispatch the appropriate LibreSign events and Activities.

The implementation must use the existing Policies & Rules architecture. Policy resolution, scopes, inheritance, delegation, and overrides should follow the same patterns already used by other LibreSign policies.

Avoid introducing temporary global app configuration that would later need to be migrated to Policies & Rules.

## Tests

Add or update tests covering at least:

* policy resolution and default behavior;
* rejection of a signature request;
* optional and required rejection comments;
* workflow continuation and cancellation behavior;
* multiple signers;
* API validation and permissions;
* persistence and retrieval of rejection information;
* actions that must be blocked after workflow cancellation;
* audit trail and relevant events.

Use unit tests for isolated business logic and Behat/integration tests for the relevant end-to-end backend flows.

Use data providers where they reduce duplicated scenarios.

## Mutation testing

New or changed backend business logic related to this feature must be covered by Infection.

For reasonably isolated new classes:

* no escaped mutants;
* no mutation errors;
* target 100% MSI.

Do not weaken Infection configuration or exclude production code only to make the tests pass.

## Quality gates

The implementation must pass the existing backend checks, including:

* PHPUnit;
* Behat/integration tests;
* Psalm;
* PHPCS;
* Infection.

## Out of scope

* the Policy Workbench UI;
* the signing/rejection frontend flow;
* the validation page UI.

These will be handled in follow-up issues linked to #7832.

Sufiyan-MSA:
Resolves :  #8159

## Summary

Adds the backend foundation for the signature rejection workflow described in #7832: a signer
can decline to sign, the refusal is persisted with its timestamp and optional comment, and the
rules governing it come from a new `signature_rejection` policy in the existing Policies & Rules
architecture. No temporary global app configuration was introduced.

Disabled by default — with the default policy value, behavior is identical to today.

## How to test

1. As admin, enable the policy:
   `POST /ocs/v2.php/apps/libresign/api/v1/policies/system/signature_rejection`
   with `value = {"enabled":true,"comment_mode":"required","cancel_workflow":true}`
2. Create a signature request with two signers.
3. As the first signer, `POST /ocs/v2.php/apps/libresign/api/v1/sign/file_id/<FILE_ID>/reject`
   with no `comment` → 422, "A comment is required to reject this signature request."
4. Retry with `comment=I do not agree` → 200, `status: 3`, `workflowCanceled: true`.
5. As the second signer, try to sign → refused; the file is now status 6 (Canceled).
6. Repeat with `cancel_workflow=false` and confirm the second signer can still sign.
7. Check the requester's Activity stream for the rejection entry.

Automated coverage: `tests/integration/features/sign/reject.feature`.

##  API / Back-end changes

- New `signature_rejection` policy provider (system + group scope, group-admin delegation).
  Sub-rules: `comment_mode` (disabled/optional/required), `cancel_workflow`,
  `public_status`, `show_comment_on_validation`. A delegated rule cannot downgrade
  a comment the parent marked as required. Comment privacy is chosen by the signer
  and is not controlled by policy.
- Rules are frozen per document through a `FilePolicy` applier writing to
  `metadata.policy_snapshot`, so a later policy change never alters an in-flight workflow.
- `SignRequestStatus::REJECTED` and `FileStatus::CANCELED`; migration
  `Version18004Date20260906000000` adds `rejected_at`, `rejection_comment` and
  `rejection_comment_private` to `libresign_sign_request`.
- `SignatureRejectionService` enforces the rules, persists the rejection, closes the workflow
  when configured, and dispatches `SignatureRejectedEvent`.
- Endpoints `POST /api/v1/sign/file_id/{fileId}/reject` and
  `POST /api/v1/sign/uuid/{uuid}/reject`.
- Rejection data is exposed on the signer payload through
  `SignatureRejectionVisibilityService`: the requester and the rejecting signer always see the
  full record; other readers see only what the policy makes public, and a comment marked
  private is never disclosed to them.
- Actions after closure are blocked by the existing `fileCanBeSigned()` status gate; a second
  rejection and rejection after signing are refused explicitly.
- `hasPendingLowerOrderSigners()` now treats a rejected signer as no longer pending, so
  ordered-numeric flows can continue past a rejection when the policy allows it.
- Rejections are registered in the audit trail via a new Activity provider and setting.


lib/Service/Policy/Provider/SignatureRejection/FilePolicy/SignatureRejectionFilePolicyApplier.php
```diff
@@ -0,0 +1,61 @@
...
+
+	#[\Override]
+	public function apply(FileEntity $file, array $data): void {
+		$user = ($data['userManager'] ?? null) instanceof IUser ? $data['userManager'] : null;
+		$activeContext = $this->extractActiveContext($data);
+		$resolvedPolicy = $activeContext === null
+			? $this->policyService->resolveForUser(SignatureRejectionPolicy::KEY, $user)
+			: $this->policyService->resolveForUser(SignatureRejectionPolicy::KEY, $user, [], $activeContext);
+		$this->storeSignatureRejectionPolicySnapshot($file, $resolvedPolicy);
+	}
```
vitormattos, line 33: I think the rejection policy should control whether the requester is allowed to enable rejection for a signature request.

When the policy is enabled, the requester should be able to choose if rejection is enabled for that file or envelope. This choice must be stored with the request and become the effective value for the signing flow.

When the policy is disabled, the requester must not be able to enable rejection, including through a direct API request.

The requester should still be able to change this choice while the signing flow has not started. Once the flow starts, the stored choice must stay frozen. A later policy change must also not change the behavior of an existing request.

Could we follow the same request override and snapshot lifecycle used in #8143, and add integration coverage for normal files and envelopes?

lib/Service/SignatureRejection/SignatureRejectionService.php
```diff
@@ -0,0 +1,180 @@
...
+		$workflowCanceled = $policy['cancel_workflow'];
+		if ($workflowCanceled) {
+			$libreSignFile->setStatusEnum(FileStatus::CANCELED);
+			$this->fileStatusService->update($libreSignFile);
+		}
+
+		$this->dispatchRejectedEvent($signRequest, $libreSignFile, $workflowCanceled);
+
+		return $signRequest;
+	}
```
vitormattos, line 89: Could we make the rejection and workflow cancellation one atomic state change when `cancel_workflow` is enabled?

The sign request is saved as `REJECTED` before the file is saved as `CANCELED`. If the second update fails, the signer is already rejected but the document can still remain open for other signers.

These state changes should succeed or fail together.

Please also add a test where saving the canceled file fails and confirm that we do not leave a partially updated workflow.

tests/integration/features/sign/reject.feature
```diff
@@ -0,0 +1,109 @@
...
+    When as user "signer2"
+    And sending "post" to ocs "/apps/libresign/api/v1/sign/file_id/<FILE_ID>"
+      | method | password |
+      | token  | password |
+    Then the response should have a status code 422
+    When as user "admin"
+    And sending "get" to ocs "/apps/libresign/api/v1/file/list?details=1"
+    Then the response should be a JSON array with the following mandatory values
+      | key                          | value |
+      | (jq).ocs.data.data[0].status | 6     |
```
vitormattos, line 109: Could we extend the integration coverage for the rejection API?

The current scenarios use the authenticated `file_id` endpoint, but the new `uuid` endpoint is also public and uses `RequireSigner`. We should test this path end to end, including a successful rejection by the expected signer and a request where the current identity must not be allowed to reject.

The tests should also confirm that the endpoint respects the rejection setting stored for the request, instead of only the current global policy.

For a canceled workflow, this scenario should verify other backend actions that are no longer valid, such as sending a reminder or modifying/removing pending signers, not only signing again.

These checks must be enforced by the backend, even if the frontend hides the actions.

vitormattos (changes requested):

lib/Service/Policy/Provider/SignatureRejection/FilePolicy/SignatureRejectionFilePolicyApplier.php
vitormattos: Thanks, this is much closer, but there are still two cases to fix.

First, if the policy has `enabled: true` and the request does not send a rejection choice, the current code enables rejection for that request automatically.

This is not the behavior we want. The policy should only allow the requester to enable rejection. If the requester does not choose it, rejection should stay disabled for that request.

Second, if a draft already has rejection disabled and later the requester edits something else without sending the rejection setting again, the current code can enable rejection because it resolves the policy again.

Please keep the stored request value when no new rejection value is sent.

Before the signing flow starts, the requester should still be able to change the rejection setting explicitly.

Please add integration tests for these cases, including an envelope:

* policy enabled + no request choice → rejection stays disabled;
* draft with rejection disabled + unrelated update → rejection stays disabled;
* requester can explicitly change the setting before the signing flow starts;
* after the signing flow starts, the stored value cannot change.

vitormattos: The code now looks correct for the cases we discussed. Thanks.

I think we only need to complete the integration coverage for envelopes before resolving this thread.

We already test that an enabled policy alone does not enable rejection for an envelope, but the request lifecycle tests below only cover a normal file.

Could we add an envelope scenario that confirms the stored rejection choice is preserved after the request is created, including after a later policy change or an unrelated draft update?

This is important because file and envelope requests can follow different backend paths.

vitormattos (changes requested):
Nice work on this PR. The backend flow is much clearer now, especially the request policy behavior, envelope support, and the new integration tests.

While reviewing the rejection comment, I thought about one more case that may be important.

I think the rejection status and the rejection comment should be handled separately.

The rejection itself may need to be visible to other people in the signing flow because it changes the workflow. But the comment can contain personal, legal, or sensitive information.

Because of this, I think the signer should always be able to mark the rejection comment as private.

In this case, the comment should only be visible to:

* the signer who rejected;
* the person who requested the signature.

The rejection status can still follow the normal visibility rules.

Today we have `allow_private_comment` in the policy. I am not sure the administrator should be able to remove this option from the signer.

Maybe the policy should only control if comments are disabled, optional, or required. If comments are allowed, the signer could always choose if their own comment is private.

By default, the comment can still be public unless the signer sends `privateComment: true`.

I think we can add this in this PR because the API and policy model are still being defined. But I am also fine with a follow-up PR if we want to keep this PR smaller.

In both cases, I think we should add a Behat integration test for this behavior.

The test should confirm that:

* the requester can see the private comment;
* the signer who rejected can see their own private comment;
* another signer can see the rejection status when allowed, but cannot see the private comment.

This will help us avoid exposing a private comment by mistake in the real API response.

vitormattos:
Some backend checks are failing after running the CI.

The failures look related to this PR. In particular, the policy snapshot validation does not accept `signature_rejection` yet, some existing `FilePolicyApplier` tests/mocks need to be updated for the new rejection applier, and the generated OpenAPI files are not up to date.

Could you please fix the failing checks and make the CI green again?

vitormattos:
Don't worry about make more commits, is more easy to make the code review with incremental commits.
The CI follow with issues, look the result of unit tests.

tests/integration/features/sign/reject.feature
```diff
@@ -0,0 +1,691 @@
...
+    And the response should be a JSON array with the following mandatory values
+      | key                            | value |
+      | (jq).ocs.data.workflowCanceled | true  |
+    When sending "post" to ocs "/apps/libresign/api/v1/sign/file_id/<FILE_ID>/reject"
+    Then the response should have a status code 422
+    And the response should be a JSON array with the following mandatory values
+      | key                   | value                                                    |
+      | (jq).ocs.data.message | The signing workflow of this document is already closed. |
+    When as user "signer2"
+    And sending "post" to ocs "/apps/libresign/api/v1/sign/file_id/<FILE_ID>"
```
vitormattos: This test is using the manager endpoint for `signer2`.

`POST /sign/file_id/{fileId}` is protected by `RequireManager`, so the request is rejected before we can test the closed workflow behavior.

Could we fetch the signer request UUID and use `POST /sign/uuid/<SIGN_REQUEST_UUID>` instead?

I would also prefer `clickToSign` here instead of `password`, because this scenario only needs to confirm that the second signer cannot sign after the workflow was canceled.

vitormattos (changes requested):

lib/Service/Policy/Provider/SignatureRejection/FilePolicy/SignatureRejectionFilePolicyApplier.php
```diff
@@ -0,0 +1,200 @@
...
+
+	/**
+	 * Once the flow starts the stored value is frozen, so the setting may no longer
+	 * be sent at all. Refusing it outright, instead of only refusing a different
+	 * value, keeps a single document and an envelope behaving the same way: the
+	 * value of an envelope lives on the documents it contains and is not reachable
+	 * from here, so a "same value" comparison is not possible for one of them.
+	 */
+	private function assertValueIsNotChangedAfterTheFlowStarted(?bool $requestedChoice): void {
```
vitormattos: Could we allow the same stored value to be sent again after the signing flow starts?

I agree that the setting must be frozen, but I think frozen should mean that its value cannot change. It should not mean that the field cannot appear in the request anymore.

A client may send the complete form state in an unrelated update. If rejection is already enabled and the request sends `enabled: true` again, this should not make the whole update fail.

The current test even expects `resending the value it already has` to fail.

Could we reject only when the requested value is different from the stored value? If envelopes make this hard to compare, I think we should find a way to read their effective stored value instead of making the API behave differently from an idempotent update.

vitormattos (changes requested):

lib/Service/Policy/AbstractFilePolicyApplier.php
```diff
@@ -20,6 +21,7 @@ public function __construct(
 		protected readonly PolicyService $policyService,
 		protected readonly FileService $fileService,
 		protected readonly ?IL10N $l10n = null,
+		protected readonly ?FileMapper $fileMapper = null,
```
vitormattos: Do we need `FileMapper` in `AbstractFilePolicyApplier`?

It looks like only `SignatureRejectionFilePolicyApplier` needs it, to resolve the frozen value for envelopes. With the current approach, every file policy applier receives this dependency even when it does not use it.

Would it make sense for `SignatureRejectionFilePolicyApplier` to receive `SignatureRejectionPolicyService` instead, and keep `FileMapper` inside that service? Then `frozenChoice()` could use the service directly, without adding `FileMapper` to the base applier.

If this needs a larger change in how appliers are created, I think the current approach is acceptable for this PR.

Sufiyan-MSA: Done, FileMapper is out of the base applier. SignatureRejectionFilePolicyApplier declares its own constructor and wraps the mapper in SignatureRejectionPolicyService, which frozenChoice() uses directly. The discoverer is unchanged; it passes the mapper as an optional trailing argument that appliers without a use for it simply ignore.

Sufiyan-MSA:
Hi [@vitormattos](https://github.com/vitormattos), I've pushed the refactor to keep the FileMapper out of the base applier. Whenever you have a moment, could you please take another look? Thanks!

vitormattos (approved):

PR #8275, opened 2026-09-06
refactor(jsignpdf): move the hash algorithm resolution to a dedicated class

maia-andre:
Resolves: # 

## 📝 Summary

Item 3 of the scope described in #8145: the hash algorithm resolution leaves `JSignPdfHandler` and
becomes a small dedicated class with isolated tests.

The algorithm that signs the document depends on the PDF version — JSignPdf still needs SHA-1 below
PDF 1.6 and refuses it from 1.7 on — so its result describes the signature only. Reusing it for the
other hashes JSignPdf takes would send SHA-1 to a timestamp authority every time the document is
older than PDF 1.6. While these rules lived in the handler as private methods there was no place
where that distinction could be stated or tested.

`HashAlgorithmResolver` now owns them, one entry point per hash:

- `forSignature(?float $pdfVersion)`: the version-aware algorithm that goes to `--hash-algorithm`.
- `requiresPdfVersionUpgradeForSha256(float $pdfVersion)`: whether the document has to be rewritten
  as PDF 1.6 before the configured SHA-256 is accepted.

`extractPdfVersion()` stays in the handler: reading the document is the handler's job, not the
policy's.

No behavior change — same version thresholds, same fallback to SHA256, same list of supported
algorithms, same parameters reaching JSignPdf. The TSA hash algorithm as a TSA policy option (items
1 and 2 of the issue) comes in a following PR, and is the reason each hash gets its own entry point
here.

## 🧪 How to test

```bash
vendor/bin/phpunit -c tests/php/phpunit.xml --no-coverage tests/php/Unit/Handler/SignEngine/
```

`HashAlgorithmResolverTest` covers the resolution in isolation (no Nextcloud server, only a mocked
`PolicyService`): every PDF version threshold, the unsupported/empty/unset configured algorithm, and
the PDF version upgrade rule. `JSignPdfHandlerTest` keeps proving, end to end, which
`--hash-algorithm` reaches JSignPdf for each document.

## ⚙️ API / Back‑end changes

- [x] `HashAlgorithmResolver` (new) resolves the signature hash algorithm and the PDF version upgrade rule
- [x] `JSignPdfHandler` receives it by dependency injection and no longer resolves the hash policy itself

No API, capability or documentation change: the class is internal and autowired.

## ✅ Checklist

- [x] Full unit suite green, except the 4 known `AEngineHandlerTest::testSetEngineConfiguresIdentifyMethodsForNoneEngine`
  failures: that class passes 26/26 when it runs alone, so they come from state left by other test
  classes, not from this change (it does not touch the certificate engine)
- [x] Infection on the changed lines: 19/19 mutants killed (100% MSI, 100% mutation code coverage)
- [x] Psalm and php-cs-fixer clean on the changed files

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Handler/SignEngine/JSignPdf/HashAlgorithmResolver.php
+++ b/lib/Handler/SignEngine/JSignPdf/HashAlgorithmResolver.php
@@ -6,7 +6,7 @@ declare(strict_types=1);
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-namespace OCA\Libresign\Handler\SignEngine;
+namespace OCA\Libresign\Handler\SignEngine\JSignPdf;
 
 use OCA\Libresign\Service\Policy\PolicyService;
 use OCA\Libresign\Service\Policy\Provider\SignatureHashAlgorithm\SignatureHashAlgorithmPolicy;
--- a/lib/Handler/SignEngine/HashAlgorithmResolver.php
+++ b/lib/Handler/SignEngine/HashAlgorithmResolver.php
@@ -1,0 +1,86 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * SPDX-FileCopyrightText: 2026 LibreCode coop and contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Libresign\Handler\SignEngine;
+
+use OCA\Libresign\Service\Policy\PolicyService;
+use OCA\Libresign\Service\Policy\Provider\SignatureHashAlgorithm\SignatureHashAlgorithmPolicy;
+
+/**
+ * Resolves which hash algorithm JSignPdf has to use.
+ *
+ * The algorithm that signs the document depends on the PDF version, so it is
+ * not a valid answer for the other hashes JSignPdf takes: each one needs its
+ * own method here.
+ */
+class HashAlgorithmResolver {
+	private const float MIN_PDF_VERSION_SHA256 = 1.6;
+	private const float MIN_PDF_VERSION_SHA1_REJECT = 1.7;
+	private const string DEFAULT_ALGORITHM = 'SHA256';
+	/** @var string[] */
+	private const array SUPPORTED_ALGORITHMS = ['SHA1', 'SHA256', 'SHA384', 'SHA512', 'RIPEMD160'];
+
+	public function __construct(
+		private PolicyService $policyService,
+	) {
+	}
+
+	/**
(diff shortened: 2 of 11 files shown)


lib/Handler/SignEngine/JSignPdf/HashAlgorithmResolver.php
vitormattos, line 1: Since this resolver contains JSignPdf-specific rules, would it make sense to start grouping JSignPdf classes in their own namespace/directory instead?

For example, we could have `SignEngine/JSignPdf/JSignPdfHandler` and `SignEngine/JSignPdf/HashAlgorithmResolver`.

This would keep the resolver name simple while making its scope clear, and could also help if we add more JSignPdf-specific classes while implementing the TSA changes.

vitormattos (changes requested):

lib/Handler/SignEngine/JSignPdf/HashAlgorithmResolver.php
maia-andre: Thanks for the direction — that was exactly the doubt I had when placing the class, and grouping them is the better answer.

I will move both to `SignEngine/JSignPdf/` (`JSignPdfHandler` and `HashAlgorithmResolver`), keeping the class names, and push it here as a separate commit so the move stays easy to read next to the extraction.

About the ordering: #8274 also touches `JSignPdfHandler.php` and is already green, so merging it before this move avoids a conflict on the same file — otherwise I rebase it right after. #8276 is stacked on this branch and I rebase it as soon as this one is updated.

maia-andre: Done in 09233c877: `JSignPdfHandler` and `HashAlgorithmResolver` (with their tests) now live in `SignEngine/JSignPdf/`, class names unchanged.

The move surfaced something worth a look: `Pkcs12Handler::getHandler()` was building the class name at runtime — `'OCA\\Libresign\\Handler\\SignEngine\\' . ucfirst($property)` — so the namespace change compiled, passed static analysis and only broke when a document was actually signed (`Could not resolve ...\SignEngine\JSignPdfHandler`, caught by `Pkcs12HandlerTest`). It now resolves through an explicit `ENGINE_HANDLERS` map, which also means an engine name that happens to match another property (`docMdpHandler`, `footerHandler`) answers the friendly "Invalid Sign engine." error instead of reaching the container.

Full unit suite green (3629 tests, apart from the 4 known `AEngineHandlerTest` failures that also happen without this branch), psalm and php-cs-fixer clean. #8276 is already rebased on top.

PR #8276, opened 2026-09-06
feat(tsa): let administrators choose the hash algorithm of the timestamp query

### LINKED ISSUE #8145: TSA timestamping always fails with JSignPdf: --tsa-hash-algorithm is never passed, SHA-1 is used

maia-andre:
Resolves: #8145

## 📝 Summary

The last of the three parts of #8145: the hash algorithm of the timestamp query becomes an option of
the TSA policy, next to the URL, the policy OID and the authentication.

Until now nobody could choose it. JSignPdf picked its own default — SHA-1 up to 2.3.0, SHA-256 since
3.1.0 — and the authorities in the report answer `HTTP 400` when that choice is not the one they
expect. The administrator who configures the authority is the one who knows what it accepts, and in
a multi-tenant instance each TSA rule now carries its own answer.

- **Policy**: `hash_algorithm` inside `tsa_settings`, default `SHA256`. `SHA1` is not offered: it is
  the algorithm the authorities in the report refuse.
- **JSignPdf**: `--tsa-hash-algorithm` is sent with the configured value, spelled the way JSignPdf
  documents it (`SHA-256`, with the hyphen, unlike `--hash-algorithm`).
- **PHP native engine**: reads the same policy instead of being pinned to SHA-256, so the option
  does not silently do nothing when that engine signs.
- The timestamp hash is resolved on its own and never from the signature one, which follows the PDF
  version and returns SHA1 for documents older than PDF 1.6 — exactly the request #8145 reports as
  rejected.

Existing instances are not affected: the policy value is decoded with `SHA256` when the key is
absent, which is what JSignPdf 3.1.0 already does.

### 🚧 Tasks

- [x] Merge #8275 first (this branch carries its commit; the diff here shrinks to the last three
  commits once it lands). #8274 is independent and can go in any order.

## 🧪 How to test

```bash
vendor/bin/phpunit -c tests/php/phpunit.xml --no-coverage --testsuite unit
npm run test
```

In the interface: *Administration settings → LibreSign → Timestamp Authority → Change*, turn the
timestamp server on and pick an algorithm. `occ config:app:get libresign tsa_settings` shows the
stored value.

### Manual verification (Nextcloud 35 devcontainer, LibreSign 16.0.0-dev)

1. Opened the TSA policy editor before and after the change — the select appears between the policy
   OID and the authentication.
2. Chose SHA-512 and saved.
3. Reloaded the page and reopened the rule: the value is still SHA-512.
4. `occ config:app:get libresign tsa_settings`:

```json
{"url":"https://freetsa.org/tsr","policy_oid":"","auth_type":"none","username":"","hash_algorithm":"SHA512"}
```

## 🎨 UI / Front‑end changes

- [x] New `TSA hash algorithm` select in the TSA rule editor (SHA-256, SHA-384, SHA-512)
- [x] Components, Unit (with vitest) and/or e2e (with Playwright) tests added - *Required*
- [ ] Accessibility verified – the select is the same `NcSelect` already used for the authentication
  type, with its own label

🏚️ Before | 🏡 After
--- | ---
 | 

Algorithms offered | Saved and reopened
--- | ---
 | 

## ⚙️ API / Back‑end changes

- [x] `TsaPolicyValue` gains `hash_algorithm` (default `SHA256`, validated against SHA256/SHA384/SHA512)
- [x] `HashAlgorithmResolver::forTsa()` answers the timestamp hash from the TSA policy, never from the signature one
- [x] `JSignPdfHandler` sends `--tsa-hash-algorithm`; `PhpNativeHandler` passes the same value to the signer library

No API or capability change: the policy value keeps the same shape, with one more key.

## ✅ Checklist

- [x] PHP unit suite: 3659 tests, only the 4 known `AEngineHandlerTest::testSetEngineConfiguresIdentifyMethodsForNoneEngine`
  failures, which come from state left by other test classes (that class passes 26/26 alone)
- [x] Frontend: 3118 vitest tests green, `vue-tsc --noEmit` clean
- [x] Infection on the changed lines: 19/19 mutants killed (100% MSI, 100% mutation code coverage)
- [x] Psalm and php-cs-fixer clean on the changed files

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Handler/SignEngine/HashAlgorithmResolver.php
+++ b/lib/Handler/SignEngine/HashAlgorithmResolver.php
@@ -1,0 +1,86 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * SPDX-FileCopyrightText: 2026 LibreCode coop and contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Libresign\Handler\SignEngine;
+
+use OCA\Libresign\Service\Policy\PolicyService;
+use OCA\Libresign\Service\Policy\Provider\SignatureHashAlgorithm\SignatureHashAlgorithmPolicy;
+
+/**
+ * Resolves which hash algorithm JSignPdf has to use.
+ *
+ * The algorithm that signs the document depends on the PDF version, so it is
+ * not a valid answer for the other hashes JSignPdf takes: each one needs its
+ * own method here.
+ */
+class HashAlgorithmResolver {
+	private const float MIN_PDF_VERSION_SHA256 = 1.6;
+	private const float MIN_PDF_VERSION_SHA1_REJECT = 1.7;
+	private const string DEFAULT_ALGORITHM = 'SHA256';
+	/** @var string[] */
+	private const array SUPPORTED_ALGORITHMS = ['SHA1', 'SHA256', 'SHA384', 'SHA512', 'RIPEMD160'];
+
+	public function __construct(
+		private PolicyService $policyService,
+	) {
+	}
+
+	/**
+	 * Algorithm used to sign a document with the given PDF version.
+	 *
+	 * @param float|null $pdfVersion null when the content carries no readable PDF header
+	 */
+	public function forSignature(?float $pdfVersion): string {
+		$configuredAlgorithm = $this->getConfiguredAlgorithm();
+		/**
+		 * Need to respect the follow code:
+		 * https://github.com/intoolswetrust/jsignpdf/blob/JSignPdf_2_2_2/jsignpdf/src/main/java/net/sf/jsignpdf/types/HashAlgorithm.java#L46-L47
+		 */
+		if ($pdfVersion === null) {
+			return $this->validate($configuredAlgorithm);
+		}
+
+		return $this->forPdfVersion($pdfVersion, $configuredAlgorithm);
+	}
+
+	/**
+	 * PDFs older than 1.6 have to be upgraded before JSignPdf accepts SHA-256.
+	 */
+	public function requiresPdfVersionUpgradeForSha256(float $pdfVersion): bool {
+		if ($pdfVersion >= self::MIN_PDF_VERSION_SHA256) {
+			return false;
+		}
+
+		return $this->getConfiguredAlgorithm() === self::DEFAULT_ALGORITHM;
+	}
+
+	private function forPdfVersion(float $pdfVersion, string $configuredAlgorithm): string {
+		// Legacy compatibility: JSignPdf still requires SHA1 for very old PDFs (< 1.6).
+		// The policy still exposes SHA1 for supported legacy workflows, and the runtime
+		// must continue enforcing this fallback for ancient PDFs that JSignPdf cannot sign otherwise.
+		if ($pdfVersion < self::MIN_PDF_VERSION_SHA256) {
+			return 'SHA1';
+		}
+		if ($pdfVersion < self::MIN_PDF_VERSION_SHA1_REJECT) {
+			return self::DEFAULT_ALGORITHM;
+		}
+		if ($configuredAlgorithm === 'SHA1') {
+			return self::DEFAULT_ALGORITHM;
+		}
+
+		return $this->validate($configuredAlgorithm);
+	}
+
+	private function validate(string $algorithm): string {
+		return in_array($algorithm, self::SUPPORTED_ALGORITHMS, true) ? $algorithm : self::DEFAULT_ALGORITHM;
+	}
+
+	private function getConfiguredAlgorithm(): string {
+		return (string)$this->policyService->resolve(SignatureHashAlgorithmPolicy::KEY)->getEffectiveValue();
+	}
+}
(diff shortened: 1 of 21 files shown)


maia-andre:
Rebased on the updated #8275: the JSignPdf classes moved to `SignEngine/JSignPdf/` there, so this branch now carries that commit too. Nothing changed in what this PR itself does.

vitormattos (approved):

Issue #8326, opened 2026-09-09
Show visible signature information in signer validation details


vitormattos:
## Context

Part of #8321.

A PDF digital signature can exist without a visible signature mark on the page.

LibreSign should make this clear also on the validation page.

This information should be shown as part of the signer details. It is metadata about how the signature is represented in the PDF, not a validation warning.

## Expected behavior

In the signer details on the validation page, show whether that signer has a visible signature element.

Suggested field:

> **Visible signature:** Yes

or:

> **Visible signature:** No

Use the existing signer details layout.

Do not use a warning, error or success message for this information.

A signature without a visible element is still a valid digital signature.

## Where to implement

The document validation flow already renders signer information through:

`src/components/validation/SignerDetails.vue`

Add the new field there with the other signer details.

Use the existing `NcListItem` pattern.

Do not create a new validation section only for this value.

## How to detect the value

The API already provides `visibleElements` in signer data.

Do not add a new backend endpoint or a new database field.

Use the existing signer data:

```ts
signer.visibleElements
```

The value can be decided as:

```ts
Array.isArray(signer.visibleElements) && signer.visibleElements.length > 0
```

If there is at least one visible element:

> **Visible signature:** Yes

If there are no visible elements:

> **Visible signature:** No

Do not try to detect this by reading the PDF in the frontend.

Do not create new logic to match signers and elements when `visibleElements` is already available on the signer.

## Envelope validation

For envelopes, keep this information at document signer level.

A signer may have a visible signature in one document and no visible signature in another document.

Because of this, do not show one global `Yes` or `No` value in the consolidated `Signers summary`.

The value should describe the signer for the specific document being inspected.

## Display rules

Show this information for signed signers.

It should appear with the existing signer metadata, such as:

* date signed;
* validation status;
* hash algorithm;
* remote address;
* user agent.

The exact position should follow the current layout, but it should not be inside `Validation status`, because having no visible signature is not a validation problem.

## Tests

Add or update the tests for `SignerDetails.vue`.

Cover at least:

* signer with one visible element shows `Visible signature: Yes`;
* signer with several visible elements shows `Yes`;
* signer with an empty `visibleElements` array shows `No`;
* signer without visible elements does not show any warning or validation error;
* existing validation information still renders normally.

If there is already validation-page integration coverage for signer details, extend it with one focused case.

Do not create a complete new validation workflow only for this field.

## Out of scope

This issue does not:

* change signature validation;
* change the PDF;
[...]

PR #8332, opened 2026-09-10
fix: populate signer visible elements in validation responses

vdriga85:
## Summary

Backend prerequisite for #8326, as requested in https://github.com/LibreSign/libresign/issues/8326#issuecomment-5619485424. This does not close the frontend issue.

- Populate `signer.visibleElements` in `FileService` and envelope child documents in `EnvelopeAssembler`, keyed by `signRequestId`.
- Reuse the existing mapper result and formatted elements; preserve document-level elements and ordering without adding queries.
- Add unit and Behat coverage for signer/document isolation.
- Describe optional child-signer visible elements in OpenAPI and regenerate the JSON specs and TypeScript types. The field remains optional because single-file summaries share this contract.

No UI, database, PDF or cryptographic validation changes are included.

## How to test

In a disposable configured Nextcloud development environment:

```sh
vendor/bin/phpunit -c tests/php/phpunit.xml --no-coverage --fail-on-warning --filter FileServiceTest tests/php/Unit/Service/FileServiceTest.php
vendor/bin/phpunit -c tests/php/phpunit.xml --no-coverage --fail-on-warning --filter EnvelopeAssemblerTest tests/php/Unit/Service/File/EnvelopeAssemblerTest.php
cd tests/integration
vendor/bin/behat -c config/behat.yml features/file/validate.feature --name "visible elements" --format=progress
```

Local results on PHP 8.3.33 / SQLite: 44 unit tests, 133 assertions; 2 Behat scenarios, 28 steps passed. Focused PHP syntax/style checks and `git diff --check` passed. Full suites and Psalm were not run.

Also rechecked the original signed PDF: validation now returns the existing element both at document level and for its matching signer; `signature_validation.isValid` remains true and CRL validation remains valid. The separately developed local UI changed from No to Yes; that UI is not part of this PR.

## Checklist

- [x] Contribution guide followed; commit includes DCO sign-off.
- [x] Unit and integration tests added.
- [x] OpenAPI specs and TypeScript types regenerated.

## AI

- [x] The content of this PR was partially or fully generated using AI.

Implemented with Codex assistance, independently reviewed with AI assistance, and manually verified by the contributor on the original signed document.

### DIFF
--- a/lib/Service/FileService.php
+++ b/lib/Service/FileService.php
@@ -520,16 +520,21 @@ class FileService {
 
 		$signers = $this->signRequestMapper->getByMultipleFileId($fileIds);
 		$fileMetadata = $this->file->getMetadata();
-		foreach ($this->signRequestMapper->getVisibleElementsFromSigners($signers) as $visibleElements) {
+		$formattedElementsBySigner = [];
+		foreach ($this->signRequestMapper->getVisibleElementsFromSigners($signers) as $signRequestId => $visibleElements) {
 			if (empty($visibleElements)) {
 				continue;
 			}
 			$elementFileId = $visibleElements[0]->getFileId();
 			$metadata = $childMetadataMap[$elementFileId] ?? $fileMetadata;
+			$formattedElementsBySigner[$signRequestId] = $this->fileElementService->formatVisibleElements($visibleElements, $metadata);
 			$this->fileData->visibleElements = array_merge(
-				$this->fileElementService->formatVisibleElements($visibleElements, $metadata),
+				$formattedElementsBySigner[$signRequestId],
 				$this->fileData->visibleElements
 			);
+		}
+		foreach ($this->fileData->signers as $signer) {
+			$signer->visibleElements = $formattedElementsBySigner[$signer->signRequestId ?? null] ?? [];
 		}
 	}
 
--- a/lib/Service/FileService.php
+++ b/lib/Service/FileService.php
@@ -53,7 +53,7 @@ use TypeError;
 /**
  * @psalm-import-type LibresignValidatedFile from ResponseDefinitions
  * @psalm-import-type LibresignSignerDetail from ResponseDefinitions
- * @psalm-import-type LibresignSignerSummary from ResponseDefinitions
+ * @psalm-import-type LibresignValidatedChildSigner from ResponseDefinitions
  * @psalm-import-type LibresignIdentifyMethod from ResponseDefinitions
  */
 class FileService {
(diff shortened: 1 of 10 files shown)


lib/Service/FileService.php
```diff
@@ -520,17 +520,22 @@ private function loadVisibleElements(): void {
...
+			$formattedElementsBySigner[$signRequestId] = $this->fileElementService->formatVisibleElements($visibleElements, $metadata);
 			$this->fileData->visibleElements = array_merge(
-				$this->fileElementService->formatVisibleElements($visibleElements, $metadata),
+				$formattedElementsBySigner[$signRequestId],
 				$this->fileData->visibleElements
 			);
 		}
+		foreach ($this->fileData->signers as $signer) {
+			$signer->visibleElements = $formattedElementsBySigner[$signer->signRequestId ?? null] ?? [];
+		}
```
vitormattos, line 538: It looks like `loadVisibleElements()` now correctly fills `this->fileData->signers[*].visibleElements`, but this value may still be lost later for the single-file child representation.

`syncSingleFileCollection()` builds `files[0].signers` through `mapSignerDetailsToSummary()`, and that mapper does not currently include `visibleElements`.

Because of this, a single-file validation response can end up with:

* `data.signers[*].visibleElements` populated correctly;
* `data.files[0].signers[*].visibleElements` missing.

Would it make sense to preserve each signer's `visibleElements` in `mapSignerDetailsToSummary()` as well? This could reuse the data already present on the signer, without adding a new query or extra matching logic.

It would also be useful to extend the unit/Behat coverage so the single-file response checks that `data.files[0].signers[*].visibleElements` matches the corresponding `data.signers[*].visibleElements`, including the empty-array case for a signer without visible elements.

That would keep the runtime response aligned with the new `ValidatedChildSigner` OpenAPI type introduced in this PR.

vitormattos (changes requested):

lib/Service/FileService.php
vdriga85: You're right, I missed this when implementing the initial fix. Thanks for catching it.
I've updated `mapSignerDetailsToSummary()` locally to preserve each signer's `visibleElements`, including empty arrays, without adding queries or extra matching logic. I also updated its return annotation to `LibresignValidatedChildSigner[]`.
The unit and Behat tests now check that `files[0].signers[*].visibleElements` matches the corresponding top-level signer, including the case without visible elements.
All 44 focused unit tests and 2 Behat scenarios passed, along with the PHP style checks. OpenAPI and TypeScript regeneration produced no content changes.
The changes have now been pushed to this PR.

vitormattos (approved):

Issue #8335, opened 2026-09-10
Add external digital signing to LibreSign


vitormattos:
LibreSign can already sign PDF documents when the signing process has direct access to the certificate and private key.

The next step is to support signatures where the private key stays outside LibreSign.

This is needed for signing models such as remote certificates, HSM-backed credentials and, in the future, certificates stored on local devices.

It also prepares LibreSign for services based on standards such as CSC and ICP-Brasil PSC.
> [!NOTE]
>
> This epic defines the goal, architecture and scope.
> **Pull requests should implement focused child issues, not the epic itself.**

> [!IMPORTANT]
>
> ## Funding and community support needed
>
> External signing can help LibreSign work with more digital identity and trust service ecosystems around the world.
>
> **We are looking for organizations interested in funding this work and for contributors who want to help build it.**
>
> Organizations that need remote certificates, HSM-backed signing, CSC, ICP-Brasil PSC or other external signing solutions are especially welcome to support this work.
>
> Community contributions are also welcome. Implementation will be split into focused issues with the expected architecture and behavior already defined.

## Goal

LibreSign must be able to prepare a digital signature without having the private key, send the required signing data to another component, receive the signature value and finish the signed PDF.

The existing signing methods must continue working as they do today.

## Architecture

Today, the signing path expects the private key to be available inside the signing process:

```mermaid
flowchart LR
    L["LibreSign workflow"] --> H["Sign engine handler"]

    subgraph C["Current signing operation"]
        P["Prepare PDF / PAdES"] --> K["Use private key"]
        K --> S["Create signature"]
        S --> F["Finalize PDF / PAdES"]
    end

    H --> P
    F --> D["Signed document"]
```

This epic does not replace this complete flow.

It adds a new signing path for cases where the private key is outside the signing engine:

```mermaid
flowchart LR
    L["LibreSign workflow"] --> H["Sign engine handler"]

    H --> C["Existing signing path"]
    C --> D["Signed document"]

    H --> P["Prepare external signature"]
    P --> T["Data to sign"]
    T --> X["External signer"]
    X --> V["Signature value"]
    V --> F["Finalize external signature"]
    F --> D

    X -.-> R["Future remote signing"]
    X -.-> LD["Future local-device signing"]
```

The important change is the new boundary between PDF/PAdES processing and the private-key operation.

The current signing path can remain unchanged while this new capability is added.

## External signing responsibilities

The PDF/PAdES side is responsible for:

[...]

PR #8347, opened 2026-09-11
refactor(sign-engine): return the sign engine interface from the factory

### LINKED ISSUE #8344: Generalize `SignEngineFactory` to return the sign engine interface
This is part of #8335.

LibreSign already has the `ISignEngineHandler` interface, and the current sign engine handlers implement it through `SignEngineHandler`.

However, `SignEngineFactory::resolve()` currently declares its return type using the concrete handlers:

```php
public function resolve(string $extension): Pkcs12Handler|Pkcs7Handler
```

This means every new handler added to this factory would also require changing its public return type.

Before adding new signing paths, we should make the factory depend on the existing handler abstraction instead.

## Current behavior

The current factory selects between two existing signing implementations:

- PDF files use `Pkcs12Handler`, which handles the current PDF signing flow.
- Other file types use `Pkcs7Handler`, which creates a detached PKCS#7 (`.p7s`) signature.

The extension matching is case-insensitive.

```mermaid
flowchart LR
    E["File extension"] --> F["SignEngineFactory"]

    F -->|"pdf"| PDF["Pkcs12Handler<br/>PDF signing"]
    F -->|"other extension"| P7["Pkcs7Handler<br/>Detached PKCS#7 signature"]

    PDF --> I["ISignEngineHandler"]
    P7 --> I
```

## Goal

Change `SignEngineFactory` so its public contract returns `ISignEngineHandler` while preserving exactly the same runtime behavior.

The factory must continue selecting:

- `Pkcs12Handler` for PDF files;
- `Pkcs7Handler` for non-PDF files.

This issue must not add external signing yet.

## Expected change

Update:

`lib/Handler/SignEngine/SignEngineFactory.php`

from:

```php
public function resolve(string $extension): Pkcs12Handler|Pkcs7Handler
```

to:

```php
public function resolve(string $extension): ISignEngineHandler
```

Do not change how the handler is selected.

The returned handler must still be the exact instance provided by the dependency injection container.

## Tests

Add focused unit tests for `SignEngineFactory`.

Mock `Psr\Container\ContainerInterface` and provide mocked handler instances.

Use a PHPUnit DataProvider to cover at least:

| Extension | Expected handler |
| --- | --- |
| `pdf` | `Pkcs12Handler` |
| `PDF` | `Pkcs12Handler` |
| `Pdf` | `Pkcs12Handler` |
| `docx` | `Pkcs7Handler` |
| `odt` | `Pkcs7Handler` |
| `txt` | `Pkcs7Handler` |

For each case, verify that:

- the expected handler class is requested from the container;
- `resolve()` returns the same handler instance supplied by the container;
- the returned value implements `ISignEngineHandler`.

Follow the existing LibreSign PHPUnit conventions for namespace, test location, mocks and naming.

## Why this helps #8335

External signing will introduce signing behavior that does not fit the current certificate-and-password flow.

This issue does not decide how that future handler will work.

It only removes an unnecessary concrete return-type restriction from the current factory, allowing the existing abstraction to remain stable when new signing paths are introduced later.

The current PDF and detached PKCS#7 signing paths must remain unchanged.

The external signing prepare/finalize architecture is being discussed separately in:

https://github.com/jeidison/signer-php/issues/25

## Scope boundaries

Please keep this change focused.

Do not:

- add a new signing handler;
- change `ISignEngineHandler`;
- change `SignEngineHandler`;
- change `Pkcs12Handler`;
- change `Pkcs7Handler`;
- change the extension-to-handler mapping;
- change the detached PKCS#7 implementation;
- add CSC or ICP-Brasil PSC support;
- add external signing methods;
- change any signing workflow.

## Good first issue

This is a small backend refactor based on an abstraction that already exists in LibreSign.

The main work is:

- change the return type of `SignEngineFactory::resolve()` to `ISignEngineHandler`;
- keep the existing PDF and detached PKCS#7 handler selection unchanged;
- add focused unit tests using a DataProvider;
- verify that the dependency injection container is asked for the expected handler;
- verify that the same handler instance is returned.

No new signing architecture, provider integration, frontend change, database migration or API change is needed.

Please keep the change focused. If the implementation seems to require changes to `ISignEngineHandler`, new handlers or a broader signing-engine refactor, discuss it in this issue before adding that complexity.

### Additional context

## Done when

- [ ] `SignEngineFactory::resolve()` returns `ISignEngineHandler`.
- [ ] PDF files still resolve to `Pkcs12Handler`.
- [ ] PDF extension matching remains case-insensitive.
- [ ] Non-PDF files still resolve to the existing detached PKCS#7 signing path through `Pkcs7Handler`.
- [ ] Unit tests cover the factory behavior using a DataProvider.
- [ ] The expected handler class is requested from the container.
- [ ] The same handler instance provided by the container is returned.
- [ ] The returned handler satisfies `ISignEngineHandler`.
- [ ] No production signing behavior is changed.

maia-andre:
Resolves: #8344

## 📝 Summary

`SignEngineFactory::resolve()` declared its return type with the two concrete handlers, so every new signing path added to the factory would also have to change its public contract. LibreSign already has the `ISignEngineHandler` abstraction that both handlers implement, so the factory can depend on it instead.

```diff
-	public function resolve(string $extension): Pkcs12Handler|Pkcs7Handler {
+	public function resolve(string $extension): ISignEngineHandler {
```

The `match`, the extension matching and the container lookups are untouched, so PDF files still resolve to `Pkcs12Handler`, everything else still resolves to `Pkcs7Handler`, and no signing behaviour changes.

### Why `SignFileService` is in this pull request

Changing only the factory does not pass the psalm job. `SignFileService::identifyEngine()` is the one caller that consumes the factory with a declared type, and it declares the abstract class:

```
ERROR: MoreSpecificReturnType - lib/Service/SignFileService.php:951
  The declared return type 'SignEngineHandler' for SignFileService::identifyEngine
  is more specific than the inferred return type 'ISignEngineHandler'

ERROR: LessSpecificReturnStatement - lib/Service/SignFileService.php:952
  The type 'ISignEngineHandler' is more general than the declared return type
  'SignEngineHandler' for SignFileService::identifyEngine
```

Three declarations carry the resolved engine through that class, and widening them to the interface clears both errors:

| line | before | after |
| --- | --- | --- |
| 89 | `private ?SignEngineHandler $engine` | `private ?ISignEngineHandler $engine` |
| 954 | `identifyEngine(File $file): SignEngineHandler` | `identifyEngine(File $file): ISignEngineHandler` |
| 1229 | `getEngine(): SignEngineHandler` | `getEngine(): ISignEngineHandler` |

Nothing else in the file changes. `PdfSignatureDetectionService`, the other caller, is unaffected: it assigns to a local variable and only calls `getCertificateChain()`, which the interface declares.

This is the only part that goes beyond the file named in the issue. It is not in the issue's "do not change" list, and without it the psalm job fails — but if you prefer a different split, say so and I will rework it.

### What `PfxProvider` tells us

`PfxProvider::getOrGeneratePfx()` keeps taking `SignEngineHandler`, because it calls three methods that are **not** part of `ISignEngineHandler`:

* `setLeafExpiryOverrideInDays()`
* `generateCertificate()`
* `getPfxOfCurrentSigner()`

Psalm reports that as one informational `ArgumentTypeCoercion` at `SignFileService.php:1240`, which does not fail CI at `errorLevel="5"`. I left it visible rather than papering over it: the signing path resolved by the factory is expressible through the interface, but the certificate-provisioning path around it is not, and that looks like useful information for #8335. Making `PfxProvider` accept the interface would require adding those three methods to `ISignEngineHandler`, which this issue's scope rules out.

### Tests

`tests/php/Unit/Handler/SignEngine/SignEngineFactoryTest.php` already existed with a DataProvider, but it resolved the real container with `\OCP\Server::get()`, so it could only assert the resulting class.

It now mocks `ContainerInterface`, which lets it check the rest of the issue's checklist:

* the expected handler class is the one requested from the container (`->with($instanceOf)`);
* `resolve()` returns the very instance the container supplied (`assertSame`);
* the returned value satisfies `ISignEngineHandler`.

The extensions it already covered are kept, and `Pdf` and `docx` from the issue's table were added: `pdf`, `PDF`, `Pdf`, `docx`, `odt`, `ODT`, `jpg`, `JPG`, `png`, `PNG`, `txt`, `TXT`.

## 🧪 How to test

```sh
composer psalm
vendor/bin/phpunit -c tests/php/phpunit.xml --no-coverage --testsuite unit --filter 'SignEngineFactoryTest'
vendor/bin/phpunit -c tests/php/phpunit.xml --no-coverage --testsuite unit --filter 'SignFileServiceTest|PfxProviderTest|PdfSignatureDetectionServiceTest'
composer cs:check
```

Run locally in the devcontainer:

| command | result |
| --- | --- |
| psalm on the four files involved | same 5 pre-existing `MissingDependency` errors as the unmodified branch, no new error |
| `SignEngineFactoryTest` | OK, 12 tests, 48 assertions |
| `SignFileServiceTest` + `PfxProviderTest` + `PdfSignatureDetectionServiceTest` | OK, 165 tests, 386 assertions |
| `php-cs-fixer` on the three changed files | 0 of 3 files need fixing |

## ⚙️ API / Back‑end changes

- [x] `SignEngineFactory::resolve()` returns `ISignEngineHandler`
- [x] The three declarations in `SignFileService` that carry the resolved engine widen to the interface
- [x] Unit tests added – *required for backend changes*
- [ ] Capabilities updated – not applicable, no capability changes
- [ ] Documentation updated – not applicable, no user-visible or API change
- [ ] `composer openapi` – not applicable, no endpoint or payload changes

## 🔁 Backport notes

The cherry-pick applies cleanly to `stable32`, `stable33`, `stable34` and `stable35` (checked with `git merge-tree --write-tree`). `SignEngineFactory.php` and the test file are identical across all four branches, and the three declarations in `SignFileService` exist in all of them, at different line numbers.

That said, this is preparation for #8335 rather than a fix, so whether it should land on the stable lines at all is your call.

## ✅ Checklist

- [x] No change to `ISignEngineHandler`, `SignEngineHandler`, `Pkcs12Handler`, `Pkcs7Handler`, the extension mapping, the detached PKCS#7 implementation or any signing workflow.
- [x] No new handler and no external signing code.

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Handler/SignEngine/SignEngineFactory.php
+++ b/lib/Handler/SignEngine/SignEngineFactory.php
@@ -16,7 +16,7 @@ class SignEngineFactory {
 	) {
 	}
 
-	public function resolve(string $extension): Pkcs12Handler|Pkcs7Handler {
+	public function resolve(string $extension): ISignEngineHandler {
 		return match (strtolower($extension)) {
 			'pdf' => $this->container->get(Pkcs12Handler::class),
 			default => $this->container->get(Pkcs7Handler::class),
--- a/lib/Service/SignFileService.php
+++ b/lib/Service/SignFileService.php
@@ -35,6 +35,7 @@ use OCA\Libresign\Exception\LibresignException;
 use OCA\Libresign\Handler\DocMdpHandler;
 use OCA\Libresign\Handler\FooterHandler;
 use OCA\Libresign\Handler\PdfTk\Pdf;
+use OCA\Libresign\Handler\SignEngine\ISignEngineHandler;
 use OCA\Libresign\Handler\SignEngine\Pkcs12Handler;
 use OCA\Libresign\Handler\SignEngine\SignEngineFactory;
 use OCA\Libresign\Handler\SignEngine\SignEngineHandler;
@@ -85,7 +86,7 @@ class SignFileService {
 	private string $userUniqueIdentifier = '';
 	private string $friendlyName = '';
 	private ?IUser $user = null;
-	private ?SignEngineHandler $engine = null;
+	private ?ISignEngineHandler $engine = null;
 
 	public function __construct(
 		protected IL10N $l10n,
@@ -950,7 +951,7 @@ class SignFileService {
 		$this->eventDispatcher->dispatchTyped($event);
 	}
 
-	protected function identifyEngine(File $file): SignEngineHandler {
+	protected function identifyEngine(File $file): ISignEngineHandler {
 		return $this->signEngineFactory->resolve($file->getExtension());
 	}
 
@@ -1225,7 +1226,7 @@ class SignFileService {
 		return strcasecmp($file->getExtension(), 'pdf') === 0;
 	}
 
-	protected function getEngine(): SignEngineHandler {
+	protected function getEngine(): ISignEngineHandler {
 		if (!$this->engine) {
 			$originalFile = $this->getFileToSign();
 			$this->engine = $this->identifyEngine($originalFile);
--- a/tests/php/Unit/Handler/SignEngine/SignEngineFactoryTest.php
+++ b/tests/php/Unit/Handler/SignEngine/SignEngineFactoryTest.php
@@ -4,10 +4,12 @@ declare(strict_types=1);
 
 namespace OCA\Libresign\Tests\Unit\Handler\SignEngine;
 
+use OCA\Libresign\Handler\SignEngine\ISignEngineHandler;
 use OCA\Libresign\Handler\SignEngine\Pkcs12Handler;
 use OCA\Libresign\Handler\SignEngine\Pkcs7Handler;
 use OCA\Libresign\Handler\SignEngine\SignEngineFactory;
 use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\MockObject\MockObject;
 use Psr\Container\ContainerInterface;
 
 /**
@@ -16,11 +18,11 @@ use Psr\Container\ContainerInterface;
  */
 
 final class SignEngineFactoryTest extends \OCA\Libresign\Tests\Unit\TestCase {
-	private ContainerInterface $container;
+	private ContainerInterface&MockObject $container;
 
 	public function setUp(): void {
 		parent::setUp();
-		$this->container = \OCP\Server::get(ContainerInterface::class);
+		$this->container = $this->createMock(ContainerInterface::class);
 	}
 
 	private function getInstance(): SignEngineFactory {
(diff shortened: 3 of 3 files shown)


vitormattos (approved):
This change looks correct for the scope of #8344.

One thing worth keeping visible for the next external-signing work: `getEngine()` now exposes the engine as `ISignEngineHandler`, but `configureEngine()` still passes it to `getOrGeneratePfxContent()`, which requires `SignEngineHandler`.

This is safe today because the factory still returns only `Pkcs12Handler` and `Pkcs7Handler`, and both extend `SignEngineHandler`. I do not think we should expand the scope of this PR to solve it.

But when we add an engine that only implements `ISignEngineHandler`, this concrete dependency will need to be addressed. It would be good to keep this constraint in mind for #8335.

vitormattos:
Thanks for checking the backport compatibility.

I think we should keep this change on `main` only. This refactor is preparation for the new signing architecture planned for the Nextcloud 36 line, rather than a bug fix for the current stable releases.

So I would not backport it to `stable32`–`stable35` unless a concrete fix in one of those branches later requires it.

Issue #8321, opened 2026-09-09
Clarify PDF signatures without visible signature fields


vitormattos:
## Clarify PDF signatures without visible signature fields

A visible signature is optional in a PDF digital signature.

A PDF can be digitally signed without showing a signature mark on the page. The digital signature is still stored in the PDF and can be validated.

This is standard PDF behavior, not a special LibreSign feature.

However, LibreSign does not explain this clearly enough. This has caused repeated confusion for both requesters and signers.

This epic replaces the related discussions from #7710 and #8320.
> [!NOTE]
>
> This epic defines the goal, architecture and scope.
> **Pull requests should implement focused child issues, not the epic itself.**

> [!IMPORTANT]
>
> ## Funding and community support
>
> This work affects the request flow, signer flow, user preferences, backend persistence and automated tests.
>
> Organizations that depend on this workflow and want to help prioritize its implementation are welcome to support the development.
>
> Community contributions are also welcome. Please work on the focused implementation issues instead of trying to complete the full epic in one pull request.

## Goal

LibreSign must keep visible signature fields optional while making the behavior clear in the interface.

The UI should clearly separate:

* the digital signature stored in the PDF;
* the optional visible representation shown on a page.

A document without visible signature fields must remain valid and signable.

## Requester flow

When preparing a signature request, the interface must not suggest that a visible signature field is required.

The requester may:

1. add one or more signers;
2. optionally open `Setup signature positions`;
3. optionally add visible signature fields;
4. continue with `Request signatures`.

Before sending, LibreSign must check the visible fields assigned to each signer.

If all signers have visible fields, keep the normal confirmation flow.

If one or more signers have no visible signature field, reuse the existing request confirmation `NcDialog` and show the warning there.

Do not add another confirmation dialog.

Suggested content:

> **Some signers have no visible signature field**
>
> A PDF can be digitally signed without showing a signature on the page.
>
> These signers will still add a digital signature that can be validated, but no visible signature mark will be shown for them.

When possible, list only the affected signers.

For example:

> No visible signature:
>
> * Bob
> * Carol

Actions:

* `Cancel`
* `Send`

The same rule must apply to the single-signer `Request signature` action. In that flow, only the selected signer must be checked.

The request must never be blocked only because a visible field is missing.

## Warning preference

The warning dialog may include this optional checkbox:

> **Do not warn me again when signers have no visible signature field**

Below it, show:

> You can enable this warning again in LibreSign preferences.

[...]

PR #8349, opened 2026-09-11
feat(account): add the missing visible signature warning preference

### LINKED ISSUE #8322: Add user preference for the missing visible signature warning
## Context

Part of #8321.

LibreSign will warn a requester when one or more signers have no visible signature field.

The requester can choose not to see this warning again.

LibreSign already has a generic API to save user configuration. This issue only needs to add the new preference to the existing user configuration flow.

## Preference

Add this user configuration key:

`warn_without_visible_signature_fields`

The value controls only whether the warning is shown.

Expected values:

* `1`: show the warning;
* `0`: do not show the warning.

When the user has no saved value, the default must be `1`.

This preference must not change signature fields, signature positions, requests or signing behavior.

## Backend changes

Reuse the existing endpoint:

`PUT /apps/libresign/api/v1/account/config/{key}`

Do not create a new controller or endpoint.

The existing endpoint already stores values for the authenticated user.

Update `AccountService::getConfig()` so the frontend receives:

```json
{
  "warn_without_visible_signature_fields": true
}
```

when the warning is enabled, and:

```json
{
  "warn_without_visible_signature_fields": false
}
```

when it is disabled.

If no value is stored, return `true`.

Follow the same pattern already used by boolean user settings such as:

* `files_list_grid_view`;
* `policy_workbench_catalog_compact_view`;
* `policy_workbench_catalog_collapsed`.

Do not add a system policy or group configuration.

## Reset behavior

The existing user-config API uses an empty value to remove/reset a user setting.

For this preference, resetting the value must restore the default:

`warn_without_visible_signature_fields = true`

No migration is needed because this is a new preference and the default applies when no value exists.

## Behat tests

Update:

`tests/integration/features/settings/user-config.feature`

This file already tests save, update and delete behavior for user configuration keys.

Add `warn_without_visible_signature_fields` to the existing CRUD scenario.

Cover:

1. save `0` and confirm the API accepts it;
2. update it to `1`;
3. reset it with an empty value;
4. confirm the returned user configuration uses the default `true` after reset.

Keep the existing test style instead of creating a separate feature file.

If the existing scenario already covers authentication for this endpoint, do not duplicate that test only for this key.

## Unit tests

Update the existing `AccountService` unit tests for `getConfig()`.

Cover at least:

* stored `1` returns `true`;
* stored `0` returns `false`;
* no stored value returns the default `true`.

Use a data provider if it keeps these cases small and clear.

## Quality

Run the related backend checks and tests.

At minimum:

* the affected PHPUnit tests;
* `tests/integration/features/settings/user-config.feature`;
* Psalm;
* PHPCS.

Do not change unrelated user configuration behavior.

## Good first issue

This is a small backend change using an existing LibreSign user configuration API.

The main work is:

* add one user configuration key to `AccountService::getConfig()`;
* reuse the existing `/account/config/{key}` endpoint;
* extend the existing PHPUnit tests;
* extend `tests/integration/features/settings/user-config.feature`.

No new database table, migration, controller or policy is needed.

Please follow the existing user configuration patterns and keep the change focused.

### Additional context

## Done when

* [ ] `warn_without_visible_signature_fields` is available in user config.
* [ ] The default value is `true`.
* [ ] `0` is returned as `false`.
* [ ] `1` is returned as `true`.
* [ ] Resetting the stored value restores the default.
* [ ] The existing user-config API is reused.
* [ ] PHPUnit covers the value mapping and default.
* [ ] Behat covers save, update and reset.
* [ ] Psalm and PHPCS pass.

maia-andre:
Resolves: #8322

Part of #8321.

## 📝 Summary

Adds the `warn_without_visible_signature_fields` user preference to `AccountService::getConfig()`, so a requester can turn off the warning about signers without a visible signature field.

```php
$info['warn_without_visible_signature_fields'] =
    $this->getUserConfigByKey('warn_without_visible_signature_fields', $user) !== '0';
```

No endpoint was added: `PUT /apps/libresign/api/v1/account/config/{key}` already stores any key for the authenticated user, and the empty value it accepts is what makes the reset work. No migration is needed.

### Why `!== '0'` instead of the `=== '1'` used by the other booleans

The three preferences the issue points at — `files_list_grid_view`, `policy_workbench_catalog_compact_view` and `policy_workbench_catalog_collapsed` — map the stored value with `=== '1'`. `getUserConfigByKey()` returns `''` when nothing is stored, so that expression **defaults to `false`**.

This preference has to default to `true`. Mapping it with `!== '0'` gives the three behaviours the issue asks for with one expression:

| stored value | result | why |
| --- | --- | --- |
| `'1'` | `true` | warning shown |
| `'0'` | `false` | warning hidden |
| `''` (never set, or reset) | `true` | default |

I checked that this is not theoretical: implementing it with `=== '1'` and running the new tests fails exactly the default case with `Failed asserting that false is identical to true`, so the test does guard the distinction.

Also relevant: `getConfig()` ends with `array_filter($info, fn ($value) => $value !== null && $value !== '')`, whose callback keeps `false`. A bare `array_filter()` would drop it and the key would vanish from the payload when the user disables the warning — see the backport notes below.

## 🧪 How to test

```sh
vendor/bin/phpunit -c tests/php/phpunit.xml --no-coverage --testsuite unit --filter AccountServiceTest
composer psalm
composer cs:check
```

Manually, as any user:

```sh
# hide the warning
curl -u user:pass -X PUT -H 'OCS-APIRequest: true' \
  'http://localhost/ocs/v2.php/apps/libresign/api/v1/account/config/warn_without_visible_signature_fields' \
  -d 'value=0'

# reset it, which restores the default
curl -u user:pass -X PUT -H 'OCS-APIRequest: true' \
  'http://localhost/ocs/v2.php/apps/libresign/api/v1/account/config/warn_without_visible_signature_fields' \
  -d 'value='
```

Run locally in the devcontainer:

| command | result |
| --- | --- |
| `AccountServiceTest` | OK, 36 tests, 86 assertions (3 new) |
| psalm on `AccountService` and `AccountController` | 4 errors before, 4 after — no new error |
| `php-cs-fixer` on the two changed PHP files | 0 of 2 files need fixing |

`tests/integration` has no vendor directory installed locally and the suite needs a running server, so Behat was left to CI.

## ⚙️ API / Back‑end changes

- [x] `warn_without_visible_signature_fields` added to `AccountService::getConfig()`
- [x] Existing `PUT /account/config/{key}` endpoint reused, no new controller or route
- [x] Unit and integration tests added – *required for backend changes*
- [ ] Capabilities updated – not applicable, this is a user preference, not a capability
- [ ] Documentation updated – not applicable, no user-visible behaviour yet; the warning itself comes with the rest of #8321
- [ ] `composer openapi` – not applicable, no endpoint, parameter or response type changed

### Tests

**PHPUnit** — `testGetConfigIncludesWarnWithoutVisibleSignatureFieldsPreference`, a DataProvider with the three cases the issue lists: stored `1` returns `true`, stored `0` returns `false`, nothing stored returns the default `true`.

**Behat** — `warn_without_visible_signature_fields` added to the three blocks of the existing CRUD scenario in `tests/integration/features/settings/user-config.feature`, in the file's own style: saved as `0`, updated to `1`, reset with an empty value.

One item from the issue's Behat list is covered by PHPUnit instead, and I would rather say so than quietly skip it. Item 4 asks to *confirm the returned user configuration uses the default `true` after reset*, but no OCS endpoint exposes `getConfig()`: `/account/me` returns `getSettings()` (`canRequestSign`, `hasSignatureFile`, `phoneNumber`), and `getConfig()` only reaches the frontend through `initialState` in `PageController`, which Behat does not read. So Behat covers the API contract — save, update, reset — and the default after reset is asserted in the unit test, where the mapping actually lives. If you would rather have an endpoint that exposes the user config so Behat can assert it end to end, that is a larger change and I would open it separately.

## 🔁 Backport notes

The cherry-pick applies cleanly to **`stable35` only**. It conflicts on `stable32`, `stable33` and `stable34`, and there is a functional reason behind the conflict rather than just drifting context:

```php
// main and stable35
return array_filter($info, static fn (mixed $value): bool => $value !== null && $value !== '');

// stable32, stable33, stable34
return array_filter($info);
```

The bare `array_filter()` on those branches drops every falsy value, so `warn_without_visible_signature_fields = false` would be removed from the payload entirely. The frontend would see the key missing and show the warning again — the opposite of what the user asked for. Backporting there would need the filter changed too, which is outside this issue.

Since this is part of #8321 and has no user-visible effect on its own, I would suggest `stable35` at most, but that is your call.

## ✅ Checklist

- [x] No new endpoint, controller, migration, policy or group configuration.
- [x] No unrelated user configuration behaviour changed.

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Service/AccountService.php
+++ b/lib/Service/AccountService.php
@@ -227,6 +227,7 @@ class AccountService {
 		$info['policy_workbench_catalog_compact_view'] = $this->getUserConfigByKey('policy_workbench_catalog_compact_view', $user) === '1';
 		$info['policy_workbench_catalog_collapsed'] = $this->getUserConfigByKey('policy_workbench_catalog_collapsed', $user) === '1';
 		$info['policy_workbench_category_collapsed_state'] = $this->getUserConfigJsonByKey('policy_workbench_category_collapsed_state', $user);
+		$info['warn_without_visible_signature_fields'] = $this->getUserConfigByKey('warn_without_visible_signature_fields', $user) !== '0';
 		$info['can_manage_group_policies'] = $this->policyAuthorizationService->canUserManageGroupPolicies($user);
 		$info['manageable_policy_group_ids'] = $this->policyAuthorizationService->getManageablePolicyGroupIds($user);
 
--- a/tests/integration/features/settings/user-config.feature
+++ b/tests/integration/features/settings/user-config.feature
@@ -95,6 +95,14 @@ Feature: settings/user-config
       | key                 | value                               |
       | (jq).ocs.data.key   | id_docs_sort                        |
 
+    And sending "put" to ocs "/apps/libresign/api/v1/account/config/warn_without_visible_signature_fields"
+      | value | 0 |
+    Then the response should have a status code 200
+    And the response should be a JSON array with the following mandatory values
+      | key                 | value                                 |
+      | (jq).ocs.data.key   | warn_without_visible_signature_fields |
+      | (jq).ocs.data.value | 0                                     |
+
     # Update all configs and validate returned value
     And sending "put" to ocs "/apps/libresign/api/v1/account/config/files_list_grid_view"
       | value | 0 |
@@ -178,6 +186,14 @@ Feature: settings/user-config
     And the response should be a JSON array with the following mandatory values
       | key                 | value                              |
       | (jq).ocs.data.key   | id_docs_sort                       |
+
+    And sending "put" to ocs "/apps/libresign/api/v1/account/config/warn_without_visible_signature_fields"
+      | value | 1 |
+    Then the response should have a status code 200
+    And the response should be a JSON array with the following mandatory values
+      | key                 | value                                 |
+      | (jq).ocs.data.key   | warn_without_visible_signature_fields |
+      | (jq).ocs.data.value | 1                                     |
 
     # Delete all configs (empty string) and validate returned value
     And sending "put" to ocs "/apps/libresign/api/v1/account/config/files_list_grid_view"
@@ -267,3 +283,11 @@ Feature: settings/user-config
       | key                 | value       |
       | (jq).ocs.data.key   | id_docs_sort |
       | (jq).ocs.data.value |             |
+
+    And sending "put" to ocs "/apps/libresign/api/v1/account/config/warn_without_visible_signature_fields"
+      | value | |
+    Then the response should have a status code 200
+    And the response should be a JSON array with the following mandatory values
+      | key                 | value                                 |
+      | (jq).ocs.data.key   | warn_without_visible_signature_fields |
+      | (jq).ocs.data.value |                                       |
--- a/tests/php/Unit/Service/AccountServiceTest.php
+++ b/tests/php/Unit/Service/AccountServiceTest.php
@@ -277,6 +277,38 @@ final class AccountServiceTest extends \OCA\Libresign\Tests\Unit\TestCase {
 		$this->assertSame($storedCollapsedState, $config['policy_workbench_category_collapsed_state']);
 	}
 
+	#[DataProvider('provideWarnWithoutVisibleSignatureFieldsCases')]
+	public function testGetConfigIncludesWarnWithoutVisibleSignatureFieldsPreference(string $storedValue, bool $expected): void {
+		$user = $this->createMock(IUser::class);
+		$user->method('getUID')->willReturn('preference-user');
+
+		$this->userConfig
+			->expects($this->atLeastOnce())
+			->method('getValueString')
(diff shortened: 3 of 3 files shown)


vitormattos (approved):
Thanks! The implementation looks good to me.

I agree with keeping the Behat coverage focused on the existing user-config API. The default `true` behavior is already covered at the `AccountService` level, where this mapping happens, so I don't think we should add another endpoint only for this test.

Regarding the backport, since this preference is part of the new #8321 feature and has no user-visible effect on its own, I think we should keep it on `main` only and introduce the complete feature from Nextcloud 36.

vitormattos:
Now we need to freeze new features to stable35 because will be released as a soon

PR #8368, opened 2026-09-12
fix: store identification documents of external signers under the file owner

### LINKED ISSUE #8364: Identification documents uploaded by an unauthenticated signer are stored in the app's internal appdata instead of the file owner's folder (viewer shows 0 pages, download returns 404)
## Environment
- LibreSign 14.1.0
- Nextcloud 34.0.4.1
- PHP 8.3.33, MariaDB 10.11.18

## Steps to reproduce
1. Enable "Require identification documents" in LibreSign admin settings.
2. Request a signature from an **external email recipient who does not have a Nextcloud account** (unauthenticated signer flow).
3. As the signer, open the sign link, and when prompted upload an identification document (photo/scan of an ID).
4. As an admin (in `approval_group`), go to LibreSign → "Documents Validation" and open the uploaded identification document ("Abrir archivo" / open file).

## Expected behavior
The PDF viewer shows the uploaded identification document, and it can be downloaded.

## Actual behavior
- The PDF viewer opens but shows "0 of 0" pages (blank).
- The displayed file name is missing its extension (e.g. `INEFrancisco2023.firmado` instead of `.pdf`).
- Attempting to download the file returns **HTTP 404**.
- No error is logged in `nextcloud.log` at upload time — the upload API call returns `200 OK`.

This was 100% reproducible (tested twice with two different identification documents).

## Root cause

`lib/Service/IdDocsService.php`:

```php
public function addFilesToDocumentFolder(
    array $files,
    SignRequest $signRequest,
): void {
    foreach ($files as $fileIndex => $file) {
        $this->validateTypeOfFile($fileIndex, $file);
    }
    foreach ($files as $fileData) {
        $dataToSave = $fileData;
        $dataToSave['signRequest'] = $signRequest;
        $dataToSave['name'] = $fileData['name'] ?? $fileData['type'];
        $file = $this->requestSignatureService->saveFile($dataToSave);          // <-- no 'userManager' set

        $this->idDocsMapper->save($file->getId(), $signRequest->getId(), null, $fileData['type']);
    }
}
```

Compare with the sibling method used for **authenticated** users uploading their own identification documents, which works correctly:

```php
public function addIdDocs(array $files, IUser $user): void {
    ...
    foreach ($files as $fileData) {
        $dataToSave = $fileData;
        $dataToSave['userManager'] = $user;   // <-- set here
        ...
        $file = $this->requestSignatureService->saveFile($dataToSave);
        ...
    }
}
```

`RequestSignatureService::saveFile()` → `FileService::getNodeFromData()` only calls `FolderService::setUserId()` when `$data['userManager'] instanceof \OCP\IUser`:

```php
if (!$this->folderService->getUserId() && $data['userManager'] instanceof \OCP\IUser) {
    $this->folderService->setUserId($data['userManager']->getUID());
}
```

Since `addFilesToDocumentFolder()` never sets `userManager`, `FolderService::userId` stays unset for the whole request. This cascades into `FolderService::getLibreSignDefaultPath()`:

```php
private function getLibreSignDefaultPath(): string {
    if (!$this->userId) {
        return 'unauthenticated';
    }
    ...
}
```

and `FolderService::getContainerFolder()`:

```php
protected function getContainerFolder(): Folder {
    if ($this->getUserId() && !$this->groupManager->isInGroup($this->getUserId(), 'guest_app')) {
        $containerFolder = $this->root->getUserFolder($this->getUserId());
        if ($containerFolder->isUpdateable()) {
            return $containerFolder;
        }
    }
    $containerFolder = $this->appData->getFolder('/');
    ...
}
```

With no `userId`, the file is written successfully — but into the app's own **`appdata_<instanceid>/libresign/unauthenticated/`** folder instead of the file owner's Nextcloud folder. The resulting `nodeId` is stored in the `File` entity's `node_id` column, but that node lives outside the owner's file tree.

Later, when the file is opened/downloaded, `AccountService::getPdfByUuid()` does:
```php
$userId = $this->fileMapper->getStorageUserIdByUuid($uuid);
$this->folderService->setUserId($userId);
...
return $this->folderService->getFileByNodeId($nodeId);
```
This correctly sets the owner's user ID this time and searches **their** user folder — but the node was never created there, so the lookup fails (`NotFoundException` → 404).

I confirmed directly on the filesystem that the uploaded files were **not lost** — they exist intact under `data/appdata_<instanceid>/libresign/unauthenticated/<timestamp>_<filename>/<filename>`, fully readable, correct file size. They're just orphaned from the normal per-user file tree that later lookups search.

## Suggested fix
In `addFilesToDocumentFolder()`, resolve the file owner before calling `saveFile()` and pass them as `userManager`, e.g.:
```php
$owner = null;
$signRequestFileId = $signRequest->getFileId();
if ($signRequestFileId) {
    $signRequestFile = $this->fileMapper->getById($signRequestFileId);
    $owner = $this->userManager->get($signRequestFile->getUserId());
}
foreach ($files as $fileData) {
    $dataToSave = $fileData;
    $dataToSave['signRequest'] = $signRequest;
    $dataToSave['name'] = $fileData['name'] ?? $fileData['type'];
    if ($owner) {
        $dataToSave['userManager'] = $owner;
    }
    $file = $this->requestSignatureService->saveFile($dataToSave);
    $this->idDocsMapper->save($file->getId(), $signRequest->getId(), null, $fileData['type']);
}
```
This mirrors the pattern already used in `addIdDocs()` and requires injecting `OCP\IUserManager` into `IdDocsService`. Tested locally and confirmed this resolves the issue (file is then created inside the owner's own `LibreSign` folder and is visible/downloadable via the normal validation UI, and via WebDAV).

maia-andre:
Resolves: #8364

## 📝 Summary

When an unauthenticated (external) signer uploads an identification document, `IdDocsService::addFilesToDocumentFolder()` called `RequestSignatureService::saveFile()` without a `userManager`. `FileService::getNodeFromData()` only sets the `FolderService` user id from that key, so the node was written to `appdata_<id>/libresign/unauthenticated/`. In the same call, `saveFile()` sets `File.user_id` to the owner of the signed file — and `AccountService::getPdfByUuid()` later uses that `user_id` to look the node up in the owner's folder, where it never was. Result: intact file on disk, empty viewer and 404 on download in the validation UI.

The fix resolves the owner of the signed file with the same lookup `saveFile()` already uses for `user_id` (`SignRequest::getFileId()` → `FileMapper::getById()` → `getUserId()` → `IUserManager::get()`) and passes it as `userManager`, mirroring `addIdDocs()`. If the owner cannot be resolved (no file id, missing row, deleted account) nothing is passed and the current behaviour is kept.

`IUserManager` is appended at the end of the constructor so the change applies to the stable branches, whose constructor differs from `main` after #8336.

## 🧪 How to test

Unit tests in `IdDocsServiceTest`:

- `testAddFilesToDocumentFolderStoresFilesUnderTheOwnerOfTheSignedFile` — fails on `main` with `Failed asserting that null is identical to an object of class IUser`, passes with the fix;
- `testAddFilesToDocumentFolderWithoutResolvableOwnerKeepsCurrentBehaviour` — `userManager` is not set when the owner cannot be resolved.

Local checks in the devcontainer: `IdDocsServiceTest` OK (7 tests, 18 assertions); psalm on `IdDocsService` unchanged (2 pre-existing `MissingDependency`); php-cs-fixer clean.

Manual, as in the issue: enable **Require identification documents**, request a signature from an email recipient without an account, upload the document as that signer, then as an `approval_group` admin open it in **Documents Validation** — the viewer shows the pages and download works. The node now lives in the owner's LibreSign folder.

## ⚙️ API / Back‑end changes

- [x] `IdDocsService` gains an `IUserManager` dependency (container-resolved); no API change
- [x] Unit tests added

## 🚧 Backport

Bug present on `stable32`–`stable35`; the cherry-pick applies cleanly to all four (checked locally on each branch).

## Not covered on purpose

- Documents already written to the appdata folder on affected installs stay unreachable through the UI; a repair command is a separate change (offered in the issue).
- The file name shown without extension is a different symptom.

## ✅ Checklist

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Service/IdDocsService.php
+++ b/lib/Service/IdDocsService.php
@@ -25,6 +25,7 @@ use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\IAppConfig;
 use OCP\IL10N;
 use OCP\IUser;
+use OCP\IUserManager;
 use Sabre\DAV\UUIDUtil;
 
 class IdDocsService {
@@ -41,6 +42,7 @@ class IdDocsService {
 		private IdentifyMethodMapper $identifyMethodMapper,
 		private ITimeFactory $timeFactory,
 		private IAppConfig $appConfig,
+		private IUserManager $userManager,
 	) {
 	}
 
@@ -120,14 +122,34 @@ class IdDocsService {
 		foreach ($files as $fileIndex => $file) {
 			$this->validateTypeOfFile($fileIndex, $file);
 		}
+		// Store the documents under the owner of the file being signed, where
+		// later lookups by the stored user_id expect them; without an IUser the
+		// node would be written to the unauthenticated appdata folder instead.
+		$owner = $this->getOwnerOfSignedFile($signRequest);
 		foreach ($files as $fileData) {
 			$dataToSave = $fileData;
 			$dataToSave['signRequest'] = $signRequest;
 			$dataToSave['name'] = $fileData['name'] ?? $fileData['type'];
+			if ($owner instanceof IUser) {
+				$dataToSave['userManager'] = $owner;
+			}
 			$file = $this->requestSignatureService->saveFile($dataToSave);
 
 			$this->idDocsMapper->save($file->getId(), $signRequest->getId(), null, $fileData['type']);
 		}
+	}
+
+	private function getOwnerOfSignedFile(SignRequest $signRequest): ?IUser {
+		$signedFileId = $signRequest->getFileId();
+		if (!$signedFileId) {
+			return null;
+		}
+		try {
+			$signedFile = $this->fileMapper->getById($signedFileId);
+		} catch (\Throwable) {
+			return null;
+		}
+		return $this->userManager->get($signedFile->getUserId());
 	}
 
 	public function list(array $filter, ?int $page = null, ?int $length = null, array $sort = []): array {
--- a/lib/Service/IdDocsService.php
+++ b/lib/Service/IdDocsService.php
@@ -146,7 +146,7 @@ class IdDocsService {
 		}
 		try {
 			$signedFile = $this->fileMapper->getById($signedFileId);
-		} catch (\Throwable) {
+		} catch (\OCP\AppFramework\Db\DoesNotExistException) {
 			return null;
 		}
 		return $this->userManager->get($signedFile->getUserId());
--- a/tests/php/Unit/Service/IdDocsServiceTest.php
+++ b/tests/php/Unit/Service/IdDocsServiceTest.php
@@ -21,6 +21,8 @@ use OCA\Libresign\Service\Validation\FileInputValidator;
 use OCA\Libresign\Service\Validation\IdentityDocumentValidator;
 use OCP\IAppConfig;
 use OCP\IL10N;
+use OCP\IUser;
+use OCP\IUserManager;
 use PHPUnit\Framework\MockObject\MockObject;
 
 /**
@@ -39,6 +41,7 @@ final class IdDocsServiceTest extends \OCA\Libresign\Tests\Unit\TestCase {
 	private RequestSignatureService&MockObject $requestSignatureService;
 	private TimeFactory&MockObject $timeFactory;
 	private IAppConfig&MockObject $appConfig;
+	private IUserManager&MockObject $userManager;
 
 	public function setUp(): void {
 		parent::setUp();
@@ -56,6 +59,7 @@ final class IdDocsServiceTest extends \OCA\Libresign\Tests\Unit\TestCase {
 		$this->requestSignatureService = $this->createMock(RequestSignatureService::class);
 		$this->timeFactory = $this->createMock(TimeFactory::class);
 		$this->appConfig = $this->createMock(IAppConfig::class);
+		$this->userManager = $this->createMock(IUserManager::class);
 	}
 
 	private function getIdDocsService(): IdDocsService {
@@ -71,6 +75,7 @@ final class IdDocsServiceTest extends \OCA\Libresign\Tests\Unit\TestCase {
 			$this->identifyMethodMapper,
 			$this->timeFactory,
 			$this->appConfig,
+			$this->userManager,
 		);
 	}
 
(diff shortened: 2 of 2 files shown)


lib/Service/IdDocsService.php
```diff
@@ -120,16 +122,36 @@ public function addFilesToDocumentFolder(
...
 	}
 
+	private function getOwnerOfSignedFile(SignRequest $signRequest): ?IUser {
+		$signedFileId = $signRequest->getFileId();
+		if (!$signedFileId) {
+			return null;
+		}
+		try {
+			$signedFile = $this->fileMapper->getById($signedFileId);
+		} catch (\Throwable) {
```
vitormattos: Could we catch only `DoesNotExistException` here?

`FileMapper::getById()` documents this exception when the file row does not exist. Catching `Throwable` also hides unexpected database or programming errors and turns them into the fallback behaviour that stores the document in appdata, which can recreate the same inconsistent state this PR is fixing.

The missing file row is an expected condition we can handle, but other failures should probably remain visible.

vitormattos (changes requested):
The fix looks correct overall. I am requesting one change before merge: please narrow the `catch (\Throwable)` to the expected exception from `FileMapper::getById()`, so unexpected failures are not silently converted into the unauthenticated storage fallback.

As a non-blocking follow-up, this bug may be worth an architecture issue. The storage decision needs the correct user context before the node is created, while `saveFile()` can also infer the owner later from the `SignRequest` when setting `File.user_id`. This allowed storage and metadata to disagree.

The amount of setup needed in the new unit tests may be another sign that this responsibility is spread across too many collaborators. It may be worth reviewing this flow separately to see if it can be made harder to misuse and simpler to test.

maia-andre:
Done in 9dd0790: `getOwnerOfSignedFile()` now catches only `DoesNotExistException`, the exception `FileMapper::getById()` documents. Two tests were added: a missing file row keeps the current fallback (no `userManager`, `saveFile()` still called), and an unexpected exception from the mapper is propagated before `saveFile()` is reached — the second one fails against the previous `\Throwable` catch. I used the fully-qualified class name in the `catch` (as `CrlService` does) so the hunk does not sit between `use` lines that differ from the stable branches; the cherry-pick of both commits still applies cleanly to `stable32`–`stable35`.

On the follow-up: agreed that the storage decision and `saveFile()`'s later owner inference are two places deciding the same thing. If you want, I can open the architecture issue with the two code paths and the test setup as evidence.

vitormattos (approved):

PR #8373, opened 2026-09-12
feat: explain when signer has no visible signature field

### LINKED ISSUE #8325: Explain when a signer has no visible signature field
## Context

Part of #8321.

A PDF digital signature does not need a visible signature on the page.

When a signer receives a request without a visible signature field, the LibreSign interface may look incomplete because there is no visible place to sign.

LibreSign should explain that this is expected and that the document can still be signed normally.

## Expected behavior

When the current signer has no visible signature field, show a short informational message in the signing flow.

Suggested content:

> **No visible signature is required**
>
> Your digital signature will still be added to the PDF and can be validated after signing.

Use `NcNoteCard` with an informational style.

Do not add a dialog or require confirmation.

## How to detect this state

The frontend already has the helper:

`hasVisibleElementsForCurrentUser(document)`

in:

`src/services/visibleElementsService.ts`

Reuse this helper.

It already checks the `signRequestId` values that belong to the current signer and matches them with the visible elements in the document.

Do not create another implementation to find the current signer or inspect visible elements.

The message should be shown when:

```ts
!hasVisibleElementsForCurrentUser(signStore.document)
```

and the current user is in a valid signing flow.

`src/views/SignPDF/_partials/Sign.vue` already imports this helper.

## UI placement

Add the message to:

`src/views/SignPDF/_partials/Sign.vue`

This component already uses `NcNoteCard`.

Show the information before the main signing action, where the signer can see it before clicking `Sign document`.

Do not show the same message in more than one place.

Do not add a special message after signing.

## Signing behavior

The message is informational only.

It must not:

* disable or hide `Sign document`;
* require another confirmation;
* create a visible signature field;
* choose a position on the PDF;
* change the signing request;
* change the digital signature;
* suggest that a signature without a visible mark is weaker or invalid.

The normal signing flow must remain unchanged.

## Tests

Update:

`src/tests/views/SignPDF/Sign.spec.ts`

The existing tests already use visible elements with `signRequestId`.

Add cases for:

* current signer has a visible element: message is not shown;
* current signer has no visible element: message is shown;
* another signer has a visible element but the current signer does not: message is shown;
* current signer has a visible element but another signer does not: message is not shown;
* `Sign document` remains available when the message is shown.

Reuse the existing test fixtures and `hasVisibleElementsForCurrentUser()` behavior.

Do not duplicate the visible-element matching logic in the component tests.

## Playwright

Add or update a signing Playwright test with no visible signature position.

Confirm that:

1. the informational message is visible;
2. `Sign document` remains available;
3. the signer can complete the signature successfully.

Keep the existing Playwright coverage for signing with a visible position.

Reuse an existing signing workflow where possible instead of creating a complete new setup.

## Out of scope

This issue does not change:

* the requester warning from #8323;
* the warning preference from #8322;
* the user preferences screen from #8324;
* signature validation;
* the validation page.

## Good first issue

The logic needed to detect this state already exists.

The main work is:

* reuse `hasVisibleElementsForCurrentUser(signStore.document)`;
* add one informational `NcNoteCard` to the existing signing view;
* add focused component tests;
* extend an existing Playwright signing flow.

Do not create new signer-detection or visible-element logic.

Please keep the change focused and reuse the current LibreSign and Nextcloud components.

### Additional context

## Done when

* [ ] The existing `hasVisibleElementsForCurrentUser()` helper is reused.
* [ ] The message is shown only when the current signer has no visible element.
* [ ] `NcNoteCard` is used.
* [ ] No new confirmation step is added.
* [ ] `Sign document` remains available.
* [ ] Signing behavior does not change.
* [ ] Component tests cover visible and non-visible cases.
* [ ] Playwright covers successful signing without a visible field.

nva138:
Resolves: #8325

## 📝 Summary

Shows an NcNoteCard in the signing sidebar when the current signer has no visible signature field, so the page does not look incomplete.

Reuses hasVisibleElementsForCurrentUser(). I could not reuse needCreateSignature, it also returns false when the instance disallows creating signatures or when the user already has one.

The card is a sibling of the existing v-if chain, not part of it, so "Sign document" stays available.

## 🧪 How to test

1. Upload a PDF in LibreSign
2. Add yourself as signer, but do not place a visible signature element
3. Request signatures and open the document to sign
4. The sidebar shows the note card above the "Sign document" button
5. Signing still works normally

```bash
npx vitest run src/tests/views/SignPDF/Sign.spec.ts -t "no visible signature notice"
npx playwright test sign-herself-with-click-to-sign
```

## 🎨 UI / Front-end changes

- [x] Informational note card when the signer has no visible signature field

🏚️ Before | 🏡 After
--- | ---
before | after

- [ ] Tested in multiple browsers (Chrome, Firefox, Safari)
- [x] Components, Unit (with vitest) and/or e2e (with Playwright) tests added
- [ ] Accessibility verified (contrast, keyboard navigation, screen reader friendly)
- [ ] Design review approved
- [ ] Documentation updated (if applicable)

Tests: component tests for both cases in Sign.spec.ts, and extended sign-herself-with-click-to-sign which already signs without a visible field.

One thing worth your review: the card also shows while a non-retriable certificate error blocks signing, so "Sign.vue - API error handling" now expects 2 note cards instead of 1. The issue describes the condition without exceptions so I kept it, but tell me if you want it hidden there.

## ✅ Checklist

- [x] Conventional commits with DCO sign-off
- [x] npm test, npm run lint and npm run ts:check pass

## 🤖 AI (if applicable)

- [x] The content of this PR was partially generated using AI

### DIFF
--- a/src/tests/views/SignPDF/Sign.spec.ts
+++ b/src/tests/views/SignPDF/Sign.spec.ts
@@ -884,7 +884,8 @@ describe('Sign.vue - signWithTokenCode', () => {
 
 			expect(wrapper.text()).toContain('Try signing again')
 			expect(wrapper.text()).not.toContain('Sign document')
-			expect(wrapper.findAll('.nc-note-card-stub')).toHaveLength(1)
+			// One card for the blocking error, one for the missing visible signature field.
+			expect(wrapper.findAll('.nc-note-card-stub')).toHaveLength(2)
 		})
 	})
 
@@ -2265,5 +2266,41 @@ describe('Sign.vue - required device geolocation', () => {
 		expect(wrapper.vm.requiresDeviceGeolocation).toBe(false)
 		expect(geolocationHelper.collectDeviceGeolocation).not.toHaveBeenCalled()
 		expect(submitSignatureMock.mock.calls[0]?.[0]).not.toHaveProperty('geolocation')
+	})
+})
+
+describe('Sign.vue - no visible signature notice', () => {
+	it('detects a visible signature field when the signer has one', async () => {
+		setActivePinia(createPinia())
+		const SignComponent = await import('../../../views/SignPDF/_partials/Sign.vue')
+		const { useSignStore } = await import('../../../store/sign.js')
+		const signStore = useSignStore()
+		signStore.document = createSignDocument({
+			nodeType: 'file',
+			signers: [
+				{ signRequestId: 501, me: true },
+			],
+			visibleElements: [
+				{ elementId: 201, fileId: 1, signRequestId: 501, type: 'signature', coordinates: { page: 1, left: 10, top: 20, width: 30, height: 40 } },
+			],
+		})
+		const wrapper = mount(SignComponent.default, createSignMountOptions())
+		expect(wrapper.vm.hasVisibleSignatureField).toBe(true)
+	})
+
+	it('detects no visible signature field when the signer has none', async () => {
+		setActivePinia(createPinia())
+		const SignComponent = await import('../../../views/SignPDF/_partials/Sign.vue')
+		const { useSignStore } = await import('../../../store/sign.js')
+		const signStore = useSignStore()
+		signStore.document = createSignDocument({
+			nodeType: 'file',
+			signers: [
+				{ signRequestId: 501, me: true },
+			],
+			visibleElements: [],
+		})
+		const wrapper = mount(SignComponent.default, createSignMountOptions())
+		expect(wrapper.vm.hasVisibleSignatureField).toBe(false)
 	})
 })
--- a/src/tests/views/SignPDF/Sign.spec.ts
+++ b/src/tests/views/SignPDF/Sign.spec.ts
@@ -884,8 +884,7 @@ describe('Sign.vue - signWithTokenCode', () => {
 
 			expect(wrapper.text()).toContain('Try signing again')
 			expect(wrapper.text()).not.toContain('Sign document')
-			// One card for the blocking error, one for the missing visible signature field.
-			expect(wrapper.findAll('.nc-note-card-stub')).toHaveLength(2)
+			expect(wrapper.findAll('.nc-note-card-stub')).toHaveLength(1)
 		})
 	})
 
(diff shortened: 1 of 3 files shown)


src/views/SignPDF/_partials/Sign.vue
```diff
@@ -37,6 +37,14 @@
 				<NcRichText :text="error.message"
 					:use-markdown="true" />
 			</NcNoteCard>
+			<NcNoteCard v-if="!hasVisibleSignatureField"
```
vitormattos: Could we avoid showing this notice when signing is blocked by a non-retriable error?

The issue says this message should be shown when the signer has no visible element **and is in a valid signing flow**. In the blocking error state, the user cannot sign at that moment, so showing “Your digital signature will still be added” together with “Signing is blocked” can be confusing.

I think the notice should be hidden when `hasBlockingSignError` is true, and the existing API error test should continue expecting only the blocking error card.

src/tests/views/SignPDF/Sign.spec.ts
```diff
@@ -2267,3 +2268,39 @@ describe('Sign.vue - required device geolocation', () => {
 		expect(submitSignatureMock.mock.calls[0]?.[0]).not.toHaveProperty('geolocation')
 	})
 })
+
+describe('Sign.vue - no visible signature notice', () => {
```
vitormattos, line 2271: Could we cover the behavior requested by the issue at the component level instead of checking only the exposed computed value?

We should verify the rendered notice for these cases:

* another signer has a visible element, but the current signer does not → notice is shown;
* the current signer has a visible element, but another signer does not → notice is not shown;
* when the notice is shown, `Sign document` is still available.

This is important because the matching by `signRequestId` is the main behavior we want to protect here. Checking `wrapper.vm.hasVisibleSignatureField` alone would still pass if the template condition were changed or removed by mistake.

vitormattos (changes requested):

src/tests/views/SignPDF/Sign.spec.ts
nva138: Done. The tests now assert the rendered notice instead of the exposed computed,
using two signers.

- only the other signer has a visible element -> notice is shown
- the current signer has a visible element -> notice is not shown
- while the notice is shown, "Sign document" is still rendered

I also removed hasVisibleSignatureField from defineExpose, since nothing reads
it anymore.

nva138: The notice is now hidden when hasBlockingSignError is true, and the API
error test expects one card again.

vitormattos (approved):

PR #8381, opened 2026-09-13
fix: use the approver's own identity when signing an identification document

### LINKED ISSUE #8365: Approver gets the original (external) signer's identity instead of their own when trying to sign/approve an identification document
## Environment
- LibreSign 14.1.0
- Nextcloud 34.0.4.1

## Related to
This surfaced while working around #2 (identification documents uploaded by unauthenticated signers) in this same install — once the storage bug is fixed and the document becomes viewable, this second, separate bug appears.

## Steps to reproduce
1. Same setup as the "identification documents from unauthenticated signers" scenario: an external signer (identified by email, no Nextcloud account) uploads their identification document as part of signing a document.
2. As an admin in `approval_group`, open LibreSign → "Documents Validation", open the identification document (viewer works fine), and click "Sign the document" to approve it.
3. A "Verificación de correo electrónico" (email verification) modal appears asking to confirm an email address, then click "Send verification code".

## Expected behavior
The admin approves/signs the identification document using their own Nextcloud account identity (they're already authenticated as an authorized approver) — no unrelated third-party verification should be involved.

## Actual behavior
The modal asks to verify the **original document's external signer's email** (not the admin's own identity) — confirmed by computing the MD5 hash of the masked email shown in the UI (`fhe***@***.com.mx`) and matching it against the actual signer's address, which matched exactly.

Clicking "Send verification code" always fails with **"Document not found"**, and — confirmed by inspecting the web server's access log across multiple attempts, in a fresh incognito window, after hard-reloading — **no HTTP request is ever sent to the server**. The error is thrown entirely client-side.

## Root cause

Frontend (`src/views/SignPDF/_partials/ModalVerificationCode.vue`), `requestCode()`:
```js
const signRequestUuid = getCurrentSignerSignRequestUuid(signStore.document)
if (!signRequestUuid) {
    throw new Error(t('libresign', 'Document not found'))
}
```
This looks for a sign-request entry belonging to the *currently logged-in user* on the loaded document. Since the admin is not a signer of the underlying `SignRequest`, this always returns falsy and throws before any network call is made — hence nothing shows up server-side.

Backend root cause, `lib/Service/File/SettingsLoader.php`, `loadApproverSignatureMethods()`:
```php
private function loadApproverSignatureMethods(stdClass $fileData): void {
    try {
        $idDocs = $this->idDocsMapper->getByFileId($fileData->id);
        $signRequestId = $idDocs->getSignRequestId();
        if (!$signRequestId) {
            return;
        }

        $signatureMethods = $this->identifyMethodService->getSignMethodsOfIdentifiedFactors($signRequestId);
        $fileData->settings['signatureMethods'] = $signatureMethods;
    } catch (\Throwable) {
    }
}
```
`$idDocs->getSignRequestId()` returns the `SignRequest` id of the **original document being signed** (the NDA in my test), because — as described in #2 — `IdDocsService::addFilesToDocumentFolder()` never creates a dedicated `SignRequest`/`IdentifyMethod` for the identification document itself; it only links the id-doc's `File` row to the pre-existing `SignRequest` of the main document via `idDocsMapper->save(...)`.

So `getSignMethodsOfIdentifiedFactors($signRequestId)` returns the identify methods of the **original external signer** (in my case, `emailToken` bound to their email), which is what gets surfaced to the approver as "the way to verify yourself" — even though that identity has nothing to do with the approver.

The backend appears aware this can happen: the `file/validate` response for this state includes, alongside `isApprover: true` and `canSign: true`, an info message:
```
"You cannot request signature for this document, please contact your administrator"
```

## Why I haven't attempted a fix
Unlike #2, there is currently no code path that creates a `SignRequest`/`IdentifyMethod` scoped to *the approver* for an identification document uploaded through the unauthenticated-signer flow — `addIdDocs()` (self-service upload) does create one, but it binds `IDENTIFY_ACCOUNT` to the **uploader's own** account, which is the wrong identity for a distinct external reviewer/approver anyway. A correct fix likely needs to lazily create (or otherwise wire up) a `SignRequest`/`IdentifyMethod` for the approving admin at approval time — `SettingsLoader::loadSettings()` is currently a read-only response-building path, so this isn't a one-line change, and I'd rather flag it for maintainer input on the intended design (should approval require a full "sign" cycle at all, or is there meant to be a lighter-weight approve action separate from `SignFileController`'s code-based sign flow?).

## Workaround
None found — identification documents uploaded by unauthenticated external signers currently cannot be approved via the "Sign the document" button. Manual/out-of-band identity verification is needed in the meantime.

maia-andre:
Resolves: #8365

## 📝 Summary

When an approver opens an identification document uploaded by an external signer, the signing step already uses the approver's own account (`SignFileService::getOrCreateApproverSignRequest()`, #6891), but the three steps before it did not:

- `SettingsLoader::loadApproverSignatureMethods()` exposed the signature methods of the sign request stored in the id docs row — the **uploader's** methods. For an external signer that is an `emailToken` bound to their masked email, so the approver was asked to verify `ext***@***.com`.
- `POST /sign/uuid/{uuid}/code` only knew signer uuids; the file uuid carried by the approver route answered *Invalid data to sign file*.
- `ModalVerificationCode.requestCode()` looked for a signer with `me: true`, which an approver never is, and failed locally with *Document not found*.

The fix makes the whole flow follow the rule the signing step already applies, as agreed in the issue (option b — nothing persisted on read):

- `IdentifyMethodService::getSignMethodsOfAccount()` builds the approver's `account` identify method **in memory** and returns its enabled signature methods; `SettingsLoader` uses it for the approver. Viewing the document does not create the approver's sign request.
- `SignFileController::requestCodeBySignerUuid()` accepts the same `idDocApproval=true` context as `sign()`: the uuid is resolved as the identification document (already authorized by the middleware) and the approver's sign request is created or reused before the code is sent.
- The frontend takes one route/context decision for signing and for requesting the code: `isIdDocApprovalContext()` is used by `buildSignUrl()` and by the new `buildRequestCodeUrl()`; the modal uses `getSigningRouteUuid()` like `Sign.vue` does.

## 🧪 How to test

Unit tests use different identities for the uploader (`external@example.com`, `emailToken`) and the approver (account `approver`) and assert that the uploader's methods and sign request never reach the approver flow:

- PHP: `SettingsLoaderTest::testLoadSettingsApproverGetsSignatureMethodsOfTheirOwnAccount`, `IdentifyMethodServiceTest::testGetSignMethodsOfAccountBuildsTheApproverIdentityInMemory` (asserts nothing is saved), `SignFileControllerTest::testRequestCodeBySignerUuidInIdDocApprovalContextUsesTheApproverSignRequest` and `…WithoutIdDocApprovalKeepsTheSignerLookup`.
- Vitest: `signRequestUuid.spec.ts` (`isIdDocApprovalContext`), `sign.spec.ts` (`buildRequestCodeUrl` takes the same decision as `buildSignUrl`), `ModalVerificationCode.spec.ts` (approver → file uuid + context; regular signer → signer uuid; approver who is also a signer → signer uuid).

Local checks: Vitest 102/102 on the three specs; ESLint and php-cs-fixer clean; psalm on the three PHP files: no errors. PHPUnit `--testsuite unit` in the devcontainer: 3810 tests, the 3 touched classes pass; 4 failures in `AEngineHandlerTest::testSetEngineConfiguresIdentifyMethodsForNoneEngine` are order-dependent (the class passes alone on this branch and on `main`, and the diff does not touch that code).

### Manual verification (devcontainer, Nextcloud 36 dev, engine OpenSSL + PhpNative)

Setup: `identify_methods` with `account` (`emailToken` only) and `email` (`emailToken`) enabled; `identification_documents` enabled with `admin` as approver; `admin` has `admin@example.com`. A signature request is sent to `external-…@example.com` (no account), the external signer uploads an identification document from the public sign page, then `admin` opens it from **Documents Validation**.

**Before** (`main`) — the approver is asked to verify the *uploader's* email:

before: approver asked for ext***@***.com

**After** — the approver's own email is shown, the code is requested with `idDocApproval=true`, the email goes to the approver, and the document is signed and approved:

| External signer uploads the id doc | Approver list |
|---|---|
|  |  |

after: approver asked for adm***@***.com

Request logged by the browser: `POST /ocs/v2.php/apps/libresign/api/v1/sign/uuid/{file uuid}/code?idDocApproval=true` → 200 `Verification code sent.`

Mailpit: code sent to admin@example.com

| Identity confirmed | Signed |
|---|---|
|  |  |

approved by admin

`POST /sign/uuid/{file uuid}?async=true&idDocApproval=true` → 200 `File signed`.

## ⚙️ API / Back‑end changes

- [x] `POST /api/v1/sign/uuid/{uuid}/code` accepts the request parameter `idDocApproval=true` (same semantics as the existing `sign()` endpoint); no change to the response.
- [x] New `IdentifyMethodService::getSignMethodsOfAccount(string $uid): array` (read-only).
- [x] Unit tests added (PHP and Vitest).

## 🚧 Backport

None: `main` only, as agreed in the issue (approver flow change, and `stable35` is frozen for features).

## Not covered on purpose

- Whether approving should be a lighter action than signing with the approver's certificate: design kept as is (decision in the issue).

## ✅ Checklist

## 🤖 AI (if applicable)


vitormattos (changes requested):
The implementation looks good to me and follows the direction we agreed in #8365.

One small thing I would improve before finishing this: I think we are adding more explanatory comments than the code needs. In a few places the comments repeat what the method names and the code already express.

For example, in `SignFileController` the important non-obvious detail is that, in the `idDocApproval` context, the UUID represents the **file**, not a SignRequest. The rest is already clear from `getFileByUuid()` and `getSignRequestToSign()`.

The same applies to `loadApproverSignatureMethods()`: I think the important architectural note is that loading settings must not persist/create the approver SignRequest. The method names already explain most of the rest.

Could you please trim the new comments and keep only the context/invariants that cannot be understood from the code itself?

Other than that, the implementation looks good to me.

maia-andre:
Trimmed, thanks. Kept only the two invariants you named: in `SignFileController`, that in the `idDocApproval` context the uuid is the file's, not a sign request's; in `SettingsLoader::loadApproverSignatureMethods()`, that loading settings must not create or persist the approver's sign request (and the matching one-liner on `getSignMethodsOfAccount()`: built in memory, not registered for `save()`). The frontend comments now only point to `buildSignUrl()`/`Sign.vue` as the shared decision. Amended into the same commit and rebased on `main`; no code change.

vitormattos:
The comment cleanup looks good.

I checked the backport targets as well. Since we will still publish one final `stable32` release before closing that branch, this fix should be included there too.

So the backport targets for this PR should be:

* `stable35`
* `stable34`
* `stable33`
* `stable32`, only as part of the final EOL release

After that final `stable32` release, we should stop regular backports to it.

vitormattos (approved):

Issue #8323, opened 2026-09-09
Warn requesters when signers have no visible signature field


vitormattos:
## Context

Part of #8321.

A PDF can be digitally signed without showing a signature on the page.

A visible signature field is optional.

LibreSign already supports this behavior, but the current request flow does not explain it. A requester may think that a visible signature field is required.

LibreSign should explain this before sending the request.

## Current flow

LibreSign already shows a confirmation dialog after the requester clicks `Request signatures`.

The current dialog shows:

> **Confirm**
>
> Send signature request?

Actions:

* `Cancel`
* `Send`

This issue must extend this existing dialog.

Do not add another modal or another confirmation step.

## Expected behavior

Before showing the confirmation, LibreSign must identify which signing participants have no visible signature field.

If every signing participant has at least one visible signature field, keep the current confirmation unchanged.

If one or more signing participants have no visible signature field, show extra information inside the same `Confirm` dialog.

## Normal confirmation

When all signing participants have visible signature fields, keep:

> **Confirm**
>
> Send signature request?

Actions:

* `Cancel`
* `Send`

Do not show any extra visible-signature information.

## Confirmation with missing visible signatures

When one or more signing participants have no visible signature field, keep the same dialog title and main question:

> **Confirm**
>
> Send signature request?

Then show:

> Some signers have no visible signature field.
>
> A PDF can be digitally signed without showing a signature on the page. Their digital signatures will still be added to the PDF and can be validated.

List only the affected signers.

For example:

> **No visible signature:**
>
> * Bob
> * Carol

Keep the same actions:

* `Cancel`
* `Send`

The request must never be blocked only because a visible signature field is missing.

## How to detect affected signers

`RequestSignatureTab.vue` already imports:

`getVisibleElementsFromDocument`

from:

`src/services/visibleElementsService.ts`

Reuse this existing service.

Do not create a new visible-element parser.

The signer data already contains `signRequestId`.

A signer has a visible signature field when there is at least one visible element where:

* `element.signRequestId` matches the signer `signRequestId`;
* `element.type === 'signature'`.

Do not use signer name or email for this match.

Do not count other visible field types as a visible signature.

For example, a signer with a text or date field but no `signature` field must still be treated as having no visible signature field.

## Reusable helper

Prefer adding a small reusable helper to:

`src/services/visibleElementsService.ts`

Suggested responsibility:

```ts
getSignersWithoutVisibleSignatureElements(document, signers)
```

The helper should:

[...]

PR #8382, opened 2026-09-13
feat(preferences): add the visible signature warning switch to user preferences

### LINKED ISSUE #8324: Add the visible signature warning option to user preferences
## Context

Part of #8321.

#8322 adds the user preference used to control the warning about signers without visible signature fields.

Users must be able to change this preference later from the LibreSign preferences screen.

The warning itself is implemented in #8323.

## Expected behavior

Add a setting to the existing LibreSign user preferences screen.

Suggested label:

> **Warn me when requesting signatures without visible fields**

Suggested description:

> Show a warning when one or more signers will sign without a visible signature on the PDF.

Use an existing Nextcloud switch component.

The current value comes from:

`warn_without_visible_signature_fields`

Expected behavior:

* `true`: warning enabled;
* `false`: warning disabled.

The default is `true`.

## Preferences screen

Add the setting to the existing:

`src/views/Preferences/Preferences.vue`

Keep the current LibreSign preferences layout and components.

Do not create:

* a new preferences page;
* a new settings framework;
* a policy definition for this option.

This is a normal user configuration, not a LibreSign policy.

## Saving

Reuse the existing user configuration API from #8322:

`PUT /apps/libresign/api/v1/account/config/{key}`

Use:

`warn_without_visible_signature_fields`

When the user changes the switch, save:

* `1` when enabled;
* `0` when disabled.

The UI must reflect the current saved value when the page is opened.

## Save errors

Follow the existing Preferences UI behavior for save feedback.

If the request fails:

* restore the previous switch value;
* show the normal LibreSign error feedback;
* do not leave the UI showing a value that was not saved.

Do not add a custom error system only for this setting.

## Restore the warning

If the requester selected the option to stop showing the warning in #8323, this switch will be disabled when the preferences page is opened.

Turning the switch on again must save:

`warn_without_visible_signature_fields = 1`

This restores the warning for future affected signature requests.

Do not add a separate button such as:

* `Reset warning`;
* `Clear preference`;
* `Show again`.

The switch is enough.

## Frontend tests

Update:

`src/tests/views/Preferences/Preferences.spec.ts`

Cover at least:

* the setting is rendered;
* default `true` renders the switch enabled;
* saved `false` renders the switch disabled;
* changing from enabled to disabled saves `0`;
* changing from disabled to enabled saves `1`;
* a failed save restores the previous value;
* a failed save shows an error.

Reuse the existing Preferences test setup and mocks.

Do not create a separate test file only for this setting unless the current file becomes difficult to maintain.

## End-to-end coverage

Do not duplicate the complete requester workflow here.

#8323 already covers the Playwright behavior of the warning and the saved preference.

This issue only needs to make sure that the user can change the same preference from the Preferences screen.

If an existing Preferences Playwright test already covers user settings, extend it with a small check for this option.

Otherwise, frontend tests in `Preferences.spec.ts` are enough for this issue, while the cross-flow behavior remains covered by #8323.

## Out of scope

* backend preference storage;
* the requester warning;
* signer-side information;
* validation-page changes;
* a new LibreSign policy.

## Good first issue

The backend preference is implemented in #8322.

The LibreSign Preferences screen and its test file already exist.

The main work is:

* add one switch to the existing Preferences screen;
* load `warn_without_visible_signature_fields`;
* save `1` or `0` through the existing user config API;
* add focused frontend tests.

Please reuse the current LibreSign and Nextcloud components and keep the change small.

### Additional context

## Done when

* [ ] The setting is shown in LibreSign user preferences.
* [ ] The saved value is loaded correctly.
* [ ] The default is enabled.
* [ ] The user can enable or disable the warning.
* [ ] Changes use the existing user config API.
* [ ] Failed saves restore the previous UI state.
* [ ] Frontend tests cover the expected behavior.
* [ ] No new policy or settings framework is introduced.

maia-andre:
Resolves: #8324

## 📝 Summary

Adds a **Signature requests** section to the LibreSign preferences screen with one switch, *Warn me when requesting signatures without visible fields*, and its description.

- It is a plain user configuration (#8322), not a policy: it has no scopes or inheritance, so it is rendered apart from the policy-driven entries of `Preferences.vue` instead of being forced into `realDefinitions`. No new policy, page or settings framework.
- Load: the value comes from the `config` initial state through the existing `userconfig` store (`warn_without_visible_signature_fields`, provided by `AccountService::getConfig()`); the switch is enabled unless the saved value is `0`.
- Save: `userConfigStore.update(key, boolean)` → the existing `PUT /apps/libresign/api/v1/account/config/{key}`, which stores `1` / `0` (`setConfig()` converts the boolean; same path as `files_list_grid_view`). A boolean is sent on purpose: the store and the `config` initial state keep this key as a boolean for the other readers (the requester warning of #8323), and an integer would not pass `setValueString()`.
- Feedback: the screen's autosave status (*Saving your preference… / Preference saved*) is reused through the same keyed helpers. When the request fails, the switch and the store go back to the previous value and the screen's usual error note (*Could not save your preference. Try again.*) is shown inside this section. The error feedback of the screen is now keyed by preference (`errorMessageByKey`, like the autosave state) instead of one shared `errorMessage` rendered in every policy section, so every preference — policies included — shows a failure only in its own section; the switch reuses that mechanism, nothing specific to it.
- Turning the switch on again after the requester chose to stop seeing the warning saves `1` — no extra "reset" button.

## 🧪 How to test

`src/tests/views/Preferences/Preferences.spec.ts` (8 new tests, same setup and stubs; the `userconfig` store and the logger are mocked like the policies store):

- the setting is rendered in its own section, as a `switch`, with the description;
- default renders the switch enabled; saved `false` renders it disabled;
- enabled → disabled saves `false` (stored as `0`) and shows the saved feedback; disabled → enabled saves `true` (`1`);
- a failed save restores the previous value in the switch and in the store, and shows the error note in this section only; a failed policy save likewise shows its note only in that policy's section;
- an emitted value equal to the current one is not saved.

Seven of them fail on `main` (the switch does not exist). Local checks: Vitest 31/31 on the file, ESLint and `vue-tsc` clean.

### Manual verification (devcontainer, Nextcloud 36 dev)

`occ user:setting admin libresign warn_without_visible_signature_fields` deleted beforehand.

| Default (no saved value) | Disabled → `PUT {"value":false}` → stored `"0"` |
|---|---|
|  |  |

Reload keeps the switch disabled. With the request aborted (simulated failure), the switch goes back to its previous state and the note is shown once; enabling again saves `"1"` (`occ user:setting` → `1`):

| Failed save | Enabled again |
|---|---|
|  |  |

No Playwright test added: the cross-flow behaviour stays with #8323, as the issue says.

## ⚙️ API / Back‑end changes

- [x] None. Frontend only; the `userconfig` store typedef declares the key.

## 🚧 Backport

None (`main` only, NC36 — part of #8321).

## ✅ Checklist

## 🤖 AI (if applicable)


src/views/Preferences/Preferences.vue
```diff
@@ -320,6 +368,37 @@ async function undoAutoSaveByKey(policyKey: string): Promise<void> {
 	}
 }
 
+// Plain user configuration (see #8322), not a policy: it has no scopes or
+// inheritance, so it is kept apart from the policy-driven entries above and
+// saved through the user config API. The backend stores it as '1' or '0'
+// and reads anything but '0' as enabled, hence the default of true.
+const WARN_WITHOUT_VISIBLE_SIGNATURE_FIELDS_KEY = 'warn_without_visible_signature_fields'
+const warnWithoutVisibleSignatureFields = ref(userConfigStore.warn_without_visible_signature_fields !== false)
+const warnWithoutVisibleSignatureFieldsError = ref('')
```
vitormattos: Could we avoid adding a separate error state only for this preference?

#8324 asks us to reuse the normal Preferences error feedback and not create a custom error system for this setting.

I see the current shared `errorMessage` is rendered inside each policy section, so using it directly would repeat the same error. But I think this should be fixed or made generic instead of introducing `warnWithoutVisibleSignatureFieldsError` only for this option.

For example, we could render the shared error feedback once outside the policy loop, or use a generic keyed error state that can also be reused by other preferences.

src/views/Preferences/Preferences.vue
```diff
@@ -320,6 +368,37 @@ async function undoAutoSaveByKey(policyKey: string): Promise<void> {
 	}
 }
 
+// Plain user configuration (see #8322), not a policy: it has no scopes or
+// inheritance, so it is kept apart from the policy-driven entries above and
+// saved through the user config API. The backend stores it as '1' or '0'
+// and reads anything but '0' as enabled, hence the default of true.
```
vitormattos: Could we simplify this comment?

Most of this explains backend behavior and implementation details that are already covered by #8322 and may become outdated.

Something short like this should be enough:

// This is a plain user preference, not a policy, so it uses the user config store.

Or we could remove the comment entirely if the code is already clear enough.

vitormattos (changes requested):

maia-andre:
Both done (amended into the commit):

- No error state specific to this option any more. The screen's error feedback is now keyed by preference (`errorMessageByKey` + `errorMessageFor(key)`, the same shape as the autosave state), set by the existing policy save/clear paths and by the switch alike, and each section renders only its own key. That also fixes the duplication you noticed: a failed policy save now shows the note only in that policy's section (new test), and the switch shows it only in its section.
- Comment reduced to the one line you suggested.

Vitest 31/31, ESLint and `vue-tsc` clean; description updated.

vitormattos (approved):

PR #8397, opened 2026-09-13
[stable33] fix: accept string node ids from the Files sidebar when requesting a signature

maia-andre:
Backport of #8367 — replaces #8393, whose cherry-pick dropped the backend commit (conflict) and carried only the frontend one.

Two commits:

1. The bot's frontend commit, unchanged (`serializeRequestFile()` forwards a string `nodeId`, `FileInfo.id: number | string`, tests).
2. A branch-specific backend fix, the same shape as the `stable35` backport (#8395): a private `RequestSignatureController::normalizeNodeId()` turns a non-negative int or its canonical decimal string into an int once, for `file` and each entry of `files`, on both the create and the update entry points, and rejects anything else with the existing *Invalid fileID* error before it can reach a later cast. No `main` architecture brought in (this branch has no `RequestSignatureWorkflowService` / `FileInputValidator`).

Tests: `RequestSignatureControllerTest` — string normalized before `validateNewRequestToFile()` / `save()`, each envelope entry, invalid value rejected with 422 before any service (run locally on this branch: 11 tests OK; php-cs-fixer clean).

**Affectedness of `stable33`, checked before opening this** (as asked in #8367): same path as `stable34` — `@nextcloud/files` 4.0.0, `tab.ts` builds `fileInfo.id` from `node.fileid ?? node.id`, `serializeRequestFile()` only accepted numbers, and on the backend `RequestSignatureService::saveFile()` / `FileService::getNodeFromData()` pass the raw value to `getByNodeId(int)` / `getFileByNodeId(int)` under `strict_types`. Nextcloud 33 is where `Node.id` became the string contract, so the branch is affected; the two frontend regression tests carried by the bot commit fail on `stable33` without the fix.

## 🤖 AI (if applicable)


maia-andre:
Pushed two follow-ups: the specs carried by the bot's frontend commit used helpers of the `main` test file (`loadTabModule`, `getRegisteredTabConfig`) and the `requirement` key of identify methods, which do not exist on this branch (the same reason the bot's own PR fails `test`/`typescript`) — adapted to this branch's helpers, 120/120 locally; and `createSignatureRequest()` now declares its return type, since psalm stopped inferring it through the new normalization step.

vitormattos (approved):

Issue #8162, opened 2026-08-31
Show signature rejection information on the validation page


vitormattos:
Update the validation page to support the signature rejection workflow defined by #7832.

Backend rejection support was added by #8159.

This issue is blocked by #8388, which defines the API visibility and redaction contract required by the validation page.

#8161 is related, but it is not a technical blocker. The validation page can be implemented and tested using backend rejection states directly.

## Goal

Make the validation page correctly represent:

* a visible rejected signer;
* rejection date and comment when available;
* a hidden rejection without discovering or reconstructing it;
* a canceled signing workflow;
* a workflow that continues after one signer rejects;
* the same behavior for normal files and envelopes.

The backend is the source of truth for rejection visibility.

The frontend must only render the representation returned by the backend.

## Backend contract

After #8388, a visible rejection may be returned as:

```json
{
  "status": 3,
  "statusText": "Rejected",
  "rejection": {
    "rejectedAt": "2026-09-09T12:00:00+00:00"
  }
}
```

When a visible comment exists:

```json
{
  "status": 3,
  "statusText": "Rejected",
  "rejection": {
    "rejectedAt": "2026-09-09T12:00:00+00:00",
    "comment": "I do not agree with this document",
    "commentPrivate": false
  }
}
```

For a privileged viewer, a private comment may be returned:

```json
{
  "status": 3,
  "statusText": "Rejected",
  "rejection": {
    "rejectedAt": "2026-09-09T12:00:00+00:00",
    "comment": "Private reason",
    "commentPrivate": true
  }
}
```

When a signer rejected but the current viewer is not allowed to know about the rejection, #8388 defines this redacted representation:

```json
{
  "status": null,
  "statusText": "Not signed"
}
```

In this case:

* there is no `rejection` object;
* `null` does not mean draft;
* `null` does not mean ready to sign;
* `null` is not a new signer workflow status;
* the frontend must not try to determine why the real status was redacted.

## Status constants

Keep frontend constants synchronized with the backend enums.

Add:

```js
FILE_STATUS.CANCELED = 6
SIGN_REQUEST_STATUS.REJECTED = 3
```

These values come from:

* `lib/Enum/FileStatus.php`;
* `lib/Enum/SignRequestStatus.php`.

Do not create another frontend workflow status.

The redacted `status: null` is an API representation and must not be added to `SIGN_REQUEST_STATUS`.

## File status presentation

Update the shared file status helpers so `FILE_STATUS.CANCELED` is a known status.

[...]

PR #8398, opened 2026-09-13
fix: apply the rejection visibility rules to signer status and to the validation API

### LINKED ISSUE #8388: Apply signature rejection visibility rules to signer status and validation API
Backend support for signature rejection was added in #8159, but rejection visibility is not applied consistently to all API responses.

This issue completes the backend contract required by #8162.

## Problem

`SignatureRejectionVisibilityService` already controls when rejection details and rejection comments may be exposed.

However, signer status fields and some response builders can still expose the real rejection state.

For example, `File\SignersLoader` currently:

* does not expose the `rejection` object;
* always exposes the real signer `status`;
* always exposes the real `statusText`.

With `public_status: false`, another signer or an anonymous viewer can therefore still receive:

```text
status: 3
statusText: "Rejected"
```

and discover information that should be hidden.

The validation endpoint also does not currently provide the rejection details required by #8162 to privileged viewers.

## Goal

Create one consistent backend visibility contract for signer status and signature rejection.

The backend must:

* keep the real workflow status unchanged;
* expose rejection information only when allowed;
* prevent hidden rejection from being discovered by comparing signer entries;
* provide a safe presentation status for API consumers;
* apply the same rules to normal files and envelopes;
* keep authorization independent from presentation state.

## Domain status and display status

`SignRequestStatus` remains the source of truth for workflow behavior.

Existing domain values remain:

```text
DRAFT = 0
ABLE_TO_SIGN = 1
SIGNED = 2
REJECTED = 3
```

Do not add privacy or presentation values such as `HIDDEN`, `REDACTED`, or `NOT_SIGNED` to `SignRequestStatus`.

Add a separate API presentation field:

```text
displayStatus
```

The allowed values are exactly:

```text
draft
ready_to_sign
signed
rejected
not_signed
```

`displayStatus` describes what the current viewer is allowed to know.

It is not used for workflow decisions or authorization.

## Normal display status mapping

When signer status is not redacted, map the real signer state as follows:

```text
DRAFT        -> draft
ABLE_TO_SIGN -> ready_to_sign
SIGNED       -> signed
REJECTED     -> rejected
```

When the real status is allowed to be disclosed, keep the existing numeric `status` and `statusText` fields for compatibility.

## Hidden rejection

When a file contains a rejection that the current viewer is not allowed to know about, redacting only the rejected signer is not sufficient.

For example:

```text
signer A -> ready_to_sign
signer B -> not_signed
```

would allow the viewer to identify signer B as the signer whose real status was hidden.

Therefore, when a rejection is hidden:

* every unsigned signer in that file must receive the same public representation;
* signed signers remain unchanged.

For every signer whose real state is:

```text
DRAFT
ABLE_TO_SIGN
REJECTED
```

return:

```json
{
  "displayStatus": "not_signed",
  "statusText": "Not signed"
}
```

Do not include the real numeric `status` for these redacted signer entries.

For signed signers, keep the normal representation, including:

```json
{
  "status": 2,
  "displayStatus": "signed"
}
```

Do not replace a hidden real status with another `SignRequestStatus` value.

## Meaning of `not_signed`

`not_signed` is intentionally less specific than the real workflow state.

It means only:

> This signer has not signed, and the current viewer is not allowed to know the more specific signer state.

It does not mean:

* draft;
* ready to sign;
* rejected;
* unable to sign;
* able to sign.

It is a presentation state only.

`statusText: "Not signed"` is kept as a neutral compatibility value for existing API consumers.

Consumers implementing the new contract should use `displayStatus` for presentation logic.

## Rejection object

Use `SignatureRejectionVisibilityService` as the central place for rejection disclosure rules.

When rejection is visible, the response may include:

```json
{
  "status": 3,
  "statusText": "Rejected",
  "displayStatus": "rejected",
  "rejection": {
    "rejectedAt": "2026-09-09T12:00:00+00:00"
  }
}
```

When a public comment may be exposed:

```json
{
  "status": 3,
  "statusText": "Rejected",
  "displayStatus": "rejected",
  "rejection": {
    "rejectedAt": "2026-09-09T12:00:00+00:00",
    "comment": "I do not agree with this document",
    "commentPrivate": false
  }
}
```

For a privileged viewer, a private comment may be exposed:

```json
{
  "status": 3,
  "statusText": "Rejected",
  "displayStatus": "rejected",
  "rejection": {
    "rejectedAt": "2026-09-09T12:00:00+00:00",
    "comment": "Private reason",
    "commentPrivate": true
  }
}
```

A private comment must never be exposed to another viewer.

`public_status: true` must not override comment privacy.

`show_comment_on_validation: true` must only expose a comment when it is not private.

## Privileged viewers

Privileged viewers are:

* the requester of the signature workflow;
* the signer who rejected, when viewing their own signer entry.

Privilege must come only from trusted backend context.

Reuse the existing resolved response context where possible.

`FileResponseOptions` already carries information such as:

* authenticated user through `getMe()`;
* current `SignRequest` through `getSignRequest()`;
* completed signer identification through `isSignerIdentified()`.

Do not create another viewer identity system.

Do not grant privileged visibility based only on:

* request body fields;
* query parameters;
* email;
* display name;
* document UUID;
* signer UUID without the normal identity flow;
* other client-controlled identifiers.

## Authorization

`displayStatus` must never be used for authorization.

All write operations must continue using the real stored workflow state.

In particular:

* a rejected signer remains `SignRequestStatus::REJECTED`;
* a rejected signer cannot sign again;
* eligible signers can continue when `cancel_workflow: false`;
* `canSign` must use the real state;
* reminders and signer management must use the real state;
* redaction must not modify any entity.

`displayStatus` exists only for API presentation.

## Disclosure boundary

When rejection is hidden, do not expose it through:

* real signer `status`;
* rejection-specific `statusText`;
* `rejection`;
* rejection timestamp;
* rejection comment;
* rejection privacy flag;
* rejection-specific labels;
* signer summaries;
* differences between unsigned signer states.

Do not change unrelated signer privacy.

Information already normally visible, such as signer identity, remains governed by existing rules.

## Response isolation

Visibility is viewer-specific.

Do not store viewer-specific values in:

* `SignRequest` metadata;
* file metadata;
* policy snapshots;
* database state.

Build `displayStatus` and rejection visibility while preparing the response.

A response generated for a privileged viewer must never be reused for an anonymous or non-privileged viewer.

If responses are cached, the authorization context must be part of the cache boundary.

## Response contracts

Update backend response definitions to include the presentation state.

Create a shared response type equivalent to:

```text
SignerDisplayStatus =
  draft
  ready_to_sign
  signed
  rejected
  not_signed
```

Review at least:

```text
LibresignSignerSummary
LibresignSignerDetail
LibresignValidatedChildSigner
```

For response contracts that may be redacted:

* `displayStatus` is required;
* numeric `status` is optional because it must be absent when the real unsigned state is redacted;
* `statusText` remains available as a neutral compatibility string.

Generated frontend TypeScript types must remain synchronized with the backend definitions.

Do not make database entities or `SignRequestStatus` optional or nullable.

Do not add `displayStatus` to persistence entities.

## Response paths

Review all backend paths that serialize or summarize signer state.

At minimum:

* `File\SignersLoader`;
* `FileListService`;
* `FileService`;
* `FileService::mapSignerDetailsToSummary()`;
* `EnvelopeAssembler`.

Do not fix only the validation endpoint while another API response continues exposing a hidden rejection.

## Validation response

The validation response must support:

* requester;
* rejecting signer after successful signer identification;
* another authenticated signer;
* anonymous viewer.

When rejection is visible:

```json
{
  "status": 3,
  "statusText": "Rejected",
  "displayStatus": "rejected",
  "rejection": {
    "rejectedAt": "..."
  }
}
```

When rejection is hidden, all unsigned signers of that file receive:

```json
{
  "displayStatus": "not_signed",
  "statusText": "Not signed"
}
```

The real numeric `status` must be absent from those redacted entries.

## Detailed file responses

Apply the same visibility contract to detailed file responses.

This includes an authenticated signer who is not the requester.

For the same viewer and same file, validation and detailed-file responses must not reveal contradictory information.

Existing file-list/sidebar consumers must continue to receive a neutral `statusText` for a redacted signer so the backend change does not introduce an unrelated UI regression.

## FileService summaries

`FileService::mapSignerDetailsToSummary()` currently normalizes signer statuses separately.

Update this path so it does not silently convert `REJECTED` to `DRAFT`.

When rejection is visible, preserve the real state and matching `displayStatus`.

When rejection is hidden:

* all unsigned signer summaries use `displayStatus: "not_signed"`;
* their numeric `status` is omitted;
* their neutral `statusText` is `"Not signed"`;
* signed signer summaries keep their real status and use `displayStatus: "signed"`.

## Envelopes

Apply the same contract to envelopes.

For every child file:

* evaluate rejection visibility for that file;
* if rejection is hidden, all unsigned signers of that child file use `not_signed`;
* omit their real numeric status;
* signed signers remain unchanged;
* do not allow comparison between signer summaries to identify who rejected.

Do not create envelope-specific privacy rules.

## Canceled workflow

`FileStatus::CANCELED` may remain visible.

When a workflow is canceled and rejection is hidden:

* viewers may know the workflow is canceled;
* they must not be able to determine which signer rejected;
* every unsigned signer uses `displayStatus: "not_signed"`;
* the real status of those unsigned signers is not exposed;
* signed signers remain `signed`.

The file-level canceled state does not authorize disclosure of the rejecting signer.

## Shared implementation

Keep the visibility decision centralized in `SignatureRejectionVisibilityService` or another shared rejection visibility abstraction.

The shared logic should determine at least:

* whether rejection is visible;
* which rejection information may be exposed;
* whether unsigned signer states must be redacted;
* which `displayStatus` should be returned;
* whether the numeric status may be exposed.

Do not duplicate these decisions in each serializer.

## Tests

This behavior must be covered by automated tests.

### Display status mapping

Test every normal mapping:

```text
DRAFT        -> draft
ABLE_TO_SIGN -> ready_to_sign
SIGNED       -> signed
REJECTED     -> rejected
```

when the real state may be disclosed.

Test `not_signed` separately as a viewer-specific redacted presentation.

Reject unsupported `displayStatus` values in generated/frontend contracts.

### Hidden rejection

With `public_status: false`:

* a file with one rejected and one pending signer returns `not_signed` for both unsigned signers;
* a file with one rejected, one draft, and one pending signer returns `not_signed` for all three;
* redacted unsigned signers do not contain numeric `status`;
* redacted unsigned signers receive neutral `statusText: "Not signed"`;
* signed signers retain their real `SIGNED` status and `displayStatus: "signed"`;
* the rejected signer cannot be identified by comparing responses;
* no redacted signer exposes `"Rejected"`;
* no redacted signer exposes `rejection`.

### Visible rejection

With rejection visible:

* rejected signer returns `displayStatus: "rejected"`;
* real `status` remains `REJECTED`;
* requester receives rejection details;
* rejecting signer receives their own rejection details;
* another viewer receives only information permitted by policy.

### Comments

Test:

* visible public comment;
* hidden comment when `show_comment_on_validation: false`;
* private comment visible to privileged viewer;
* private comment hidden from another viewer;
* `public_status: true` does not override `commentPrivate`.

### Authorization

Test that:

* `displayStatus` does not change `canSign`;
* absence of numeric `status` from a response does not affect backend authorization;
* rejected signer cannot sign again;
* another eligible signer can sign when `cancel_workflow: false`;
* client-provided identity fields cannot grant privileged visibility;
* document UUID alone does not grant privilege;
* signer UUID alone does not bypass normal signer identification.

### Response isolation

Test sequential requests for the same file:

1. requester;
2. anonymous viewer;
3. rejecting signer;
4. another signer.

Verify that each receives the correct viewer-specific response.

Generating one response must not modify:

* persisted signer state;
* file metadata;
* later responses for another viewer.

### Response consistency

Cover:

* validation response;
* detailed file response;
* detailed file response as another signer;
* single-file summary;
* envelope child response.

The same viewer must receive equivalent visibility semantics across all paths.

### Workflow

Cover both:

```text
cancel_workflow: false
cancel_workflow: true
```

For continued workflow:

* real eligible signers remain able to sign;
* public presentation may still be `not_signed` when needed for privacy.

For canceled workflow:

* file may expose `CANCELED`;
* all unsigned signers use the same redacted presentation when rejection is hidden;
* rejecting signer cannot be identified.

## Test levels

Use Behat/integration tests for observable API and authorization behavior.

Add explicit validation coverage for:

* anonymous validation;
* identified signer UUID validation;
* requester;
* another authenticated signer.

Use PHPUnit for:

* display status mapping;
* rejection visibility decisions;
* serializer behavior;
* response isolation where appropriate.

Use data providers for mapping and viewer/policy matrices where useful.

## Mutation testing

Changed visibility and display-status business logic must be covered by Infection.

For isolated mapping and visibility methods:

* no escaped mutants;
* no mutation errors;
* target 100% MSI.

Do not weaken Infection configuration or exclude production code only to make tests pass.

## Quality gates

The implementation must pass:

* PHPUnit;
* Behat/integration tests;
* Psalm;
* PHPCS;
* Infection.

## Out of scope

* validation page UI;
* signer Reject UI;
* requester configuration UI;
* Policy Workbench UI.

The validation frontend is handled by #8162.

maia-andre:
Resolves: #8388

## 📝 Summary

Backend contract for rejection visibility, as specified in #8388 after the discussion in #8162 (option (a): redact **every unsigned signer** of the file when the viewer may not know about a rejection, so nobody is singled out by comparison).

**One shared decision** — `SignatureRejectionVisibilityService`:
- `hasHiddenRejection(file, signers, privilegedSignRequestIds)`: whether the file holds a rejection the viewer may not know about. Privileged = the requester of the file (for every entry) and the signer who rejected (for their own entry); everybody else only sees it when `public_status` is true.
- `presentSigner(signer, file, privileged, hiddenRejectionInFile)` → `SignerPresentation` (`displayStatus`, real `status` or `null`, `statusText`, `rejection` under the existing comment rules). With a hidden rejection every unsigned signer (`DRAFT`, `ABLE_TO_SIGN`, `REJECTED`) becomes `not_signed` / *Not signed* **without** the real status; signed signers keep `signed`; the viewer's own **pending** entry is redacted like the others (a real status there would tell them who rejected by comparison), and the only entry a viewer keeps is their own **rejection**, which they already know about.
- `SignerDisplayStatus` enum: `draft | ready_to_sign | signed | rejected | not_signed` — presentation only, not persisted, never read for authorization. `SignRequestStatus` is untouched.

**Every response path consumes it** — no serializer reconstructs the rules:
- `File\SignersLoader` (validation endpoint) — gains the `rejection` object it never had, and `FileResponseOptions::isViewerOfSigner()` (the identity check the loader used to do inline: identify method of the session, uid or e-mail of the authenticated user) so the envelope path resolves the viewer the same way;
- `FileListService` — detailed file, file with children (authenticated and anonymous) and the child summaries; the child summary used to encode `status: 1` for a signed signer and `0` otherwise, it now carries the real state and `displayStatus` like the contract says;
- `EnvelopeAssembler` — child documents, same rules;
- `FileService::mapSignerDetailsToSummary()` — passes the presentation through instead of mapping `REJECTED` (and anything unknown) to `DRAFT`.

**Contracts:** `LibresignSignerDisplayStatus`; `LibresignSignerSummary` (and `LibresignSignerDetail` / `LibresignValidatedChildSigner` built on it) declare `displayStatus` as required and `status?: 0|1|2|3` as optional. Generated schema: `SignerDisplayStatus` enum, `status` no longer required; TypeScript `displayStatus: components["schemas"]["SignerDisplayStatus"]`, `status?: 0 | 1 | 2 | 3`.

The `buildSignerRejection()` comment branch was folded into one return with identical behaviour (it was an equivalent mutant).

## 🧪 How to test

**Behat** (`tests/integration/features/sign/reject.feature`, 3 new scenarios, 90 steps, run locally):
- *private status*: requester (list + validation: real state, `rejection.comment`), the rejecter (own entry `rejected`, the other `ready_to_sign`), the other signer (every unsigned entry `not_signed`, their own included — no entry differs from `not_signed` — no `status`, no `rejection`, on validation, on the list and on the child summary), an authenticated bystander and an anonymous reader (`[.signers[].displayStatus] | unique == ["not_signed"]`, no `status`, no `rejection`); then the rejected signer still cannot reject/sign again and the requester still sees `3` — redaction is presentation only;
- *public status*: anonymous gets `rejected` / `3` / `rejectedAt` and no comment (`show_comment_on_validation: false`), the requester gets the comment;
- *canceled workflow*: anonymous gets `file.status: 6` and every signer `not_signed` without `status`; the requester sees who rejected.

**PHPUnit** (all run locally): `SignatureRejectionVisibilityServiceTest` (26 — hidden decision incl. privileged/public/multiple rejections, the four mappings, redaction of each unsigned state, signed never redacted, own pending entry redacted, own rejection kept, rejection object), `SignerDisplayStatusTest`, `SignerPresentationTest`, `FileResponseOptionsTest` (viewer identity), and the viewer matrix — anonymous, another user, the pending signer, the requester, the rejecter — on `SignersLoaderTest`, `FileListServiceTest` (detailed, with children, child summary) and `EnvelopeAssemblerTest`; `FileServiceTest` summaries (visible rejection stays `3`, redacted entry has no status, unknown status is no longer `DRAFT`).

**Infection** on `SignatureRejectionVisibilityService`, `SignerPresentation` and `SignerDisplayStatus`: 68 mutants, 68 killed, **MSI 100%**, no errors.

Full unit suite: 3857 tests; the only failures are the 4 order-dependent ones in `AEngineHandlerTest` that also fail on `main`. php-cs-fixer clean; psalm on the changed files: no new errors. OpenAPI and TypeScript types regenerated.

## ⚙️ API / Back‑end changes

- [x] New response field `displayStatus` on every signer entry; numeric `status` becomes optional (absent for a redacted entry); new neutral `statusText` "Not signed"
- [x] Validation endpoint now returns `rejection` to privileged viewers
- [x] Child file summaries carry the real signer status (`2` for signed) instead of `1`/`0`
- [x] `FileResponseOptions::isViewerOfSigner()`; `SignatureRejectionVisibilityService` gains `IL10N`

**Frontend:** the regenerated `SignerSummaryRecord` requires `displayStatus`, so the three places that build such records locally (`VisibleElements.vue`, `SignPDF.vue` — `draft` placeholders — and `normalizeCertificateSigner()` in `validationDocument.ts`, `signed`/`draft` from the certificate status) and five specs set it; no behaviour change there. One behaviour change in `src/store/sign.js`: the sign page's `ableToSign` gate used to require the real `status` on the viewer's own entry, which is now redacted while a rejection is hidden, so an own entry without `status` no longer blocks signing — the backend keeps enforcing the signing order ("You are not allowed to sign this document yet"). Vitest (incl. a new store case), `vue-tsc` and ESLint clean.

## Note for #8162

The validation frontend is out of scope here, as the issue says. Two facts for whoever picks it up (me, next): `validationDocument.ts` accepts only signer `status` `0|1|2`, so on `main` the validation page already fails to parse any file with a rejected signer (`3`) — this PR does not change that (redacted entries omit `status`, which the parser rejects the same way); and the list/sidebar components read `statusText`, so they show *Not signed* for a redacted entry without further change.

## 🚧 Backport

None (`main` only; the rejection feature is not released).

## ✅ Checklist

## 🤖 AI (if applicable)


lib/Service/SignatureRejection/SignatureRejectionVisibilityService.php
```diff
@@ -27,9 +29,69 @@
...
+	 * @param bool $hiddenRejectionInFile the result of hasHiddenRejection() for the file
+	 */
+	public function presentSigner(
+		SignRequestEntity $signer,
+		?FileEntity $libreSignFile,
+		bool $privileged,
+		bool $hiddenRejectionInFile,
+	): SignerPresentation {
+		$status = $signer->getStatusEnum();
+		if ($hiddenRejectionInFile && $status !== SignRequestStatus::SIGNED && !$privileged) {
```
vitormattos: I think this `!$privileged` condition does not follow the rule defined in #8388.

When a rejection is hidden, all unsigned signers must have the same `not_signed` presentation. Otherwise it is still possible to find who rejected by comparing the signers.

For example, if signer A rejected and signer B can still sign, B currently sees:

* A as `not_signed`
* B as `ready_to_sign`

Since B knows that they did not reject, they can know that A rejected.

The privilege to see your own rejection should be separate from the rule that hides the status of all unsigned signers.

Please change this so a pending signer does not bypass the `not_signed` redaction, while the signer who rejected can still see their own rejection details. The PHPUnit and Behat tests that currently expect the pending signer to stay as `ready_to_sign` also need to be updated.

vitormattos (changes requested):

lib/Service/SignatureRejection/SignatureRejectionVisibilityService.php
maia-andre: Done in 54ea5d2: being the signer no longer lifts the redaction of a pending entry. `presentSigner()` keeps only the viewer's **own rejection**; with a hidden rejection every other unsigned entry, the viewer's own pending one included, is `not_signed` without the real status. The PHPUnit matrix (visibility service, `SignersLoader`, `FileListService`, `EnvelopeAssembler`) and the Behat scenario were updated — for the other signer the scenario now asserts that **no** entry differs from `not_signed` on the validation payload, the file list and the child summary. Infection on the service stays at 100% MSI.

One consequence I had to handle in the same commit: the sign page (`/p/sign/{uuid}`) feeds `useSignStore().ableToSign` from these same signers, and that gate required `mySigner.status === ABLE_TO_SIGN`. With the own entry redacted, the pending signer would get "Unable to sign." whenever another signer had rejected privately. The store now treats an own entry without `status` as signable and leaves the signing order to the backend, which already enforces it (`SignerValidator`: "You are not allowed to sign this document yet"). Residual corner: sequential flow + private rejection + viewer not yet on turn → the button shows and the backend answers with that message instead of the pre-emptive "Unable to sign.". If you want that closed too, the natural place is `settings.canSign` reflecting the viewer's own real state (today it is `me && !signed`, which also feeds "You need to sign this document") — I left it out to keep this PR to the contract.

One more observation, not blocking: even with every unsigned entry redacted, a viewer who knows the behaviour can still infer that *a* rejection exists, because their own entry reads `not_signed` instead of `ready_to_sign` only when something is hidden — with two signers that identifies the rejecter. Closing it fully would mean presenting unsigned signers as `not_signed` whenever `public_status` is false, rejection or not. That is a product decision, so I did not touch it; happy to do it here or in a follow-up if you want.

vitormattos: The implementation here is consistent with #8388, but the discussion exposed a broader product and architecture question around rejection behavior.

Before merging this PR, I would like to define that contract in #8405, especially around:

* whether rejection cancels or continues the workflow;
* who can see the rejection during the workflow;
* public validation visibility;
* comment visibility;
* how admin policy and requester choices interact;
* separation between `SignRequestStatus`, `displayStatus`, and `canSign`.

This is not a problem with the implementation in this PR. The contract itself may need to change before the feature is released.

I created #8405 so we can define the expected behavior and decide whether this PR should be adapted or merged as-is.

It would be useful if you could review the proposal there as well.

vitormattos: I added a first version of the proposal to #8405, so I think we already have enough there to start reviewing the model properly.

The problem turned out to be a bit more complex than I first expected, especially around workflow behavior, visibility and policy inheritance.

Your feedback there is welcome.

vitormattos:
@maia-andre could you make a rebase to solve conflicts? I think that you will need to run again the command `composer openapi` and also to fix other conflicts after last update of `main` branch.

vitormattos (changes requested):
After reviewing this again with the decisions from #8405, I think most of the presentation work here can still be kept.

The main changes will be to make the visibility service use the new policy model instead of `public_status`, and to remove the frontend fallback where a missing signer status is treated as signable. `settings.canSign` should handle that.

I think it makes sense to keep this PR open and adapt it after the policy model is in place.

For now, the rebase is the first step so we can get the branch back in sync with `main`.

maia-andre:
Rebased on `main` (`d9f5b7cd8`) and `composer openapi` re-run; the generated spec now only differs from `main` by `SignerDisplayStatus` and the `displayStatus` / optional `status` fields of `LibresignSignerSummary`.

The conflicts came from #8143 (observer role) and #8412 (ValidateHelper removal). One of them needed a decision rather than a merge: an observer has `signed = null`, so the "every unsigned signer becomes `not_signed` while a rejection is hidden" rule would have redacted observers too, and `SignerDisplayStatus::fromSignRequestStatus()` had no case for `OBSERVING`. I added `SignerDisplayStatus::OBSERVING` (`observing`) and observers are never redacted, for the same reason a signed signer is not: they could not have rejected, so their real state does not point at anyone. Covered in `SignerDisplayStatusTest` and `SignatureRejectionVisibilityServiceTest`.

Everything else is unchanged and stays as agreed in #8405: once the policy model lands, the visibility service reads the effective `rejection.visibility` instead of `public_status`, and the frontend fallback that treats a missing signer status as signable goes away in favour of `settings.canSign`.

vitormattos:
I reviewed the changes after the rebase.

The `OBSERVING` handling makes sense. Since an observer cannot reject, keeping its real status visible does not expose who rejected.

And yes, keeping the other parts as they are for now is fine. We can adapt `public_status` to the new visibility policy and replace the frontend fallback with `settings.canSign` when the corresponding work from #8405 lands.

maia-andre:
Thanks for the review.

Just to make the state explicit: the branch is at `14aff2480`, rebased on `main`, and all checks are green. The only thing holding the merge is the earlier `CHANGES_REQUESTED` review, which is still recorded against the pre-rebase commits — GitHub does not let me request a re-review from a fork, so this comment is the way to ask for it.

Nothing else is pending from my side here. I am starting the policy model PR for #8405 (the five scalar `rejection.*` keys); the follow-up that adapts this PR — visibility service reading the new policy and the frontend fallback replaced by `settings.canSign` — comes after that one lands.

maia-andre:
Rebased on current `main` (`bcfef8424`); the branch is now at `31706f67b`. The rebase was clean — no conflicts, including in `reject.feature`, which `main` touched meanwhile — and `composer openapi` produces no diff, so the generated spec is still in sync.

The previous green run was from 16 September, against a `main` that has moved 296 commits since, so this push re-runs the checks against today's `main` rather than leaving a week-old result behind.

Verified locally before pushing: PHPUnit for the eight test classes this PR touches (340 tests), `reject.feature` 25/25, `php-cs-fixer`, `psalm` and `npm run ts:check`, all clean. `file/list.feature:45` fails in my devcontainer, but it fails the same way on an untouched `main` — it compares `<BASE_URL>` against the local `overwritehost`, so it is environmental and not related to this PR.

Nothing is pending from my side. The only thing holding the merge is the `CHANGES_REQUESTED` review from 16 September, which is still recorded against the pre-rebase commits and which you have since resolved in the comments above; a fork author cannot request a re-review through the API, so this comment is the way to ask for it.

One note on ordering, since both are open: this PR and #8423 both touch `SignatureRejectionVisibilityService`, its test, `ResponseDefinitions`, `reject.feature` and the generated specs. Whichever lands second will need a rebase with real conflict resolution. This one is the smaller of the two and its content is already agreed, so merging it first keeps that work to a minimum — and the follow-up that adapts it to the new policy model comes after #8423, as planned.

vitormattos (changes requested):
#8423 has been merged, so this branch now needs a rebase on current `main`.

After that, this PR still needs the planned adaptation to the new rejection visibility policy and to `settings.canSign` once the behavior/capability work from #8405 is in place.

PR #8406, opened 2026-09-15
fix(listener): notify callback based on signedNodeId/signedHash

henmohr:
## Problem

The webhook/callback configured via the `callback` field of `request-signature` was never fired after the document was signed.

## Root cause

`SignedCallbackListener::handle()` checked for a `signed` key in the entity's updated fields:

```php
if (isset($updatedFields['signed']) && $updatedFields['signed'] === true) {
    $this->signFileService->notifyCallback($event->getSignedFile());
}
```

However, the `File` entity does not have a `signed` field. The signature is tracked by `signedNodeId` and `signedHash`, which are set in `SignFileService::updateLibreSignFile()` (via `setSignedNodeId`/`setSignedHash`). As a result the condition was always false and the callback was never notified.

## Fix

Check `signedNodeId`/ `signedHash` instead:

```php
if (!empty($updatedFields['signedNodeId']) || !empty($updatedFields['signedHash'])) {
    $this->signFileService->notifyCallback($event->getSignedFile());
}
```

## Testing

Verified against Nextcloud 35 + LibreSign `stable35`: after signing, the callback URL now receives the multipart POST (uuid, status, file) automatically. Previously it required a manual invocation of the callback endpoint.


vitormattos (changes requested):
I think we should stop this PR here for now and first review this callback mechanism.

There is no regression test reproducing the reported problem, so we don't have a test showing the current behavior and what this change is expected to fix.

The PR is also targeting `stable35`, while our normal flow is to fix bugs in `main` first and backport them later when needed.

But before asking for these changes, I think there is a bigger question we need to answer. This callback mechanism is old and has almost no coverage. We need to check if it still makes sense to maintain it, or if this should be replaced by a better integration using the event/webhook mechanisms available in Nextcloud.

I will keep this PR as a draft while we investigate this. If we decide to keep the callback, we can continue from here with a regression test and a branch based on `main`. If we decide to replace or remove it, we can close this PR without spending more time on this implementation.

PR #8416, opened 2026-09-17
fix(migration): stop deleting LibreSign appdata on every update

### LINKED ISSUE #8415: The background signature image is resetted after updating
### Describe the bug

After each update, I need to manually re-add my custom background image (under the signature)

### To reproduce

- I'm an admin
- I have an old version installed
- I upload a custom background image in the admin settings -> it works and is used
- I upgrade LibreSign
- The custom background image is gone, no warnings

### Expected behavior

The customisation is not lost when upgrading

### Screenshots

_No response_

### Operational System

Docker, Debian 13

### Browser Type and Version

_No response_

### Libresign Version

14.1.0

### Nextcloud Server Version

34.0.4

Didn't check

_No response_

### Additional context

_No response_

ciaens:
Resolves: #8415

## 📝 Summary

`DeleteOldBinaries` is a `post-migration` repair step, so it runs on **every** app update. It walked the LibreSign AppData folder and deleted every node that was not in an allow list of things to keep. That list only contained the downloaded binaries and a few config folders, so everything else the app stores in AppData was silently removed on each update — including `signature/background.png`, the custom background uploaded by the admin (#8415). Since the background type stays `custom` while the file is gone, `getImage()` falls back to the default logo and the admin just sees the default background come back, with no warning.

Following review, the logic is now **inverted**: the step knows the binary paths it is responsible for removing, and never decides based on what it does *not* know. An incomplete list of obsolete paths only leaves files behind; an incomplete allow list destroys data belonging to the admin and to the signers.

What it deletes:

| Path | Reason |
| --- | --- |
| `libresign-cli/`, `java/`, `jsignpdf/`, `pdftk/`, `cfssl/` at the AppData root | Layout used before the binaries were split by architecture |
| `{arch}/jsignpdf/jsignpdf-*` other than `jsignpdf-{JSignPdfRelease::VERSION}` | Previous JSignPdf releases (e.g. `jsignpdf-2.3.0`) |
| `{arch}/{distro}/java/jdk-*` other than `jdk-{InstallService::JAVA_URL_PATH_NAME}-jre` | Previous Java releases (e.g. `jdk-21.0.2+13-jre`) |

for `{arch}` in `x86_64`, `aarch64` and `{distro}` in `linux`, `alpine-linux`.

The current release is read from `JSignPdfRelease::VERSION` and `InstallService::JAVA_URL_PATH_NAME` instead of being hardcoded, so bumping a dependency cleans up the previous one automatically and this list does not have to be maintained in parallel. Inside those folders only nodes named like a release of that dependency are considered, so anything else is left alone. `JAVA_URL_PATH_NAME` was made `public` for this, next to the already public `JAVA_VERSION` and `JSIGNPDF_VERSION`.

### A note on `java-11`

The review listed `x86_64/linux/java-11/` as a cleanup target. That directory never existed: LibreSign has only ever installed Java 21 (`JAVA_URL_PATH_NAME` went `21.0.2+13` → `21.0.5+11` → `21.0.6+7` → `21.0.7+6` → `21.0.8+9`), there is no `java-11`, `jdk-11` or `temurin11` in any commit, and no dynamic `'java-' . $version` naming. The Java directory has always been `{arch}/{distro}/java/`, with the JRE extracted to `jdk-<version>-jre` **inside** it. So the real leftovers are the outdated `jdk-*-jre` directories, and they are handled above for both architectures and both distributions rather than for `x86_64/linux` only.

`jsignpdf-2.3.0` was confirmed: `JSIGNPDF_VERSION` was `'2.3.0'` and is now `JSignPdfRelease::VERSION = '3.1.0'`.

## 🧪 How to test

1. On an installed LibreSign, go to the admin settings and upload a custom background image under the signature.
2. Check that the file exists: `ls data/appdata_*/libresign/signature/background.png`.
3. Run the repair steps the way an update does: `occ maintenance:repair` (or `occ app:update libresign`).
4. Before this PR the file is gone and the admin settings show the default background again. With this PR the file is still there and the custom background is kept.

The same applies to `data/appdata_*/libresign/certificate-policy.pdf`, to `data/appdata_*/libresign/guest_app/` (files of signers of the `guest_app` group, which are stored in AppData instead of a home folder) and to `generated_crl/`.

To check the cleanup side, create `data/appdata_*/libresign/x86_64/jsignpdf/jsignpdf-2.3.0/` and a `data/appdata_*/libresign/java/` folder and confirm both are removed while `jsignpdf-3.1.0/` and everything else stays.

## ⚙️ API / Back‑end changes

- [x] `DeleteOldBinaries` deletes only explicitly known obsolete binaries and never removes an AppData entry just because it does not know it
- [x] Outdated releases are derived from the version currently installed, so a dependency bump does not require updating this step
- [x] Unit and/or integration tests added – *required for backend changes*<br>`tests/php/Unit/Migration/DeleteOldBinariesTest.php` covers the cleanup contract: known obsolete paths are deleted, current binary paths are kept, non-release files inside the binary folders are kept, and unknown AppData entries (`signature`, `certificate-policy.pdf`, `guest_app`, `generated_crl`, `jsignpdf_home`, `future-feature`) are never deleted.
- [ ] Capabilities updated (if applicable) – not applicable
- [ ] Documentation updated (if applicable) – not applicable
- [ ] API documentation updated with the command `composer openapi` if necessary – no API change

## ✅ Checklist

- [ ] The tests were not run locally (this clone is not inside a Nextcloud server checkout, which `tests/php/bootstrap.php` requires) — relying on CI.

## 🤖 AI (if applicable)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ciaens:
I looked at the CI logs and it's not related to the PR (network errors: empty reply / failed downloads)

lib/Migration/DeleteOldBinaries.php
```diff
@@ -23,7 +23,18 @@
...
+	 *
+	 * This repair step runs on every update and deletes each node that isn't
+	 * listed here, so a new entry stored in appdata MUST also be added to this
+	 * list, otherwise it is silently removed when the admin updates the app.
+	 *
+	 * A string is the name of a node to keep as is, an array is a folder whose
+	 * content is filtered by the same rules.
+	 */
 	protected array $allowedFiles = [
+		// Binaries downloaded by InstallService, one folder per architecture
```
vitormattos: I think we should avoid growing this list with all LibreSign AppData entries.

This class should only know the old binary paths that it is responsible for removing. Other AppData entries should not need to be registered here only to survive an update.

Could we invert this logic and keep an explicit list of obsolete binary paths to delete instead?

tests/php/Unit/Migration/DeleteOldBinariesTest.php
```diff
@@ -0,0 +1,93 @@
...
+
+use OCA\Libresign\Migration\DeleteOldBinaries;
+use OCP\Files\AppData\IAppDataFactory;
+use OCP\Files\File;
+use OCP\Files\Folder;
+use OCP\Files\IAppData;
+use PHPUnit\Framework\TestCase;
+
+final class DeleteOldBinariesTest extends TestCase {
+	public function testKeepsDataStoredByLibresignInAppData(): void {
```
vitormattos: With the cleanup logic inverted, I think this test should also change.

We should not need to list all LibreSign AppData entries that must survive. Instead, please add a test with an arbitrary unknown entry, for example `future-feature`, and assert that it is never deleted.

The tests should focus on the cleanup contract:

* explicitly known obsolete paths are deleted;
* an unknown AppData entry is never deleted;
* current binary paths are not deleted.

This way a future feature can add a new AppData folder without also having to update this test.

vitormattos (changes requested):
I think we can fix this in a simpler and safer way.

`DeleteOldBinaries` should only delete paths that we explicitly know are obsolete. It should not keep a list of everything that LibreSign is allowed to store in AppData.

For this PR, these are the confirmed cleanup targets:

| Path to delete                     | Reason                                                                                                                |
| ---------------------------------- | --------------------------------------------------------------------------------------------------------------------- |
| `java/`                            | Old Java location at the AppData root. Java is now stored inside the architecture and Linux distribution directories. |
| `x86_64/linux/java-11/`            | Old Java directory. The current Java installation uses `x86_64/linux/java/`.                                          |
| `x86_64/jsignpdf/jsignpdf-2.3.0/`  | Old JSignPdf version. The current version is `jsignpdf-3.1.0`.                                                        |
| `aarch64/jsignpdf/jsignpdf-2.3.0/` | Old JSignPdf version. The current version is `jsignpdf-3.1.0`.                                                        |

Please change the cleanup logic to delete only explicitly listed obsolete paths like these.

Anything else in LibreSign AppData must not be deleted just because `DeleteOldBinaries` does not know about it.

The tests should cover this rule directly: the paths listed for cleanup are deleted, the current binary paths remain, and an arbitrary unknown AppData entry is never deleted.

This keeps `DeleteOldBinaries` focused on its original purpose and fixes #8415 without making every AppData feature depend on this repair step.

ciaens:
> I think we can fix this in a simpler and safer way.
> 
> `DeleteOldBinaries` should only delete paths that we explicitly know are obsolete. It should not keep a list of everything that LibreSign is allowed to store in AppData.
> 
> For this PR, these are the confirmed cleanup targets:
> 
> Path to delete	Reason
> `java/`	Old Java location at the AppData root. Java is now stored inside the architecture and Linux distribution directories.
> `x86_64/linux/java-11/`	Old Java directory. The current Java installation uses `x86_64/linux/java/`.
> `x86_64/jsignpdf/jsignpdf-2.3.0/`	Old JSignPdf version. The current version is `jsignpdf-3.1.0`.
> `aarch64/jsignpdf/jsignpdf-2.3.0/`	Old JSignPdf version. The current version is `jsignpdf-3.1.0`.
> Please change the cleanup logic to delete only explicitly listed obsolete paths like these.
> 
> Anything else in LibreSign AppData must not be deleted just because `DeleteOldBinaries` does not know about it.
> 
> The tests should cover this rule directly: the paths listed for cleanup are deleted, the current binary paths remain, and an arbitrary unknown AppData entry is never deleted.
> 
> This keeps `DeleteOldBinaries` focused on its original purpose and fixes #8415 without making every AppData feature depend on this repair step.

I'm not sure where the java 11 comes from. Also missing aarch64 equivalent?

I agree on the logic change, but then we still have to keep a list that we'll have to maintain in regard to version changes:
- when JSignPdf goes to 3.2.0 someone must remember to add jsignpdf-3.1.0
- when Java goes to 21.0.9, add jdk-21.0.8+9-jre,
- etc.

ciaens:
Updated PR desc. (with delete table notably)

vitormattos (approved):
You are right about Java 11. I got that part wrong.

I reviewed the new implementation and this approach makes sense to me. Using the release prefix inside the JSignPdf and Java directories avoids having to update the cleanup list on every dependency update, while still keeping the deletion limited to directories managed by the installer.

The tests also cover the important cases now: old releases are removed, the current release stays, and unknown AppData entries are not deleted.

Looks good from my side.

Issue #8405, opened 2026-09-15
Define configurable rejection workflow and visibility policies


vitormattos:
## Context

#8388 and #8398 showed that rejection is not only a presentation problem.

LibreSign needs a clear contract for:

* whether rejection is allowed;
* what happens after a rejection;
* whether a rejection comment is allowed or required;
* who can see the rejection;
* who can see the comment;
* what the administrator may enforce;
* what the requester may configure.

The rejection feature has not been released yet, so we can define this contract before release.

## Goal

Use the existing LibreSign policy architecture so that:

* administrators define allowed values and defaults;
* requesters choose within those limits;
* the final effective configuration is stored in `policy_snapshot`;
* active workflows use the frozen configuration;
* backend validation enforces all dependencies;
* workflow state, presentation and signing capability remain separate.

## Policy lifecycle

```mermaid
flowchart TD
    A[Administrator policy] --> B[Allowed values and defaults]
    B --> C[Requester configuration]
    C --> D[Validate]
    D --> E[Resolve effective configuration]
    E --> F[Store in policy_snapshot]
    F --> G[Signing workflow]
```

While the request is still a draft, the requester may change its rejection configuration.

Before the signing workflow starts:

1. resolve the current administrator policy;
2. validate requester choices;
3. calculate the effective configuration;
4. store it in `policy_snapshot`.

After the workflow starts, rejection settings are immutable.

Later administrator policy changes affect new workflows only.

A draft must be revalidated against the current administrator policy before it is sent.

## Policy model

Use one scalar policy key per configurable setting, grouped as one rejection policy:

```text
rejection.enabled
rejection.behavior
rejection.comment_mode
rejection.visibility
rejection.comment_visibility
```

Each key should use the existing `value_choice` resolution mode where applicable.

This allows each setting to have:

* allowed values;
* default value;
* lower-layer override rules.

Do not keep rejection as one composite value where only `enabled` can be overridden.

## Policy precedence

The administrator defines which values are allowed.

Example:

```text
rejection.behavior:
    allowedValues:
        - cancel
    default: cancel
```

`cancel` is mandatory.

With:

```text
rejection.behavior:
    allowedValues:
        - cancel
        - continue
    default: cancel
```

the requester may choose.

Do not create rejection-specific locking rules outside the existing policy system.

## Cross-setting validation

Some rules depend on more than one setting.

Use a rejection-specific validator when:

* administrator policy configuration is saved;
* requester/file configuration is validated.

[...]

PR #8423, opened 2026-09-17
refactor(policy): model signature rejection as five scalar policy settings

maia-andre:
Part of #8405 — first of the three pull requests agreed in the issue: this one is the policy model, the next one is the behavior (`continue` completing the file, `canSign`), and the third adapts #8398 to the new visibility settings.

## 📝 Summary

The rejection rules were a single composite policy value, `signature_rejection`, holding `enabled`, `comment_mode`, `cancel_workflow`, `public_status` and `show_comment_on_validation`. A composite value can only be allowed or refused as a whole, which is why only `enabled` could be overridden per request, and why the allowed values of each setting could not be expressed at all.

This replaces it with one scalar key per setting, all in the existing `value_choice` resolution mode and grouped under the setting the others depend on, the way the expiration rules are grouped under the maximum validity:

| Key | Values | Default |
| --- | --- | --- |
| `rejection_enabled` | `false`, `true` | `false` |
| `rejection_behavior` | `cancel`, `continue` | `cancel` |
| `rejection_comment_mode` | `disabled`, `optional`, `required` | `disabled` |
| `rejection_visibility` | `requester` < `participants` < `public` | `requester` |
| `rejection_comment_visibility` | `requester` < `participants` < `public` | `requester` |

The issue writes these keys as `rejection.enabled` and so on. The policy API restricts a policy key to `[a-z0-9_]+` (`PolicyController`), so they keep the spelling the rest of the project uses; the model is the one the issue describes.

**Cross-setting validation.** Rules that span more than one setting cannot live in a single key's validator, so every write also goes through `SignatureRejectionPolicyValidator` with the final combined configuration: when an administrator saves a layer (through the per-key persistence validator, reading the effective value of the other four in the same context) and when a signature request is configured. No generic cross-policy hook was added to `PolicySpec`.

A layer and a request are not checked against the same rules, on purpose. A policy layer is written one key at a time and describes a standing rule, so an administrator may set the comment mode before enabling rejection; only a combination that would disclose the comment to an audience that may not see the rejection is refused there. A signature request is submitted as a whole, so a setting that cannot take effect on that document — a comment audience while comments are disabled, anything at all while rejection is disabled — is refused instead of silently dropped. Either way the outcome does not depend on the order in which the keys are saved.

**Combined configuration.** `SignatureRejectionPolicyConfig` is what the signing flow, the snapshot and the validator work with. Building one always collapses what cannot apply: nothing is configurable while rejection is disabled, the comment audience is meaningless while comments are disabled, and the comment audience is narrowed to the rejection audience. A stale or hand-written layer therefore cannot disclose more than the rules allow. Explicitly submitting such a combination is refused instead, so a client never silently gets something other than what it asked for; what was merely inherited is corrected, so an administrator narrowing the rejection audience does not break the requests that already resolved a wider comment audience.

**Snapshot and drafts.** The document freezes one snapshot entry per setting. A draft is now revalidated against the current administrator policy on every update: a choice that is no longer allowed falls back to what the document inherits, so a workflow never starts under a revoked rule. Once the signing flow starts the configuration is frozen, and resending it stays an idempotent update.

Two consequences of the model, both intentional and worth calling out:

- **An enabled policy now applies to a request that says nothing about rejection.** Opting in per request was a property of the composite value, not a rule of the contract: the requester now chooses inside what the administrator allows, like every other policy. The system default stays disabled, so nothing changes until an administrator turns it on.
- **`cancel` is the default behavior.** The old `cancel_workflow` default mapped to `continue`, which cannot complete a file yet — that is the behavior part of the issue and comes in the next pull request, so defaulting to `cancel` avoids shipping the stuck state in between.

`FilePolicyApplier` now builds one applier per class instead of one per policy key: the rejection applier answers for five keys and would otherwise run five times on every request.

No compatibility code and no migration: the feature is not released, so the old value is simply gone, as agreed in the issue.

## ❓ Two points for you

1. **The wording of the issue.** The body says one visibility hierarchy, `requester < participants < public`, and "Do not use separate workflow and validation visibility settings", while the comment that came with the update says workflow and public validation visibility were kept as separate settings. I followed the body: five keys, one hierarchy. Tell me if the comment is what you meant and I adapt.
2. **Refusing rejection outright.** With the model, `rejection_enabled = false` at the system scope cannot enforce anything: `PolicySource::loadSystemPolicy()` treats a system layer saved with the default value as "not configured" and returns it with `allowChildOverride = true`, for every policy, not only this one. Enforcing the disabled state therefore only works through a group or user layer today (covered by an integration scenario). This is pre-existing behavior of the policy source, so I left it alone; say the word if you want a follow-up that lets an administrator enforce a default value at the system scope.

## 🧪 How to test

The five keys go through the existing policy API. In a devcontainer, as `admin`:

```bash
# the administrator allows rejection and lets the requester choose the rest
curl -u admin:admin -H 'OCS-APIRequest: true' -X POST \
  '.../ocs/v2.php/apps/libresign/api/v1/policies/system/rejection_enabled' \
  -d 'value=true' -d 'allowChildOverride=true'

# the comment cannot reach further than the rejection: this one is refused with 400
curl -u admin:admin -H 'OCS-APIRequest: true' -X POST \
  '.../ocs/v2.php/apps/libresign/api/v1/policies/system/rejection_visibility' \
  -d 'value=participants' -d 'allowChildOverride=true'
curl -u admin:admin -H 'OCS-APIRequest: true' -X POST \
  '.../ocs/v2.php/apps/libresign/api/v1/policies/system/rejection_comment_visibility' \
  -d 'value=public' -d 'allowChildOverride=true'
```

A signature request carries the requester choices in `policy.overrides`, one key each:

```json
{"overrides":{"rejection_enabled":true,"rejection_behavior":"continue","rejection_comment_mode":"required"}}
```

and the document metadata then holds one `policy_snapshot` entry per setting.

Automated coverage:

```bash
composer test:unit -- --filter 'SignatureRejection|FilePolicyApplier'
cd tests/integration && BEHAT_ROOT_DIR=../../../../ BEHAT_RUN_AS=www-data vendor/bin/behat features/sign/reject.feature
```

`reject.feature` keeps its scenarios and gains five: the policy default applying to a request, a requester opting out, an enforced setting the requester cannot change, the comment audience refused on both writing paths, and a draft revalidated against a policy that changed under it. Every scenario passes on its own here; running the whole feature in a devcontainer, a couple of scenarios fail because a policy saved by one request is not yet visible to the next one, which also happens on `main` (two scenarios failed there in the same run), so it looks like a pre-existing caching issue of the integration environment rather than something this brings.

## ⚙️ API / Back‑end changes

- [x] Five scalar `rejection_*` policy keys replace the `signature_rejection` composite value, with a rejection-specific cross-setting validator on the policy save path and on the signature request path.
- [x] One `policy_snapshot` entry per setting, draft revalidation against the current administrator policy, configuration frozen once the flow starts.
- [x] Unit and integration tests added – *required for backend changes*
- [x] API documentation updated with the command `composer openapi`

## ✅ Checklist

- [x] `composer cs:check`, `composer psalm` and `composer test:unit` are clean on this branch.

## 🤖 AI (if applicable)


lib/Service/Policy/Provider/SignatureRejection/SignatureRejectionPolicy.php
```diff
@@ -8,115 +8,253 @@
...
+			persistenceValidator: $this->combinedConfigurationValidator($key),
+		);
+	}
+
+	/**
+	 * Every saved layer is validated against the configuration it produces once
+	 * combined with the other four settings resolved in the same context, so the
+	 * result does not depend on which key was saved first.
+	 */
+	private function combinedConfigurationValidator(string $key): \Closure {
```
vitormattos, line 204: I think this still makes the result depend on the order in which the keys are saved.

For example, starting with:

`rejection_visibility = requester`
`rejection_comment_visibility = requester`

and changing both to `public`:

* saving `rejection_visibility` first works;
* saving `rejection_comment_visibility` first is rejected because the sibling reader still sees the old `requester` value.

#8405 requires cross-setting validation to use the final combined configuration, independently of the order in which individual keys are persisted.

Please change this so a compound policy update validates the complete intended set of rejection values before persisting the individual keys. The validation of `rejection_visibility` and `rejection_comment_visibility` should therefore produce the same result regardless of which key is written first.

Please also add a test that performs the same multi-key change in both orders and verifies that both are accepted.

vitormattos (changes requested):
I reviewed the PR against the current #8405 proposal.

For the visibility question, please follow the current issue body. The intended model is the single `requester < participants < public` hierarchy, so the five policy keys used in this PR are correct.

I found one remaining blocker and left an inline comment: cross-setting validation still depends on the order in which visibility keys are persisted. The compound rejection policy update should validate the final intended configuration before saving the individual keys, and the tests should cover both write orders.

For the `rejection_enabled = false` system-scope limitation, I agree this is a general limitation of the policy framework. I don't think we should add rejection-specific behavior here. We can track that separately.

I also checked the failing CI jobs. Playwright lost the test server, and Behat MySQL failed with an empty server response in an unrelated file-list scenario. I don't see evidence that either failure was caused by this PR.

lib/Service/Policy/Provider/SignatureRejection/SignatureRejectionPolicy.php
maia-andre: Agreed — the order dependence is real, and it comes from where the check lives: the cross-setting validation runs as a per-key `persistenceValidator`, so it only sees the siblings already persisted.

Before implementing, I need your call on the transport, because there is no compound write path in the framework yet: all four policy write routes are per key, and what the workbench calls a compound save is two independent parallel requests.

Note one consequence: for sequential per-key writes, accepting both orders means the transient combination cannot be refused — the first write cannot know what the second will be. So the refusal moves to where the complete set arrives at once, and a per-key layer write stops being refused on a sibling's stale value. That is safe here, because `SignatureRejectionPolicyConfig` already narrows `rejection_comment_visibility` to `rejection_visibility` when the configuration is built, so no transient or hand-written state can disclose a rejection to a wider audience than the rejection itself.

Two ways to let a caller submit the complete set, both driven by `compositeChildren` so the existing pairs could adopt them later:

**A.** Composite payload on the parent key: the routes already accept an array `value`, so when the value for `rejection_enabled` is a map of the family keys, `PolicyService` validates the combined configuration once and then persists each key, keeping the per-key scope checks and wrapping the group/user writes in a transaction. No new routes.

**B.** Dedicated compound endpoints (`POST /policies/system/compound/{parentPolicyKey}` plus the group and user variants) taking a `values` map. More explicit, at the cost of three new routes and their OpenAPI surface.

I would go with A, but it is your API — tell me which you prefer and I will implement it, with tests covering the multi-key change in both orders (both accepted), a compound write whose final combination is invalid (refused), and the request-path refusal unchanged. I will rebase onto `main` in the same push.

vitormattos: I would go with **B**.

The reason is not only that the endpoint is more explicit. There is an architectural boundary we should preserve in the policy system.

The direction is for each policy provider to behave as an isolated business module. Ideally, adding a new policy means adding its own provider/module, without spreading its business logic through `PolicyService`, controllers or the frontend.

This isolation is also important if we eventually want policies to become modules/plugins.

I see the policy framework as an orchestrator:

* it knows the contract;
* it knows when to execute a policy;
* it provides the input and context;
* it handles infrastructure concerns such as scope, transaction and persistence;
* it receives the result.

The policy itself owns the business logic. It receives its inputs, validates/processes them and returns the result.

Because of that, I don't like option A. It makes `rejection_enabled` have two meanings: a scalar boolean policy and also the transport for the complete rejection configuration.

With **B**, the compound operation is explicit.

The compound transport/orchestration can be generic, but the business semantics must stay inside the provider. The core may know that several values belong to the same provider and must be processed atomically, but it should not understand the relationship between them.

For example, the core should never know that `rejection_comment_visibility` cannot be wider than `rejection_visibility`. The complete final input should be passed to the rejection provider, and `SignatureRejectionPolicy` / `SignatureRejectionPolicyValidator` should decide whether it is valid.

So the boundary I would like to preserve is:

`orchestrator -> contract -> isolated policy provider -> result`

This keeps the policy implementation isolated today and also leaves a cleaner path if these providers become modules/plugins in the future.

maia-andre: Implemented **B**. The branch is now at `a38b6b819`, rebased on current `main`.

Four compound routes, one per layer:

```
POST /policies/compound/system/{parentPolicyKey}
PUT  /policies/compound/group/{groupId}/{parentPolicyKey}
PUT  /policies/compound/user/{parentPolicyKey}
PUT  /policies/compound/user/{userId}/{parentPolicyKey}
```

The payload carries `values` keyed by policy key plus `allowChildOverride` per key, and every key keeps the same scope and permission checks as its single-key route. `PolicyService` completes the set — the keys not sent keep the value they already resolve to in the same scope — hands the whole set to the parent policy before anything is written, and persists the keys inside one transaction.

The boundary you asked for is preserved. The framework only knows which keys belong to the family and which ones the caller is writing; whether the combination is acceptable is answered by the policy that owns it, through the new `IPolicyDefinition::validateCompositeValuesForPersistence()`. `SignatureRejectionPolicy` registers that validator on `rejection_enabled` only, and nothing in `PolicyService`, `PolicyController` or `PolicySpec` knows how the five rejection settings relate to each other.

The per-key `persistenceValidator` is gone, and with it the order dependence: a key written on its own is not a statement about its siblings. What a stored combination can disclose is still bounded, because `SignatureRejectionPolicyConfig::fromValues()` narrows the comment audience to the rejection audience whenever a configuration is built. `SiblingPolicyEffectiveValueReader` had no other caller and was removed.

Tests cover the same multi-key change written in both orders (accepted), a compound write whose resulting configuration is invalid (refused before any key is saved), a single-key write that no longer depends on its siblings, a failure halfway through (rolled back), and a group write judged against that group's own values. In `reject.feature`, the scenario that refused a wider comment audience now states the rule where it lives — the administrator path refuses the compound write — and a new scenario widens both audiences in one write with the values given in either order. The refusal on the request path is unchanged.

That stale scenario was also the only real failure in the previous CI run: it still expected `400` from the single-key route, which by design no longer carries the cross-setting rule.

Verified locally: PHPUnit for the policy and rejection suites (271 tests), `reject.feature` 27/27, `php-cs-fixer`, `psalm`, `composer openapi` (no diff) and `npm run ts:check`.

Could you re-review when you have a moment?

vitormattos (approved):
I reviewed the compound policy changes.

The order-dependent validation issue is resolved, and the implementation keeps the generic policy framework responsible for orchestration while the rejection-specific rules stay in the policy provider.

The atomic write and regression tests also cover the cases we discussed.

Looks good to me now.

PR #8447, opened 2026-09-19
refactor(release): automate release preparation and drafts

vitormattos:
> [!IMPORTANT]
> **Proof of concept only — do not merge.**
>
> Production release automation is being extracted to reusable tooling tracked by LibreCodeCoop/github-workflows#70. This PR is retained temporarily only as a source of experiments/tests and will be closed once the reusable implementation and focused LibreSign integration PRs replace it.

## Summary

Refactors the LibreSign stable release process from preparation through publication while keeping the Nextcloud-derived App Store publisher.

The workflows are intentionally sequenced by filename and UI name:

- `release-00-tests.yml` — **Release 00 - Automation tests**
- `release-10-prepare-pr.yml` — **Release 10 - Prepare release PR**
- `release-20-finalize-preparation.yml` — **Release 20 - Finalize preparation**
- `release-30-prepare-draft.yml` — **Release 30 - Prepare draft**
- `appstore-build-publish.yml` — **Release 40 - Build, sign and publish App Store** (keeps the catalog-managed filename; LibreSign-specific differences live in a local `.patch`)

The numeric gaps are intentional so intermediate stages can be added later without renaming the whole sequence.

### Release CLI

LibreSign-specific release logic is moved out of workflow shell blocks into `scripts/release/` and exposed through `scripts/release/release.php`.

The CLI now owns:

- release state and version validation
- pull request collection from the release commit range
- pending backport detection
- version bump and changelog planning
- release file updates and validation
- milestone lookup/finalization
- release notes generation
- GitHub Release draft creation/update
- release PR scope validation

Git/GitHub commands are executed through an injectable `CommandRunner`, allowing unit tests to validate command composition and state transitions without hitting GitHub.

### Per-major changelogs

Release history is now split by LibreSign major under `docs/changelogs/changelog-<major>.md`, following the repository layout used by Talk while keeping LibreSign documentation in its separate repository.

- release preparation updates only the changelog for the selected app major
- the root `CHANGELOG.md` becomes an index/entry point
- `make appstore` selects the changelog matching `appinfo/info.xml`
- the selected file is copied into the package root as `CHANGELOG.md`
- `verify-appstore-package` checks that non-development releases contain their release section
- package creation remains self-contained and does not depend on `LibreSign/documentation`

### Release 10 - Prepare release PR

- defaults to `dry_run=true`
- validates the latest stable release state
- blocks pending `backport-request` pull requests
- validates the matching `Next Patch (XX)` milestone
[...]

PR #8496, opened 2026-09-21
ci: adopt cataloged App Store release workflow

vitormattos:
Align `appstore-build-publish.yml` with the organization catalog introduced in github-workflows v0.2.0, while preserving only the two LibreSign-specific requirements as a local patch: recursive submodule checkout and providing the app private key before `make appstore` so LibreSign can sign setup binaries during packaging.


vitormattos:
Superseded by #8447. The release automation PR now keeps `appstore-build-publish.yml` on the catalog-managed filename and carries the same LibreSign-specific adaptations in a local `.patch`, so merging both would duplicate/conflict on the same workflow.

